PR #27, opened 2019-02-10
Add array support to the storage.

stancl:
(Existing tests have been modified, but only in a way that makes them more correct than they were before, making this a backwards compatible feature.)


stancl:
Added regression tests.

stancl:
All old tenant data will be json_decoded as null\*, so this change isn't 100% backwards compatible. This isn't an issue since this package doesn't have any users at this point. Though it should be kept in mind that passing tests aren't always a reliable indicator of backwards compatibility.

\*

```php
>>> tenancy()->init()
=> [
     "uuid" => null,
     "domain" => null,
     "abc" => null,
     "foo" => null,
   ]
```

Issue #12, opened 2019-01-27
Allow tenants to use multiple domains


stancl:
I'm not sure if there are many practical uses for this, but code-wise this shouldn't be hard to add.

Also add API to easily add and remove domains.

dhsont:
This will be nice feature

PR #64, opened 2019-07-10
functions to add an alias domain for a tenant

clizaola:
Often when I am doing some type of multi-tenant I have the problem in regards two or three different domains needs to be able to access the same tenant, so What I try to do is create some functions to add that functionality to your current project

### DIFF
--- a/src/Interfaces/StorageDriver.php
+++ b/src/Interfaces/StorageDriver.php
@@ -9,6 +9,7 @@ interface StorageDriver
     public function getTenantById(string $uuid, array $fields = []): array;
     public function getTenantIdByDomain(string $domain): ?string;
     public function createTenant(string $domain, string $uuid): array;
+    public function aliasTenant(string $aliasDomain, string $uuid): array;
     public function deleteTenant(string $uuid): bool;
     public function get(string $uuid, string $key);
     public function getMany(string $uuid, array $keys): array;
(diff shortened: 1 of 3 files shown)


stancl:
In the next major version (2.0.0), the domain property will be changed into an array -- `domains`; see #12.

Aliases could work as a temporary solution for 1.x, but it would be more complex than this PR's implementation.

Could you tell me more about the use case of multiple domains for a single tenant?

Thanks.

drbyte:
One use-case for multiple domains per tenant is when you allow them to bring their own domain name to your service. 
So, considering you offer a SaaS platform at `www.stancl-saas.com` and I register as a tenant,
initial onboarding would give me a subdomain on your platform: `drbyte.stancl-saas.com`
but then when ready to go live I could point my DNS for: `drbyte.com` to your server,
which internally would treat it as an alias to the `drbyte.stancl-saas.com` tenant.
This would let me do plenty of testing and demo activity before committing past a trial or upgrading to a higher plan etc.

stancl:
I see, thanks @drbyte. Okay, I'll make some changes to this PR and merge. I'll include the alias solution in 1.5.0 and the domains array solution in 2.0.0.

stancl:
Actually, changing the interface is a breaking change. I just realized that as I was implementing a similar solution to this PR.

Therefore, this feature is coming in 2.0.0.

@clizaola If you really need this feature right now, you can copy `RedisStorageDriver` and add the alias method to it. If you add it to the `StorageDriver` interface as well, that change will result in conflicts when this package receives a minor update and you run `composer update`. I don't think you need to add it to the interface, though. You're getting an instance of `RedisStorageDriver` from the service container, so the worst case scenario should be that your IDE will shout at you.

clizaola:
Copy that @stancl, keep up the good work, I will be watching closely all the progress in this package! Thanks

PR #82, opened 2019-08-08
[1.7.0] Add DB storage driver

### LINKED ISSUE #53: Add a MySQL/SQLite storage driver

stancl:
Closes #53

#### TODO:
- [ ] Write docs for publishing & customizing the migration
    - [ ] Update TOC
    - [ ] Add note about `get(['key1', 'key2'])` not returning associative arrays, create an issue for v2
- [x] Write the storage driver
- [x] Run tests with both storage drivers
- [x] Fix coverage merging
- [x] Move assets to `/assets`
- [x] Fix how the tenant model is resolved from the SC
- [x] Tenant model todos
- [x] Migration cannot be published
- [x] Fix json encoding


carlos-mora:
Cool!
I pulled the branch and after a couple of touches it seems to start.
for the Tenant model I've added:

`<?php

namespace Stancl\Tenancy;

use Illuminate\Database\Eloquent\Model;

class Tenant extends Model
{
    /**
     * The primary key associated with the table.
     *
     * @var string
     */
    protected $primaryKey = 'uuid';
    public $incrementing = false;
    
    /**
     * The attributes that are mass assignable.
     *
     * @var array
     */
    protected $fillable = [
        'uuid', 'domain', 'data',
    ];
}`

In DatabaseStorageDriver

    public function getTenantById(string $uuid, array $fields = []) : array
    {
        $tenant = Tenant::find($uuid)->toArray();
        return $tenant;
    }

With that, copying the tenant table migration (Added timestams() there) things at least get to boot.

Thanks! KR

stancl:
Hi @carlos-mora. I'm working on that right now. If you're interested in helping with this, we can discuss this in real time on Discord (stancl on Laravel Discord) or Telegram (@samuelstancl).

I went with $guarded instead of $fillable, because my idea is that by default, data gets stored in the data JSON column, but developers can create their own columns for some keys. This is useful when indexes and similar features are needed.

carlos-mora:
Hi Samuel,

It's a little bit late here in Madrid, my wife is arriving so I'm really
sorry, I can't go online right now. Anyway I'm following the project very
close and trying it asap, I have to make multitenant an existing project,
and your package fits perfectly, besides the fact I don't use Redis.

Please consider declaring the PK in Tenant's model, and set it as
non-integer, otherwise the uuid attribute behaves weird. P/E, $tenant->uuid
returns 0 instead of the real attribute value. Adding this will fix the
behavoir:

/**
* The primary key associated with the table.
*
* @var <https://github.com/var> string
*/
protected $primaryKey = 'uuid';
public $incrementing = false;

If you don't change the $primaryKey, Tenant::find($uuid) will crash
with a "Column
not found: 1054 Unknown column 'tenants.id' in 'where clause' (SQL: select
* from `tenants` where `tenants`.`id` =
d51e24a0-badf-11e9-99d8-03661283d12a limit 1)"


Please remember to add to the TODO list an artisan command to export the
Tenants's table migration.

KR, and thanks!

stancl:
No problem. Thanks for the tips about the public key. Right now the tests are passing for me even though they shouldn't. Seems like the tests still use the Redis storage driver. I will look into that tomorrow. I might end up writing a test class for each storage driver, but that feels a bit unnecessary, so I hope the current tests can cover this well enough.

I'll merge this PR tomorrow and tag a new version so that this feature is released this weekend.

I wanted to add the Telescope tags feature in this release but it's probably not going to be finished soon enough and I'd also like to get some feedback on what's the best way to set tags as a package, ideally from Laravel staff.

---

> Merging #82 into 1.x will decrease coverage by 90.86%.

There seems to be an issue with the way the coverage clovers are merged.

carlos-mora:
Here Carlos reporting !
I'm testing the package by using it, so I'll find some issues.
In Tenant model, it is 'primaryKey', not 'publicKey'

This line in DatabaseStorageDriver crashes:
return Tenant::find($uuid)->only($fields)->toArray();
only() returns an array, so toArray will fail with an "Call to a member
function toArray() on array" error

carlos-mora:
Another update in DatabaseStorageDriver. If no $fields given, only($fields)
returns [], so, according to the RedisStorage bahavoir:

public function getTenantById(string $uuid, array $fields = []): array
{
$fields = (array) $fields;

if (! $fields) {
return Tenant::find($uuid)->toArray();
}
return Tenant::find($uuid)->only($fields);
}

stancl:
`dd(config('tenancy.storage_driver'), tenancy()->getStorageDriver());` in setUp() shows

```php
"Stancl\Tenancy\StorageDrivers\DatabaseStorageDriver"
Stancl\Tenancy\StorageDrivers\RedisStorageDriver^ {#421
```

So the issue is that I'm just the config value after TenantManager has been initialized.

stancl:
The tests now say that the tenants table was not found, so that's good. Should work after I find a way to execute the migration in setUp().

carlos-mora:
Another touch for Tenant.php model: please add a

public $timestamps = false;

or add $table-timestamps() in Tenant's migration, Otherwise it complains in
the attempt to create a new tenant:

>>> $domain = 'tenant1.portal.local'
=> "tenant1.portal.local"
>>> tenant()->create('tenant1.local')
Illuminate/Database/QueryException with message 'SQLSTATE[42S22]: Column
not found: 1054 Unknown column 'updated_at' in 'field list' (SQL: insert
into `tenants` (`uuid`, `domain`, `updated_at`, `created_at`) values
(2fbd0580-bb43-11e9-b9a4-a50e2f24d16e, tenant1.local, 2019-08-10 07:48:04,
2019-08-10 07:48:04))'

src/StorageDrivers/DatabaseStorageDriver.php
```diff
@@ -2,16 +2,10 @@
 
 namespace Stancl\Tenancy\StorageDrivers;
 
-use Stancl\Tenancy\Interfaces\TenantModel;
 use Stancl\Tenancy\Interfaces\StorageDriver;
```
carlos-mora, line 7: use Stancl\Tenancy\Tenant;

src/StorageDrivers/DatabaseStorageDriver.php
```diff
@@ -32,53 +26,53 @@ public function identifyTenant(string $domain): array
...
+        return Tenant::find($id)->delete();
     }
 
     public function getAllTenants(array $uuids = []): array
     {
-        return $this->tenant->all()->map(function ($model) {
+        return Tenant::all()->map(function ($model) {
             return $model->toArray();
         })->toArray();
     }
```
carlos-mora, line 59: `

        if ($uuids) {
            $tenants = Tenant::find($uuids);
        } else {
            $tenants = Tenant::all();
        }
        return $tenants->toArray();
`

src/Tenant.php
```diff
@@ -21,9 +18,19 @@ class Tenant extends Model implements TenantModel
      */
     private $dataObject;
 
+    public function dataColumn()
+    {
+        return config('tenancy.db_storage.data_column');
+    }
```
carlos-mora, line 24: `return config('tenancy.db_storage.data_column', 'data');`
a little bit safer

src/Tenant.php
```diff
@@ -21,9 +18,19 @@ class Tenant extends Model implements TenantModel
...
+
+    public function customColumns()
+    {
+        return config('tenancy.db_storage.custom_columns');
+    }
```
carlos-mora, line 69: like 'data' column, could be default to []

src/assets/migrations/2019_08_08_000000_create_tenants_table.php
```diff
@@ -16,7 +16,7 @@ public function up()
         Schema::create('tenants', function (Blueprint $table) {
             $table->string('uuid', 36)->primary();
             $table->string('domain', 255)->index();
-            $table->string('data', 1024);
+            $table->string('data', 1024)->default('{}');
         });
```
carlos-mora: `$table->text('data')->default('{}');`

or 

`$table->json('data');`

src/Tenant.php
```diff
@@ -12,6 +12,7 @@ class Tenant extends Model implements TenantModel
     protected $guarded = [];
     protected $primaryKey = 'uuid';
     public $incrementing = false;
+    public $timestamps = false;
```
carlos-mora, line 13: // As recommended in Laravel docs, when the primary key is not an integer then:
`

    /**
     * The "type" of the auto-incrementing ID.
     *

     * @var string

     */

    protected $keyType = 'string';

`

carlos-mora (changes requested):

src/StorageDrivers/DatabaseStorageDriver.php
stancl: Nice catch, I would have missed that $uuids arg.

src/assets/migrations/2019_08_08_000000_create_tenants_table.php
stancl: I'll go with `$table->json('data')->default('{}');`

src/Tenant.php
stancl: > The "type" of the auto-incrementing ID.

I don't think this is necessary, since we have set `$incrementing = false`

PR #127, opened 2019-09-24
Rename migration stubs so they show next to each other

drbyte:
This is helpful since Laravel 6's default install gives a `failed_jobs` migration file that sits between these files before renaming them.


stancl:
The CommandsTest file has a test case that makes sure the migration file is created.

Can you update that reference too?

Speaking of that, I probably forgot to add the assertion for the domains migration.

drbyte:
Fixed

stancl:
Thanks!

PR #128, opened 2019-09-25
[2.x] Add `tenancy:migrate-fresh` command

drbyte:
This calls the `db:wipe` command, like `artisan migrate:fresh` does before it does a migration.

It can't inherit from `FreshCommand` because it doesn't inherit from `BaseCommand` which detects migration paths. So, had to copy in (part of) the code from `FreshCommand` to do the wipe portion. The rest then just does migration.


drbyte:
I didn't add tests because I couldn't decide "what" to test, without manually creating unused tables and then asserting that they were gone. Maybe that's the right way, but I figured it better to have you weigh in on your perspective for this.

stancl:
I think it should extend FreshCommand to have the same options. And call tenants:migrate inside. I'll look into how the db:wipe command should be scoped to the tenant.

drbyte:
You'll have to also add complete support for passing in the migrations directory, since FreshCommand has no extensibility for that.

stancl:
I'd use the same migrations directory as `tenants:migrate` by default. See this trait: https://github.com/stancl/tenancy/blob/2.x/src/Traits/DealsWithMigrations.php

drbyte:
Ya, but like I said in my initial comments, that trait can't be used by FreshCommand because FreshCommand inherits from a different parent that knows nothing of getMigrationPath/etc. It was a long rabbit-hole to travel, where Laravel's commands are inconsistent. I dunno if they'd be open to a PR to make them match.

stancl:
Ah, I see: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Console/Migrations/FreshCommand.php#L49

I think @antonkomarev created the PR to laravel/framework that introduced `db:wipe`. Anton, perhaps your perspective on this could be useful?

antonkomarev:
I'm using `db:wipe` very frequently while developing applications. But usually I'm wiping only system central database and then deleting all tenants databases manually. Especially it's useful on early stages when you are developing tenant creation & deployment process.

stancl:
Do you think we need all the features of Illuminate `FreshCommand` or we're fine with a custom, simpler implementation? I probably agree with @drbyte and think that this implementation is really enough. But I'm a little unsure because of some of the DB engine-specific options like `drop-views`, `drop-types` etc.

antonkomarev:
I don't have a solid decision at this moment. But adding new options to command is not a breaking change and it could be done in minor releases.

src/Commands/MigrateFresh.php
```diff
@@ -0,0 +1,77 @@
...
+namespace Stancl\Tenancy\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Database\Console\Migrations\MigrateCommand;
+use Illuminate\Database\Migrations\Migrator;
+use Stancl\Tenancy\DatabaseManager;
+use Stancl\Tenancy\Traits\DealsWithMigrations;
+use Stancl\Tenancy\Traits\HasATenantsOption;
+
+class MigrateFresh extends MigrateCommand
```
stancl: I'll review this thoroughly later, but now one thing I'd do is make this class final:
```suggestion
final class MigrateFresh extends MigrateCommand
```
@antonkomarev pointed out that
> adding new options to command is not a breaking change

That is assuming users won't extend this class. So for now, I'd make it final.

src/Commands/MigrateFresh.php
```diff
@@ -0,0 +1,77 @@
...
+        tenancy()->all($this->option('tenants'))->each(function ($tenant) {
+            $this->line("Tenant: {$tenant['id']}");
+
+            // See Illuminate\Database\Migrations\DatabaseMigrationRepository::getConnection.
+            // Database connections are cached by Illuminate\Database\ConnectionResolver.
+            $this->input->setOption('database', 'tenant');
+            tenancy()->initialize($tenant);
+
+            // Fresh
+            $this->call('db:wipe', array_filter([
```
stancl, line 62: Why the `array_filter`?

stancl:
Closing in favor of #148

src/Commands/MigrateFresh.php
drbyte: That's a fragment from Laravel's `wipe` command. It should have been removed when I removed the other options that it filters against.

antonkomarev: I've added `array_filter` to `db:wipe` just because other Laravel commands has it in programmable executed calls:
```php
$this->call('migrate:install', array_filter([
    '--database' => $this->option('database'),
]));

$this->call('db:seed', array_filter([
    '--database' => $database,
    '--class' => $this->option('seeder') ?: 'DatabaseSeeder',
    '--force' => true,
]));

```

stancl: I see. It's probably because it will remove flags like the `--force` if it's false. I'll add that to the other PR.

antonkomarev: `array_filter` was introduced in this PR: https://github.com/laravel/framework/pull/27020

stancl: Makes sense. Thanks for the link @antonkomarev.

PR #133, opened 2019-09-26
[2.x] Use ::class instead of strings in config

antonkomarev:


drbyte:
This change will require updating the docs to match.
For example: https://github.com/stancl/tenancy-docs/blob/master/docs/source/v2/configuration.blade.md
`This config is an array. The key is the alias and the value is the full class name.` and the examples.

stancl:
Thanks @antonkomarev.

@drbyte I opened an issue (https://github.com/stancl/tenancy-docs/issues/13). It's not really urgent, but for consistency, it should be fixed one day.

PR #135, opened 2019-09-26
[2.x] Add missing field in InitializeTenancy middleware

antonkomarev:
`onFail` field was declared dynamically.

And one more question accordingly this middleware: if `$onFail` is null - from where `$e` will be taken for the fallback closure?

### DIFF
--- a/src/Middleware/InitializeTenancy.php
+++ b/src/Middleware/InitializeTenancy.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Stancl\Tenancy\Middleware;
 
 use Closure;
+use Stancl\Tenancy\Exceptions\TenantCouldNotBeIdentifiedException;
 
 class InitializeTenancy
 {
@@ -31,7 +32,7 @@ class InitializeTenancy
     {
         try {
             tenancy()->init();
-        } catch (\Exception $e) {
+        } catch (TenantCouldNotBeIdentifiedException $e) {
             ($this->onFail)($e);
         }
 
--- a/src/Middleware/InitializeTenancy.php
+++ b/src/Middleware/InitializeTenancy.php
@@ -11,7 +11,7 @@ class InitializeTenancy
     /**
      * @var \Closure
      */
-    private $onFail;
+    protected $onFail;
 
     public function __construct(Closure $onFail = null)
     {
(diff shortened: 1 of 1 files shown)


src/Middleware/InitializeTenancy.php
```diff
@@ -8,6 +8,11 @@
 
 class InitializeTenancy
 {
+    /**
+     * @var \Closure
+     */
+    private $onFail;
```
stancl: `protected` please.

stancl (changes requested):

stancl:
The exception is taken from the catch block and if no callback is provided, the default one just throws it.

https://github.com/cybercog/stancl-tenancy/blob/cb07fde42887498354bee7c2b52f641e15025e86/src/Middleware/InitializeTenancy.php#L34

antonkomarev:
Done.

IMHO: I don't like that it catches `\Exception`. If you are trying to catch specific exception from `tenancy()->init()` - then it will be better create custom exception class. If you want to catch any exceptions - then you'd better catch `\Throwable` interface.

stancl:
Good point. Can you update it to catch only TenantCouldNotBeIdentifiedException?

src/Middleware/InitializeTenancy.php
```diff
@@ -5,9 +5,15 @@
...
 
 use Closure;
+use Stancl\Tenancy\Exceptions\TenantCouldNotBeIdentifiedException;
 
 class InitializeTenancy
 {
+    /**
+     * @var \Closure
+     */
+    protected $onFail;
```
stancl, line 13: Please make the annotation single-line:
```php
/** @var callable */
```

src/Middleware/InitializeTenancy.php
```diff
@@ -5,9 +5,15 @@
...
+
     public function __construct(Closure $onFail = null)
```
stancl:
```suggestion
     public function __construct(callable $onFail = null)
```

I think this should be callable, not just `Closure`. Thoughts @antonkomarev?

stancl (changes requested):

src/Middleware/InitializeTenancy.php
antonkomarev: It seems reasonable to use `callable` type here.

stancl (approved):

PR #136, opened 2019-09-26
[2.x] Fix variables code style consistency

antonkomarev:


stancl:
That's actually correct. StyleCI probably doesn't have fixers for variable name style, but the style I use here is `snake_case` for local variables and `camelCase` for object properties.

antonkomarev:
But why this class code style is different from other ones?

stancl:
You're right. It looked right to me in this case but I out of the few local variables I use, I seem to use camelCase. These variables are not even necessary, I just added them so that the middleware's logic is more readable than an if statement with an `&&`. I'll take another look at this class.

PR #137, opened 2019-09-26
[2.x] Add missing parent::__construct call in exceptions

antonkomarev:
Some of the exceptions missing `parent::__construct()` call.


stancl:
Thanks!

PR #139, opened 2019-09-26
[2.x] Reorder Service Provider methods in execution sequence

antonkomarev:
Not so far Laravel reordered methods in the same way. Initially they were ordered alphabetically (`boot`, `register`), and now they are ordered in execution sequence (`register > boot`).


stancl:
Thanks! Makes sense that it's reordered now.

PR #141, opened 2019-09-26
[2.x] Make Domain's model domain key as a primary

### LINKED ISSUE #140: [2.x] Tenant Domain model identifier
Domain model doesn't have identifier, so how users will edit those records?

```php
$domain = Domain::where('domain', 'test.example.com')->first();
```

If column `domain` is an identifier - how will it be represented in API?

```js
{
    "type": "Domain",
    "id": "test.example.com"
    "attributes": {
        "domain": "test.example.com
    }
}
```

antonkomarev:
Fixes #140

### DIFF
--- a/assets/migrations/2019_09_15_000020_create_domains_table.php
+++ b/assets/migrations/2019_09_15_000020_create_domains_table.php
@@ -16,8 +16,8 @@ class CreateDomainsTable extends Migration
     public function up()
     {
         Schema::create('domains', function (Blueprint $table) {
-            $table->string('domain', 255)->primary(); // don't change this
-            $table->string('tenant_id', 36); // 36 characters is the default uuid length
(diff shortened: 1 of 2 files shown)


assets/migrations/2019_09_15_000020_create_domains_table.php
```diff
@@ -16,8 +16,8 @@ class CreateDomainsTable extends Migration
     public function up()
     {
         Schema::create('domains', function (Blueprint $table) {
+            $table->string('domain', 255)->primary(); // don't change this
```
stancl: Realized I have way too many of these comments in the codebase.
```suggestion
            $table->string('domain', 255)->primary();
```

stancl (changes requested):

stancl:
Thanks!

PR #142, opened 2019-09-26
[2.x] Change Tenant id string column type to UUID

antonkomarev:
Laravel has special `uuid` type in migrations which stores IDs in `UUID` if available in driver, otherwise fallback to `CHAR` or `VARCHAR`.

Additionally added `void` return type to follow same code style and replaced strict `drop` method to `dropIfExists` for cases with corrupted databases.

### DIFF
--- a/assets/migrations/2019_09_15_000010_create_tenants_table.php
+++ b/assets/migrations/2019_09_15_000010_create_tenants_table.php
@@ -13,10 +13,10 @@ class CreateTenantsTable extends Migration
      *
      * @return void
      */
-    public function up()
+    public function up(): void
     {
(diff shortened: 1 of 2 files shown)


stancl:
It's intentionally `id` because it doesn't *have* to be a uuid.

You can do `Tenant::create('foo.localhost', ['id' => '123'])` and the tenant will have an id of `123`.

I just mention that it's the length of a uuid, because that's what the package uses by default when no `id` is explicitly set.

But I'll accept the `dropifExists` changes :+1:

antonkomarev:
Added it as another PR #143

PR #144, opened 2019-09-26
[2.x] Remove request helper from TenantManager

antonkomarev:
I've removed `request()` helper from TenantManager's `init` method because it's called only from single place (from middleware) and there we have $request already.

### DIFF
--- a/src/Middleware/InitializeTenancy.php
+++ b/src/Middleware/InitializeTenancy.php
@@ -29,7 +29,7 @@ class InitializeTenancy
     public function handle($request, Closure $next)
     {
         try {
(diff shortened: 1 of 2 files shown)


stancl:
The middleware change is nice because it's more explicit and uses a variable instead of a helper. But I'd keep the default value in init() for convenience and backwards compatibility.

antonkomarev:
As you wish. Just added `|null` type in DocBlock

stancl:
I was thinking that removing the default wouldn't really break anything, but I think it's a nice "fallback", when no domain is specified and this is called perhaps in some custom middleware.

PR #150, opened 2019-09-30
[2.x] Replace webpatser/laravel-uuid with ramsey/uuid

### LINKED ISSUE #134: [2.x] Replace Webpatser UUID with Ramsey UUID
Laravel uses `"ramsey/uuid": "^3.7",` package. I think it would be better to use it too and not add extra dependency.

stancl:
Resolves #134

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,8 @@
     ],
     "require": {
         "illuminate/support": "^6.0",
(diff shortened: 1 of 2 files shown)


stancl:
I initially wanted to use namespace (DNS)-based uuids, but they return the same result given the same domain, which might be problematic, so I just went with a fully random UUID type 4.

PR #159, opened 2019-10-07
[2.x] Add migration alert to install command

drbyte:


src/Commands/Install.php
```diff
@@ -74,12 +74,14 @@ public function handle()
 
         $this->line('');
         $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. If you're going to use the database storage, you need to create a tenants table.");
+        $migrationsPublished = false;
         if ($this->confirm('Do you want to publish the default database migrations?', true)) {
             $this->callSilent('vendor:publish', [
             '--provider' => 'Stancl\Tenancy\TenancyServiceProvider',
             '--tag' => 'migrations',
             ]);
             $this->info('✔️  Created migrations.');
```
stancl: Maybe we could just append `DON'T FORGET TO RUN THEM` to this line?

Not sure about the caps, I'll think about it.

drbyte: I considered just appending, but thought the "error" highlighting would draw more attention.
But, appended text in caps, or surrounded with a bunch of asterisks may be sufficient, since it will break pattern against the other status messages and therefore will stand out.

stancl: I think that the install command is not verbose enough to need highlights actually. I'd just go with this:

```suggestion
            $this->info('✔️  Created migrations. Don't forget to run them.');
```

PR #160, opened 2019-10-08
[2.x] Allow automatic seeding after automatic migrations

### LINKED ISSUE #158: [2.x] Add automated seeding to automatic migration
#### Description
v2 allows for automatic migrations upon tenant creation. 
It should also allow for a defined Seeder (which could call sub-seeders) to be run as part of those migrations.

drbyte:
Fixes #158

If `config('tenancy.seeder_class')` is null or a blank string, the seeder-class parameter is not passed, and therefore Laravel's default option of calling 'DatabaseSeeder' will occur.

As with Laravel's normal seeding process, the class is treated as the "root" seeder class, which can call other seeder classes in turn.

### DIFF
--- a/src/TenantManager.php
+++ b/src/TenantManager.php
@@ -76,6 +76,21 @@ class TenantManager
                     ]);
                 },
             ];
+
+            if ($this->shouldSeedAfterMigration()) {
+                $seederClassName = $this->getSeederRootClass();
+                $seederClassParameter = ! empty($seederClassName) ? ['--class' => $seederClassName] : [];
+
+                $afterCreating += $this->databaseCreationQueued() ? [
+                    new QueuedTenantDatabaseSeeder($tenant, $seederClassName),
+                ] : [
+                    function () use ($tenant, $seederClassParameter) {
+                        $this->artisan->call('tenants:seed', [
+                            '--tenants' => [$tenant['id']],
+                         ] + $seederClassParameter);
+                    },
+                ];
+            }
         }
 
         $this->database->createDatabase($tenant, $afterCreating);
@@ -321,6 +336,11 @@ class TenantManager
         return $this->app['config']['tenancy.migrate_after_creation'] ?? false;
     }
 
+    public function shouldSeedAfterMigration(): bool
+    {
+        return $this->shouldMigrateAfterCreation() && $this->app['config']['tenancy.seed_after_migration'] ?? false;
+    }
+
     public function databaseCreationQueued(): bool
     {
         return $this->app['config']['tenancy.queue_database_creation'] ?? false;
@@ -329,6 +349,11 @@ class TenantManager
     public function shouldDeleteDatabase(): bool
     {
         return $this->app['config']['tenancy.delete_database_after_tenant_deletion'] ?? false;
+    }
+
+    public function getSeederRootClass()
+    {
+        return $this->app['config']['tenancy.seeder_class'] ?? null;
     }
 
     /**
--- a/src/TenantManager.php
+++ b/src/TenantManager.php
@@ -10,6 +10,7 @@ use Illuminate\Support\Collection;
 use Stancl\Tenancy\Contracts\TenantCannotBeCreatedException;
 use Stancl\Tenancy\Exceptions\TenantCouldNotBeIdentifiedException;
 use Stancl\Tenancy\Jobs\QueuedTenantDatabaseMigrator;
+use Stancl\Tenancy\Jobs\QueuedTenantDatabaseSeeder;
 
 /**
  * @internal Class is subject to breaking changes in minor and patch versions.
@@ -67,7 +68,7 @@ class TenantManager
         $afterCreating = [];
 
         if ($this->shouldMigrateAfterCreation()) {
-            $afterCreating += $this->databaseCreationQueued() ? [
+            $afterCreating = array_merge($afterCreating, $this->databaseCreationQueued() ? [
                 new QueuedTenantDatabaseMigrator($tenant),
             ] : [
                 function () use ($tenant) {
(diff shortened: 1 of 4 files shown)


stancl:
Looks very good! I was thinking that there should be a separate job for the seeding, but:
1. dispatching two jobs for this could be a bit painful (race conditions)
2. You can't seed without migrating anyway

So using the same job is probably the right approach. I'll think about how to rename it, so that it isn't called "migrator" when it also seeds.

Speaking of point 1, there is probably a race condition in my code: the migrate job could run sooner than the DB creation job. I'll look into that.

rodruiz:
> Looks very good! I was thinking that there should be a separate job for the seeding, but:
> 
> 1. dispatching two jobs for this could be a bit painful (race conditions)
> 2. You can't seed without migrating anyway
> 
> So using the same job is probably the right approach. I'll think about how to rename it, so that it isn't called "migrator" when it also seeds.
> 
> Speaking of point 1, there is probably a race condition in my code: the migrate job could run sooner than the DB creation job. I'll look into that.

@stancl maybe you can use [Job Chaining](https://laravel.com/docs/6.x/queues#job-chaining). Job chaining allows you to specify a list of queued jobs that should be run in sequence after the primary job has executed successfully. 
```php
SetupTenant::withChain([
    new CreateTentantDB,
    new MigrateTentantDB,
])->dispatch();
```

stancl:
@rodruiz Great suggestion, thanks! I don't work with jobs a lot so I haven't thought about that. Could make the code a lot cleaner :+1:

src/TenantManager.php
```diff
@@ -71,6 +71,14 @@ public function createTenant(Tenant $tenant): self
...
                     '--tenants' => [$tenant['id']],
                 ]);
+
+                if ($this->shouldSeedAfterMigration()) {
+                    $seederClass = $this->getSeederRootClass();
+                    $seederClassParameter = ! empty($seederClass) ? ['--class' => $seederClass] : [];
+                    $this->artisan->call('tenants:seed', [
+                        '--tenants' => [$tenant['id']],
+                    ] + $seederClassParameter);
+                }
```
stancl:
```php
if ($this->shouldSeedAfterMigration()) {
    $afterCreating += $this->databaseCreationQueued() ? [
        new QueuedTenantDatabaseSeeder($tenant),
    ] : [
        function () use ($tenant) {
            $this->artisan->call('tenants:seed', [
                '--tenants' => [$tenant['id']],
            ]); // + the options
        },
    ];
}
```

src/TenantManager.php
```diff
@@ -311,6 +319,11 @@ public function shouldMigrateAfterCreation(): bool
         return $this->app['config']['tenancy.migrate_after_creation'] ?? false;
     }
 
+    public function shouldSeedAfterMigration(): bool
+    {
+        return $this->app['config']['tenancy.seed_after_migration'] ?? false;
```
stancl: Safety check:

```suggestion
        return $this->shouldMigrateAfterCreation() && $this->app['config']['tenancy.seed_after_migration'] ?? false;
```

src/Jobs/QueuedTenantDatabaseMigrator.php
```diff
@@ -34,5 +42,11 @@ public function handle()
         Artisan::call('tenants:migrate', [
             '--tenants' => [$this->tenant->id],
         ]);
+
+        if ($this->shouldRunSeeder) {
```
stancl: Could you extract this logic into a separate, queuable `QueuedTenantDatabaseSeeder`?

stancl (commented):
I made some changes to how the queued jobs work:
- there's now only one config that decides if the jobs will be queued (`queue_database_creation`)
- job chains are used to prevent race conditions

Could you fix conflicts & update some things as per my review?

If you don't want to deal with the conflicts feel free to create a new PR or I can create a new PR.

src/TenantManager.php
```diff
@@ -67,15 +68,30 @@ public function createTenant(Tenant $tenant): self
         $afterCreating = [];
 
         if ($this->shouldMigrateAfterCreation()) {
-            $afterCreating += $this->databaseCreationQueued() ? [
+            $afterCreating = array_merge($afterCreating, $this->databaseCreationQueued() ? [
```
drbyte: Note: I changed these to `array_merge` because the `+=` was behaving like `array_unique` instead. :(

stancl: Yeah, I also noticed some weird behavior with that operator. Thought the syntax looked nice. Look at my PR to your `seeder` branch, I use `[]` there, since the arrays have only one item anyway.

drbyte: Oh, duh. Just adding using `$foo[] =` is definitely easier!

tests/TenantManagerTest.php
```diff
@@ -235,6 +237,28 @@ public function automatic_migrations_work()
         $tenant2 = Tenant::create(['bar.localhost']);
         tenancy()->initialize($tenant2);
         $this->assertTrue(\Schema::hasTable('users'));
+        $this->assertCount(0, \DB::select('select * from users'));
```
drbyte: I guess this could be changed to `assertSame` too?
Maybe it's not even needed.
I only included it to differentiate between outcomes of migrations vs seeders.

stancl: Oh, yeah, nice catch.

drbyte: Maybe it's not even needed?

stancl: Actually, on second thought, I'd remove the assertion. Just assert that the table doesn't exist. Making queries to nonexistent tables is confusing and I'm surprised it doesn't throw an exception.

drbyte: Removal makes sense.

There's no exception because the table does exist at that point. Just empty

stancl: Ah, I misread the line above, thought it's `assertFalse`. But yeah, I'd remove it anyway because we're testing migrations at this point.

drbyte: Done

PR #169, opened 2019-10-12
[2.1.0] Initialize tenancy before executing controller constructors

### LINKED ISSUE #168: [2.1.0] Controller constructors are executed before tenancy is initialized
Tenancy is initialized using route middleware. Laravel executes route middleware *after* controller constructors.

This will be fixed in 2.1.0.

### LINKED ISSUE #170: 404 on tenant routes when running HTTP tests
I'm really enjoying this package so far and really appreciate all your hard work! 
I have run into something I can't figure out.

I would like to be able to write a feature test that sets up a Tenant in the database, creates a user and then attempts to login via an HTTP request similar to the [Laravel HTTP Tests Documentation](https://laravel.com/docs/5.7/http-tests#introduction)

**The Problem**
When the test sends the POST request to attempt a login - I receive a 404 on a route that is defined in the `routes/tenant.php` file.

The `Tenant::createFrom()` method is successfully creating the tenant in the central database, creating the tenants DB and creating a user with the correct credentials in the respective `Tenant`'s database. I can verify this by visiting the tenant site and logging in with my browser.

Is it possible that the `routes/tenant.php` file is not loaded automatically for testing?

~~~php

<?php

namespace Tests\Feature;
use App\Models\Tenant;
use Tests\TestCase;

class TenantRegistrationTest extends TestCase
{

    /** @test */
    public function a_new_tenant_can_login()
    {

        $fqdn = str_random(5).'-login.multi-tenant-app.test';

        $tenant = Tenant::createFrom($fqdn, 'name', 'bat@man.com', 'password');

       tenancy()->initializeTenancy($tenant->tenant);

        $this->withoutExceptionHandling();

        
        $form = $this->post('https://'.$fqdn.'/login', [
                                'email' => 'bat@man.com',
                                'password' => 'password'
                            ]);
        
        /** 
        *   Fails here - Response status code [404] is not a successful status code. Failed asserting 
        *   that false is true.
        **/
        $form->assertSuccessful();

    }
}
~~~

I have the following in `routes/tenant.php`
~~~php
    Route::get('login', 'Auth\LoginController@showLoginForm');
    Route::post('login', 'Auth\LoginController@login')->name('login');
    Route::post('logout', 'Auth\LoginController@logout')->name('logout');
~~~

`App\Models\Tenant.php`
~~~php
public static function createFrom($fqdn, $name, $email, $password = null): self
    {
        $tenant = StanclTenant::create([$fqdn]);

        tenancy()->init($fqdn);

        static::addRolesAndPermissions($fqdn);
        static::addDefaultSettings($fqdn);


        $admin = static::makeAdmin($fqdn, $name, $email, $password);

        tenancy()->end();

        return new self($tenant, $admin);
    }
~~~

stancl:
Fixes #168, fixes #170

TODO:
- [x] Update install command?
- [x] Update docs docs:30

### DIFF
--- a/src/Commands/Install.php
+++ b/src/Commands/Install.php
@@ -59,9 +59,9 @@ class Install extends Command
 | Tenant Routes
 |--------------------------------------------------------------------------
 |
-| Here is where you can register tenant routes for your application. These
-| routes are loaded by the TenantRouteServiceProvider within a group
-| which contains the \"InitializeTenancy\" middleware. Good luck!
+| Here you can register the tenant routes for your application.
+| These routes are loaded by the TenantRouteServiceProvider
+| with the tenancy and web middleware groups. Good luck!
 |
 */
 
@@ -73,11 +73,11 @@ Route::get('/app', function () {
         $this->info('✔️  Created routes/tenant.php');
 
         $this->line('');
-        $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. If you're going to use the database storage, you need to create tables for tenants and domains.");
-        if ($this->confirm('Do you want to publish the default database migrations?', true)) {
+        $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. To store data about tenants in a relational database, you need a few database tables.");
+        if ($this->confirm('Do you wish to publish the migrations that create these tables?', true)) {
             $this->callSilent('vendor:publish', [
-            '--provider' => 'Stancl\Tenancy\TenancyServiceProvider',
-            '--tag' => 'migrations',
+                '--provider' => 'Stancl\Tenancy\TenancyServiceProvider',
+                '--tag' => 'migrations',
             ]);
             $this->info('✔️  Created migrations. Remember to run [php artisan migrate]!');
         }
--- a/src/Commands/Install.php
+++ b/src/Commands/Install.php
@@ -73,7 +73,7 @@ Route::get('/app', function () {
         $this->info('✔️  Created routes/tenant.php');
 
         $this->line('');
-        $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. If you're going to use the database storage, you need to create a tenants table.");
+        $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. If you're going to use the database storage, you need to create tables for tenants and domains.");
         if ($this->confirm('Do you want to publish the default database migrations?', true)) {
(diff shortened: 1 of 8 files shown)


stancl:
Just tested this in a fresh Laravel application and everything seems to work now. Central/tenant routes 404 on wrong domains and tenancy is initialized before the constructor is executed.

This is how the request lifecycle works now, top to bottom:
- **global middleware:** if the current route is not exempt from tenancy, initialize tenancy
- **router**
- **controller constructor**
- **route middleware**: `PreventAccess`

When visiting a central route from a tenant domain, the constructor would be executed prior to the route middleware terminating the request. As long as the constructor does what it should do -- dependency injection and controller-specific middleware -- there should be no issues with it. I can only imagine this being problematic if the constructor created some DB records prior to the request being terminated.

@drbyte Can you test the `dev-controller-constructor` version constraint?

drbyte:
Works great!

src/Commands/Install.php
```diff
@@ -73,7 +73,7 @@ public function handle()
         $this->info('✔️  Created routes/tenant.php');
 
         $this->line('');
-        $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. If you're going to use the database storage, you need to create a tenants table.");
+        $this->line("This package lets you store data about tenants either in Redis or in a relational database like MySQL. If you're going to use the database storage, you need to create tables for tenants and domains.");
```
drbyte: perhaps `for tenants and domains` should be `for tracking tenants and domains`?

I know the tables that the migrations will create are literally called `tenants` and `domains`, but when running `install` the new user of the package won't understand that yet, so adding the word `tracking` (or "storing"?) may add clarity about purpose.

stancl: I originally wanted to go with something like `you need to create a tenants table and a domains table` but that felt repetitive and I couldn't find anything that sounded better. Any idea how to shorten that sentence? :smile:

drbyte: Sure: "You need migrations. Say yes to install them."  :D 

Okay... Second sentence, this should suffice:
`For database storage we have provided some migrations.`

(They can study the migrations themselves if they want to. Most won't look into them, and won't need to anyway. Win win.)

stancl: Maybe something along the lines of:
> To store data about tenants in a relational database, you need a few database tables. Do you wish to publish the migrations that create these tables?

drbyte: Yes, that's fair.

PR #171, opened 2019-10-14
Update PreventAccessFromTenantDomains.php

drbyte:
Typo in comment

### DIFF
--- a/src/Middleware/PreventAccessFromTenantDomains.php
+++ b/src/Middleware/PreventAccessFromTenantDomains.php
@@ -8,7 +8,7 @@ use Closure;
 
 /**
  * Prevent access to non-tenant routes from domains that are not exempt from tenancy.
- * = allow access to central routes only from routes listed in tenancy.exempt_routes.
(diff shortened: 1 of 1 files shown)


src/Middleware/PreventAccessFromTenantDomains.php
```diff
@@ -8,7 +8,7 @@
 
 /**
  * Prevent access to non-tenant routes from domains that are not exempt from tenancy.
- * = allow access to central routes only from routes listed in tenancy.exempt_routes.
+ * = allow access to central routes only from routes listed in tenancy.exempt_domains.
```
stancl: Can you fix also the
> only from routes listed

stancl (changes requested):

src/Middleware/PreventAccessFromTenantDomains.php
drbyte: Ha! missed that one :)

stancl:
Decided to make it shorter because I couldn't find a way to keep it Laravel-styled :smile:

PR #180, opened 2019-10-16
Use config charset and collation for MySQL

jonagoldman:
https://github.com/stancl/tenancy/issues/172


stancl:
Great solution, thanks!

PR #186, opened 2019-10-19
allow for custom colums on creation of tenant

ryanhungate:
I ran into this while trying to add in a few custom columns on creation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ psysh
 phpunit_var_*.xml
 coverage/
 clover.xml
+.idea/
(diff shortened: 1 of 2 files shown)


ryanhungate:
@stancl i must be missing something small here - but I was just trying to show that during the `create` function of a new tenant, the custom columns are not addressed. In my local app that I originally started working on - I actually solved the problem by using an observer on the TenantModel to properly set the values from the data array into the model during create... I can delete this PR I just wanted to bring it up and it's failing the code coverage you're using.

stancl:
Hi, thanks for the bug report.

I'll see if I can make your PR work. If not, I'll open a new PR fixing this. I'll also add regression tests to make sure this isn't broken in the future again.

stancl:
Opened a PR with a bit cleaner code and regression tests: #191

PR #193, opened 2019-10-20
Creates route helper method on tenant

drfraker:
# What does this do?
This PR is helpful while testing your application. It is often useful to assert where a specific action redirects, and it was becoming cumbersome to try to create tenant routes each time.

# How is it used?
Imagine a test that is testing a tenant create route.
```
$response = $this->post(route('register'), $data);
// some assertions about the created user and tenant

// Assert that the controller redirected where it was supposed to
$response->assertRedirect(tenant()->route('/dashboard'));
```

## Quesions
- Is there already a way to do this that I'm missing?
- Is this even something you would want in the codebase?
- Since there can be more than 1 domain per tenant, is there a way the rest of the package decides which domain to use? Currently this code just grabs the first from the tenant's domains array.
- Is there another place this code would be more appropriate?

### DIFF
--- a/src/Tenant.php
+++ b/src/Tenant.php
@@ -385,14 +385,14 @@ class Tenant implements ArrayAccess
     }
     
      /**
-     * Get a route specific to this tenant.
+     * Generate a route specific to this tenant with the given path.
      *
      * @param String $path
      * @return string
      */
-    public function route($path)
+    public function route($path): string
     {
-        $path = $path[0] !== '/' ? '/'.$path : $path;
+        $path = $path[0] === '/' ?: sprintf('/%s', $path);
 
         $url = $this->app->config->get('app.url');
         $subdomain = $this->domains[0];
--- a/src/Tenant.php
+++ b/src/Tenant.php
@@ -383,6 +383,23 @@ class Tenant implements ArrayAccess
 
         return $result;
     }
+    
+     /**
+     * Get a route specific to this tenant.
+     *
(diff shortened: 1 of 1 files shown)


stancl:
Ideally I'd want to be able to do

```php
route('dashboard')->tenant($domain);
```

But `route()` returns a string, so your solution might be the best option we have.

I think this is related to the `TenantRedirect` feature, so maybe I could make `Tenant` macroable and we could register a macro in `TenantRedirect`?

I'm not sure about `Macroable` because of adding features in a backwards compatible way, though.

stancl:
`route()` returns a string, so we need a new helper. Probably something like `tenant_route($domain, $name, $parameters = [], $absolute = true)`. Going with `$domains[0]` is not ideal.

I'd be willing to accept a PR that adding that helper.

stancl:
```php
// tenant_route('foo'); // current tenant
// tenant_route('bar', ['a' => 'b']); // current tenant
// tenant_route('abc', [], 'anotherdomain.com');

function tenant_route(string $route, array $parameters = [], string $domain = null): string
{
    $domain = $domain ?? request()->getHost();

    // replace first occurance of hostname fragment with $domain
    $url = route($route, $parameters);
    $hostname = parse_url($url, PHP_URL_HOST);
    $position = strpos($url, $hostname);

    return substr_replace($url, $domain, $position, strlen($hostname));
}
```
I will probably include this in 2.3.0.

PR #199, opened 2019-10-22
Added Default TenancyTestCase

JapSeyz:


assets/tenancy_test_case.php.stub
```diff
@@ -0,0 +1,77 @@
...
+     * Create a new Tenant
+     *
+     * @param array $domains
+     * @return Tenant
+     */
+    protected function createTenant($domains = ['test.localhost']): Tenant
+    {
+        return Tenant::new()
+            ->withDomains($domains)
+            ->withData(['name' => 'test', 'host' => 'test'])
```
JapSeyz, line 46: Remove the extra data here

assets/tenancy_test_case.php.stub
```diff
@@ -0,0 +1,77 @@
...
+     * @return void
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        config(['database.default' => 'sqlite']);
+        config(['tenancy.database.suffix' => '.sqlite']);
+
+        $this->artisan('migrate');
```
JapSeyz, line 28: migrate:fresh as the system-connection is not recreated with each test.

Could also unlink and recreate it and then run migrate

Issue #173, opened 2019-10-14
Support Path Identification


rrpadilla:
#### Description
This great package **ONLY** support domain/subdomain identification.

Path identification would be identifying the tenant based on a part of the URL path, specifically the first section after the domain. An example of this would be 
```config 
https://myawesome.app/app/tenant-slug
``` 
where "**tenant-slug**" is the tenant identifier. 

Even we can use the domain too in case you want to reuse the domains table:
```config 
https://myawesome.app/app/tenant1.com
``` 
in this case "**tenant1.com**" is the tenant identifier. 

Basically we should be able to use: "**slug**", "**uuid**" or "**domain**".

#### Why this should be added
- You do not have to deal with domains/subdomains configuration and your customers do not have to change DNS records.
- **HTTPS supported** for every tenant **out of the box**. With domains/subdomains you and your customers have to configure and maintain multiple certificates.
- Vapor Support. See #74. Vapor does not support multiple second level domains.
- Really easy to test locally because you do not have to update the .hosts file.

Thanks!

rrpadilla:
@stancl will be great if domain, subdomain and path identification can be supported at the same time. Then your SaaS can have:
- basic plan (Only path identification. NO custom domain supported.)
- premium plan (custom domain supported)
Thanks!

stancl:
I think this is a good consideration given the things I've been changing about tenant identification lately.

The ideal scenario would be that both domain-based identification and path-based identification would work.

I think with 2.0.1, you should be able to use a route group with the InitializeTenancy middleware for this.

However, in 2.1.0 I'm changing the InitializeTenancy middleware to be executed globally. It initializes tenancy if the current domain is not a central domain, and then when the route middleware is being executed, the PreventAccessFromTenantDomains middleware checks that central routes can't be visited on tenant domains, and vice versa.

The reason I'm changing this is that it's the only way to initialize tenancy *before* running controller constructors.

So you *do* need some concept of a tenant domain.

However, I think you should be able to use a single subdomain not listed in the exempt_domains config, such as app.yoursaas.com and then inside you'd initialize tenancy manually inside a route group. Though the global middleware initializes tenancy using a domain, hmm. I think I'll make the middleware swappable.

Anyway. I think that using subdomains, such as tenant1.yoursaas.com is good enough functionality. But, if possible, I'd want a solution that lets you use path-based identification too.

stancl:
[...]

PR #207, opened 2019-10-27
Added request data identification middleware

JapSeyz:


src/Middleware/InitializeTenancyByRequestData.php
```diff
@@ -0,0 +1,63 @@
...
+                $this->parseTenant();
+            } catch (TenantCouldNotBeIdentifiedException $e) {
+                ($this->onFail)($e);
+            }
+        }
+
+        return $next($request);
+    }
+
+    protected function parseTenant()
```
stancl: Perhaps we could accept `Request $request` here from the `handle()` method and use that instead of the `request()` helper?

JapSeyz: Sure, any negatives to using the request helper?

stancl: It's unnecessarily calling the service container when you can just pass one variable

src/Middleware/InitializeTenancyByRequestData.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    protected function parseTenant(Request $request)
+    {
+        if (tenancy()->initialized) {
+            return;
+        }
+
+        $tenant = null;
+        if ($request->hasHeader('X-Tenant')) {
```
viezel: change to a config value as not all wants a `X-Tenant` in their app.

JapSeyz: Alright, also changed the possible value for the query parameter

assets/config.php
```diff
@@ -88,6 +88,10 @@
     'storage_to_config_map' => [ // Used by the TenantConfig feature
         // 'paypal_api_key' => 'services.paypal.api_key',
     ],
+    'identification' => [
+        'header' => 'X-Tenant', // Can be anything, but should really start with "X-",
+        'query_parameter' => '_tenant'
```
antonkomarev, line 93: Do we really need `_tenant` as default query parameter? I saw `tenant` query parameter in couple of real projects before and there were no issues with it.

src/Middleware/InitializeTenancyByRequestData.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    protected function parseTenant(Request $request)
+    {
+        if (tenancy()->initialized) {
+            return;
+        }
+
+        $header = config('tenancy.identification.header');
+        $query = config('tenancy.identification.query_parameter');
```
stancl, line 50: This should use dependency injection (`Illuminate\Contracts\Config\Repository`).

assets/config.php
stancl: I like the prefix because you will never run into an issue when sending data in your e.g. Vue components. You just set `_tenant` globally for axios.

src/Middleware/InitializeTenancyByRequestData.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $tenant = null;
+        if ($request->hasHeader($header)) {
+            $tenant = $request->header($header);
+        } elseif ($request->has($query)) {
+            $tenant = $request->get($query);
+        } elseif (! in_array($request->getHost(), config('tenancy.exempt_domains', []), true)) {
+            $tenant = explode('.', $request->getHost())[0];
```
stancl, line 58: I'm thinking if there are too many fallbacks in this middleware. I think I'm fine with the first two but I'd remove this one since this is request *data* identification MW, so it shouldn't care about the request *host*.

stancl:
I've left this open for some time but I think I'll merge this in 2.3.0.

Since we're adding identification MW, I might write one for path identification as well (think Slack) #173.

src/Middleware/InitializeTenancyByRequestData.php
antonkomarev: Isn't it two separate identification strategies?
1) by request header
2) by query path

stancl: It is. But I'd say that having two MW classes for that would be unnecessary.

My thinking is that "You'd either use the header, or you'd not. In which case you would use the request data."

Hmm, I wonder what's the use case for even that fallback. I think everyone can make their HTTP client set request headers.

@JapSeyz Are you using the fallback to request data in your app?

I'd say that we can remove all the fallbacks and just use the header. If that doesn't work for someone, they can copy the class, change it to use request data and update their config file.

Unexpected fallbacks, similarly to unexpected global state, can result in confusing debugging.

SergeyMiracle:
@stancl Hi, can you say when this PR will be merged? Don't wanna play with wildcard certificates :D

stancl:
@SergeyMiracle When I set aside some time to work on all the 2.3.0 features. The middleware is pretty simple and you can use it even before I merge anything. Just create your own middleware class, and configure the package to use it instead of the one it uses now:

https://github.com/stancl/tenancy/blob/06ee1ff0e28f131bf8a4e064cdd3e8a1c1b11293/assets/config.php#L107-L109

SergeyMiracle:
@stancl Ok, i'll try custom middleware for now. Thx for your reply 😃

SergeyMiracle:
@stancl One more thing with custom middleware - I made one for initialization with a header, but I need to disable `PreventAccessFromTenantDomains` and now idea how... help plz

`
`

stancl:
@SergeyMiracle Don't use `routes/tenancy.php` or the `tenancy` MW group on your routes, then.

stancl:
@JapSeyz I will make lots of changes to this but want you to get some credits for your work so I'll merge this into a wip branch and work there. Thank you.

PR #211, opened 2019-11-03
[2.2.0] Allow --path to be specified for migration commands

### LINKED ISSUE #202: Tenants Migrations
Hi, good job.
I'm testing the package into my core to Erp, the app are modular and i need make migration to tentants all core migrations and after for each module as select user to use and then run migration and your seed from multiples locations depending the module selected, but cmd tenants:migration not take params --path

Command: php artisan tenants:migrate --path=Modules/ModuleName/Database/Migration

image

stancl:
Closes #202

### DIFF
--- a/src/Traits/DealsWithMigrations.php
+++ b/src/Traits/DealsWithMigrations.php
@@ -8,6 +8,10 @@ trait DealsWithMigrations
 {
     protected function getMigrationPaths()
     {
+        if ($this->input->hasOption('path')) {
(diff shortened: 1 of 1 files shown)


stancl:
The failing build seems to be related to https://github.com/laravel/framework/issues/30505 in some way. Not sure why this is happening only in this branch and not in `2.x`, but I could reproduce similar behavior in a Laravel app using duplicate (central & tenant) migrations for queues.

Issue #100, opened 2019-08-16
Add more tests


stancl:
- [x] ~For custom columns with DB storage driver~ **Fixed:** #122
- [ ] For Telescope
- [x] ~For conflicting routes~ #114 conflicting routes are unsupported
- [ ] For Horizon?
- [ ] For queues
- [ ] Application events
- [ ] Tenant aware commands

stancl:
Irrelevant, codebase is mostly rewritten.

PR #221, opened 2019-11-08
[2.2.0] Tenant aware commands

### LINKED ISSUE #218: TenantAware Commands
#### Description
Often when running a SaaS, we need to run tenant specific commands. Having a trait or BaseCommand to extent - or even magical automatic setup. 

#### Why this should be added
I know you have the `tenancy()->all($this->option('tenants'))->each(...)` syntax in place, but I will suggest to have a forced command with only one tenant. do to the fact that if you forget the argument, then it runs the command on all tenants - which can lead to incorrect operations.

I suggest

```php
tenancy()->only($this->option('tenant'))->run(function ($tenant) {
    // call the command 
});
```

and a new `HasATenantOption` that is required. 

```php
trait HasATenantOption
{
    protected function getOptions()
    {
        return array_merge([
            ['tenant', null, InputOption::VALUE_REQUIRED, '', null],
        ], parent::getOptions());
    }
}
```

I can do a PR, but still a bit new to the project in terms of how to do those magical setups. so I will let you decide how to proceed.

stancl:
Resolves #218 

The `TenantAwareCommand` trait has an abstract method to get an array of the tenants for which the command should be executed. This method is implemented by both `HasATenantsOption` and the newly introduced `HasATenantOption`, so you can use these traits like this:

```php
class MyCommand extends Command
{
    use TenantAwareCommand, HasATenantOption;

    public function handle()
    {
        //
    }
}
```

TODO:
- [ ] ~Tests~ #100
- [x] Docs docs:39


viezel:
👏

PR #229, opened 2019-11-20
[2.3.0] tenant_route() helper

stancl:
```php
tenant_route('foo'); // current hostname
tenant_route('bar', ['a' => 'b']); // current hostname
tenant_route('abc', [], 'anotherdomain.com');
```

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 use Stancl\Tenancy\Tenant;
 use Stancl\Tenancy\TenantManager;
 
-if (! \function_exists('tenancy')) {
+if (! function_exists('tenancy')) {
     /** @return TenantManager|mixed */
     function tenancy($key = null)
     {
@@ -17,7 +17,7 @@ if (! \function_exists('tenancy')) {
     }
 }
 
-if (! \function_exists('tenant')) {
+if (! function_exists('tenant')) {
     /** @return Tenant|mixed */
     function tenant($key = null)
     {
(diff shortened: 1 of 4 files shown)


src/helpers.php
```diff
@@ -29,24 +29,38 @@ function tenant($key = null)
...
-if (! \function_exists('global_cache')) {
+if (! function_exists('global_cache')) {
     function global_cache()
     {
         return app('globalCache');
     }
 }
+
+if (! function_exists('tenant_route')) {
+    function tenant_route(string $route, array $parameters = [], string $domain = null): string
```
viezel, line 55: Strict mode improvement: 

Please specify that `$domain` can be nullable. 

```php
..., ?string $domain = null)
```

stancl: This is not needed. Setting the default value to null automatically makes the type nullable.

ahmedsayedabdelsalam:
I read the docs. if i need to store subdomain i need to store `foo` in domain column in domains table right? not the full domains structure.

but when i use `tenant_route('foo', 'dashboard')` it returns `https://foo/dashboard`  not `https://foo.example.test/dashboard`

dylanBcp:
+1 Same problem as @ahmedsayedabdelsalam

dylanBcp:
If you want to redirect to other subdomain use : 

```php
    $domain = 'foo';
    $domain = Domain::domainFromSubdomain($domain);
    tenant_route($domain, 'dashboard')
```

stancl:
`tenant_route()` expects hostname. It can't build a URL by combining the tenant's `domain` (subdomain) with a central domain, as there may be an unlimited number of central domains.

In setups that have more strictly defined how central domains work, e.g. by limiting them to one domain like our boilerplate, you can use things like `domainFromSubdomain()` (which itself is a boilerplate abstraction, not a Tenancy abstraction) yeah.

Issue #242, opened 2019-12-10
double database when using custom server connection


hajrovica:
Hi, question if i may,
when adding new tenant to application with **custom _tenancy_db_connection**  database for tenant is created on default connection server - not on server specified via **_tenancy_db_connection**
Database is empty - no seeding enabled. If i manually delete that database on primary connection i am getting error Database ... does not exist.

Is this duplication intended behaviour or am i missing something?

thank you in advance

stancl:
Hi.

So the issue is that even though `_tenancy_db_connection` is respected for tenant identification, it is not respected for DB creation and the default central connection is used?

hajrovica:
Hi, 
did more testing, so to clarify:

- during tenant creation `_tenancy_db_connection` is specified - which is pointing to different server
- database for tenant is created using central default connection,  `_tenancy_db_connection` is ignored
- database is not created on  `_tenancy_db_connection`
- running migrations for tenant  at this point will throw error because there is no database on  `_tenancy_db_connection` server, i have added database manually and migrations passed
- tenant identification is going through central default connection. So default server has empty tenant database an the tenant identification is working i can open tenant app. 
- If i delete tenant DB at default server tenant identification fails ignoring the fact that tenant connection points to different server

In short database creation and tenant identification are ignoring  `_tenancy_db_connection` while tenant app, if tenant is identified (empty tenant database on default connection), is working fine getting data from  `_tenancy_db_connection`

I hope i was able to explain it in sufficient detail, thanks in advance

stancl:
I see.

Seems like a bug. I'll try to fix it tomorrow & write regression tests to make sure this doesn't happen again.

Thanks for writing that up.

stancl:
Right, the issue is that for example the `MySQLDatabaseManager` uses the connection configured in `tenancy.database_manager_connections.mysql` instead of the tenant's connection.

stancl:
@hajrovica 

Could you try running

```
composer require stancl/tenancy:dev-respect-custom-connection
```

and seeing if that fixes the problem?

hajrovica:
@stancl issue fixed,  after running command, database is created on designated connection !

stancl:
Great! And does tenant identification work correctly too? Or does that still need fixing?

hajrovica:
Tenant identification working flawlessly, seeding successful, logged to tenant  - no errors 👍

stancl:
[...]

PR #243, opened 2019-12-10
Added support for migration parameters

jezzdk:
I added the support for migration parameters because I was facing an issue with running the migrations in production. In order to add the "--force" flag, I had to implement the same logic as for seeders.


stancl:
Looks great, thanks!

jezzdk:
@stancl  Thanks :) Can it go into a version 2.2.4 maybe?

stancl:
I think I'll be making a patch release to fix #242 today, so your PR will be included.

jezzdk:
Awesome! Thanks for doing a great job!

PR #276, opened 2020-01-23
Reinitialize tenancy for queued jobs if tenant id has changed

### LINKED ISSUE #274: Check that tenants are not persisted between jobs
There have been some reports that dispatching multiple queued jobs results in the first tenant being persisted for the following jobs in the queue worker state.

seanmtaylor:
Hi

I have put in the fix as mentioned in #274.

Previously the code would only initialise tenancy when processing a queued job if tenancy wasn't already initialised. 

This will now initialise the tenancy if the previously initialised tenant id is different to the `tenant_id` in the job payload.

--- 

Fixes #274

### DIFF
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -109,10 +109,12 @@ class TenancyServiceProvider extends ServiceProvider
 
         // Queue tenancy
         $this->app['events']->listen(\Illuminate\Queue\Events\JobProcessing::class, function ($event) {
-            if (array_key_exists('tenant_id', $event->job->payload())) {
-                if (! tenancy()->initialized) { // dispatchNow
-                    tenancy()->initialize(tenancy()->find($event->job->payload()['tenant_id']));
-                }
+            $tenantId = $event->job->payload()['tenant_id'] ?? null;
+            if (
+                $tenantId !== null &&
+                (! tenancy()->initialized || tenancy('id') !== $tenantId)
+            ) {
+                tenancy()->initById($tenantId);
             }
         });
     }
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -110,12 +110,20 @@ class TenancyServiceProvider extends ServiceProvider
         // Queue tenancy
         $this->app['events']->listen(\Illuminate\Queue\Events\JobProcessing::class, function ($event) {
             $tenantId = $event->job->payload()['tenant_id'] ?? null;
-            if (
-                $tenantId !== null &&
-                (! tenancy()->initialized || tenancy('id') !== $tenantId)
-            ) {
-                tenancy()->initById($tenantId);
+
+            // The job is not tenant-aware
+            if (! $tenantId) {
+                return;
             }
+
+            // Tenancy is already initialized for the tenant (e.g. dispatchNow was used)
+            if (tenancy()->initialized && tenant('id') === $tenantId) {
+                return;
+            }
+
+            // Tenancy was either not initialized, or initialized for a different tenant.
+            // Therefore, we initialize it for the correct tenant.
+            tenancy()->initById($tenantId);
         });
     }
 }


stancl:
So this makes both `queue:work` and `queue:listen` work as expected?

@viezel Can you check the original issue https://github.com/stancl/tenancy/issues/274 and this PR and see if this fixes your "persistent config between jobs" issue?

seanmtaylor:
> So this makes both `queue:work` and `queue:listen` work as expected?

@stancl Yes they are both working for me now. The fix ensures that tenancy is reinitialized if the `tenant_id` in the payload is different to the id of the initialized tenant.

viezel:
@stancl @seanmtaylor well I don't think this solves it. Hence, this design does not support the custom implementation developers create. An example would be **Feature Flags**. 
Imagine a developer creating feature flags in a SaaS app that sets feature for each tenant. Now the ` tenancy()->initById($tenantId);` does not reinitialize the custom feature flags because it does not know how to. If we don't reinitialize the feature flags, then the next queue job will have incorrect tenant settings. 


If we look into others that have the issue of resetting, then this package has defined a config to able the developer to add classes implementing an Interface to be reset on each reinitialize. https://github.com/swooletw/laravel-swoole/blob/master/config/swoole_http.php#L92-L94s
Therefore, I suggest that the solution will include a way for developers to define what would be reset by this package.

stancl:
Can you elaborate on feature flags? `initById()` and initializing using a tenant object are almost the same.

viezel:
@stancl it was just an example. What i'm trying to say is that whatever a developer chooses to develop on the "master side" of the SaaS needs to be reinitialized on each queue job. Typically tenant-specific config. 

In my platform, we have things like: CRM Oauth credentials, Email Templates, extensions enabled, extensions settings etc. All these are different from tenant to tenant and need to be reset before the next queue job gets processed.

stancl:
> whatever a developer chooses to develop on the "master side" of the SaaS needs to be reinitialized on each queue job

I don't see what you mean by this.

> In my platform, we have things like: CRM Oauth credentials, Email Templates, extensions enabled, extensions settings etc. All these are different from tenant to tenant and need to be reset before the next queue job gets processed.

Why would you need to reset anything? Those things should be stored in Tenant Storage. Since the Tenant instance that is bound in the container is changed when you initialize tenancy for a different tenant, you don't have to manually reset anything.

viezel:
I see... Maybe in but fully aware of the possibilities with Tenant Storage in this package. 
If that's the case - then, by all means, merge it in :)

stancl:
Even if you configure tenant storage-to-config bindings, they are reset after tenancy is ended/reinitialized, so there shouldn't be any issues with global state.

src/TenancyServiceProvider.php
```diff
@@ -109,10 +109,12 @@ public function boot(): void
...
-            if (array_key_exists('tenant_id', $event->job->payload())) {
-                if (! tenancy()->initialized) { // dispatchNow
-                    tenancy()->initialize(tenancy()->find($event->job->payload()['tenant_id']));
-                }
+            $tenantId = $event->job->payload()['tenant_id'] ?? null;
+            if (
+                $tenantId !== null &&
+                (! tenancy()->initialized || tenancy('id') !== $tenantId)
+            ) {
+                tenancy()->initById($tenantId);
```
stancl: This basically initializes tenancy if the job payload has a tenant_id and the tenant is different from the current tenant (to avoid re-initializing tenancy in case of `dispatchNow`, but to re-initialize tenancy if a new job for a different tenant should be executed), right.

I wonder if this could be refactored somehow. This takes me at least a minute to parse in my head which I don't like. But I can't think of a much better way to write this. Maybe something like this?

```php
$tenantId = $event->job->payload()['tenant_id'] ?? null;

if (! $tenantId) { // The job is not tenant-aware
    return;
}

if (tenancy()->initialized && tenant('id') === $tenantId) { // Tenancy is already initialized for the tenant (e.g. dispatchNow was used)
   return;
}

// Tenancy was either not initialized, or initialized for a different tenant. Therefore, we initialize it for the correct tenant.
tenancy()->initById($tenantId);
```

seanmtaylor: Completely agree @stancl. I've refactored the logic to be much easier to understand what's going on

Issue #301, opened 2020-02-27
Maintenance mode doesn't work on tenant domains


tamkeen-tms:
#### Describe the bug
As per Laravel's code:
```php
    public function isDownForMaintenance()
    {
        return file_exists($this->storagePath().'/framework/down');
    }
```
The check for the "down" file is done in the app storage path, which at this point, will be pointing to the tenant's storage path, which wont have the "down" file. 

So, when the app is taken down, only the central app will be down, not the tenants.

stancl:
Does putting
```php
\App\Http\Middleware\CheckForMaintenanceMode::class,
```
at the very top of your [`$middlewarePriority`](https://github.com/laravel/laravel/blob/6.x/app/Http/Kernel.php#L73) fix this?

stancl:
Hmm, seems like that does not work. I'll see what I can do.

stancl:
This is a bug but I'm removing it from the 2.3.0 milestone to ship the release faster.

stancl:
Fixed in v3.

PR #237, opened 2019-12-01
Add support for postgres schema

nuradiyana:
Hi. this is PR for #235

In this PR i introduce new config called `using_schema_connection` with default value is `false`. I tired to get `DatabaseManager` base on `driver` so i using this config to check if should change schema instance of database.

What do you think?

PR #304, opened 2020-03-04
Laravel 7 support

### LINKED ISSUE #303: laravel 7 compatibility issue
I will like for this package to support laravel 7

stancl:
This adds Laravel 7 support.

Resolves #303


stancl:
While all tests pass and the updated version constraints in composer.json should work with both Laravel 6 and Laravel 7, I can't merge this until @sebastianbergmann tags a new release of phpcov - https://github.com/sebastianbergmann/phpcov/pull/95#issuecomment-594753728

stancl:
I'll try using `dev-master` for phpcov, thanks for the suggestion @morloderex.

Dockerfile
```diff
@@ -4,27 +4,29 @@ LABEL maintainer="Samuel Štancl"
...
-    && apt-get install -y curl zip unzip git sqlite3 \
-    php7.2-fpm php7.2-cli \
-    php7.2-pgsql php7.2-sqlite3 php7.2-gd \
-    php7.2-curl php7.2-memcached \
-    php7.2-imap php7.2-mysql php7.2-mbstring \
-    php7.2-xml php7.2-zip php7.2-bcmath php7.2-soap \
-    php7.2-intl php7.2-readline php7.2-xdebug \
+RUN apt-get update
+
+RUN apt-get install -y software-properties-common
```
morloderex, line 8: Not sure but maybe we can combine these with the `&&` operator?

Something like:
```
RUN apt-get update -y && apt-get install -y software-properties-common && add-apt-repository -y ppa:ondrej/php
```

stancl: Good point

stancl: This is related to the line below too though. And I don't want to combine all 3, so I'll keep this the way it is. I cleaned up a different part of the file.

morloderex (approved):

stancl:
For some reason the testbench version constraint doesn't work when `composer require laravel/framework:^6.0` is used?

https://travis-ci.com/stancl/tenancy/jobs/294122914#L2232

.travis.yml
```diff
@@ -1,20 +1,20 @@
-# env:
-#   - LARAVEL_VERSION="^6.0" TESTBENCH_VERSION="~4.0" REDIS_DRIVER=phpredis
+env:
+  - LARAVEL_VERSION="^7.0"
```
morloderex, line 2:
```
env:
  - LARAVEL_VERSION="^6.0" TESTBENCH_VERSION="~4.0"
  - LARAVEL_VERSION="^7.0" TESTBENCH_VERSION="~5.0"
```

Might work (combined with the composer reqiure stuff down below) to resolve the issues related to https://github.com/stancl/tenancy/pull/304#issuecomment-594857474

stancl: I think it would work, but I wonder why the current solution doesn't work

stancl:
I might slightly delay this, but (in the next few days anyway) I'd like to make this part of the 2.3.0 release, now that #237 can be merged, which is also more semantically correct than making this a patch release.

stancl:
If this https://github.com/stancl/tenancy/issues/301#issuecomment-599201192 solves #301 then it should already work in L7 with the different installation steps. I'll have to add it to the L6 installation steps.

.travis.yml
stancl: Seems like this just started working?

PR #305, opened 2020-03-04
Support Laravel 7.x

morloderex:
Did initial work to get this package to run on laravel 7.x, e.g updated dependencies etc. 

~This is set as a draft because the package `phpunit/phpcov` (used in require-dev) depends on `symfony/console` and there is not yet a released package with "5.x" support yet.~

I have also only bumped the dependencies to laravel 7.x no actual code changes have been done sense i don't believe such changes is needed for this update. 

Because of the use of `phpunit/phpunit` "^9.0" i had to change our docker image to PHP 7.3 instead of 7.2

Edit: Added support for laravel 6.x back in.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,27 +4,30 @@ LABEL maintainer="Samuel Štancl"
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+RUN apt-get update && apt-get install -y software-properties-common
+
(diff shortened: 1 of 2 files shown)


stancl:
I already updated #304 with your `dev-master` suggestion for phpcov. Thanks, though!

PR #316, opened 2020-03-13
[2.3.0] Cached tenant lookup

### LINKED ISSUE #215: Cached tenant db lookup
#### Description

Today on each request we have 3 queries to the db:
- `select * from domains where domain = ? limit 1`
- `select * from tenants where tenants.id = ? limit 1`
- `select * from domains where tenant_id = ?`

This is quite fine for development, but will become a bottleneck as many request to db does not scale well. Instead using a key value store like redis will scale and can easily withhold this information.

#### Why this should be added
1. use the `globalCache` to store the those values (typically in redis) and add this cache check to the init process.
2. use tagging as Tenants can be changed/deleted and therefore cache invalidation on a specific tag.
3. (optional) - if developers have multiple tables related to the tenant in the "central" database, it could be performant to eager load those relationships and cache those. Hence, create a new config array to specify relationships what will go into a `->with([...])` statement.

stancl:
Resolves #215 

- [x] Finish tests
- [x] Finish the CachedTenantResolver class (assuming tests are failing)
- [x] Check that tests pass
- [x] Remove config from TestCase.php

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -16,6 +16,8 @@ return [
                 'tenants' => 'tenants',
                 'domains' => 'domains',
             ],
+            'cache_store' => false, // What store should be used to cache tenant resolution. Set to false to disable cache, null to use default store, or a string with a specific cache store name.
+            'cache_ttl' => 3600, // seconds
         ],
         'redis' => [
             'driver' => Stancl\Tenancy\StorageDrivers\RedisStorageDriver::class,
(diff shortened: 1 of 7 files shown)


stancl:
@viezel If you're interested, you can review.

The domain to tenant id mapping is cached and the tenant id to data & domains mapping is cached, resulting in `findByDomain()` and `findById()` being cached after the original query. Cache invalidation is covered by tests.

Seems like a pretty good solution to me. Avoids making those extra queries on a separate DB connection on each request.

viezel:
👏🏻👏🏻

PR #323, opened 2020-03-16
[2.3.0] More identification middleware

stancl:
- [x] Request data
- [ ] ~~Path~~


stancl:
I wanted to add path identification middleware too but that is probably better done just like this:

```php
Route::prefix('/tenant/{tenant}')->group(function () {
    // ...
});
```

With
```php
Route::bind('tenant', function ($id): Tenant {
    try {
        return tenancy()->find($id);
    } catch (TenantDoesNotExistException $e) {
        abort(404);
    }
});
```

PR #333, opened 2020-03-20
Add return value to execute method

devonmather:
Fixes the following error as demonstrated with test in PR and error below. Occurs after updating to Laravel 7.x and Tenancy 2.3

```
TypeError: Return value of "Stancl\Tenancy\Tests\Etc\AddUserCommand::execute()" must be of the type int, NULL returned.
```

Let me know if further changes are needed ✌🏻

### DIFF
--- a/src/Traits/TenantAwareCommand.php
+++ b/src/Traits/TenantAwareCommand.php
@@ -26,6 +26,8 @@ trait TenantAwareCommand
                 $this->laravel->call([$this, 'handle']);
             });
         }
+
+        return 1;
     }
 
     /**
--- a/tests/Etc/AddUserCommand.php
+++ b/tests/Etc/AddUserCommand.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Tests\Etc;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Str;
+use Stancl\Tenancy\Traits\HasATenantsOption;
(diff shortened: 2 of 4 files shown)


stancl:
Perfect quality PR, thank you so much! Will tag a release with this and one other bug fix tomorrow.

src/Traits/TenantAwareCommand.php
```diff
@@ -26,6 +26,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 $this->laravel->call([$this, 'handle']);
             });
         }
+
+        return 1;
```
stancl, line 30: @devonmather Actually, why are we returning `1` here? What's the reason for this and why `1` and not `0`?

devonmather: The reason is it must return an integer value, I think it's related to the latest symfony component for commands. Upon further reading just now. I misunderstood the error and figured a 0 would be if the command failed. But upon further reading it looks like the default return value is zero. I can created another PR this evening to fix.

stancl: @devonmather Seems like Laravel casts the return value from `handle()` into an int and returns that. See this: https://github.com/laravel/framework/blob/7.x/src/Illuminate/Console/Command.php#L134

I guess most `handle()` methods return nothing, so null, which is cast into `0` - as success? And then the `handle()` method can optionally return for example `1` as a specific exit code? I'll open a PR and ask you for review, if you'd be so kind. I don't have that much experience with commands.

PR #336, opened 2020-03-24
[2.3.1] Fix exit codes in TenantAwareCommand

stancl:
Follow-up to #333

@devonmather Please review.

If there is only one tenant, we return the command's exit code. If there are multiple tenants, we return the latest non-0 exit code.

### DIFF
--- a/src/Traits/TenantAwareCommand.php
+++ b/src/Traits/TenantAwareCommand.php
@@ -10,7 +10,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 trait TenantAwareCommand
 {
-    /** @return mixed|void */
+    /** @return int */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $tenants = $this->getTenants();
@@ -21,13 +21,18 @@ trait TenantAwareCommand
             });
         }
 
+        $exitCode = 0;
         foreach ($tenants as $tenant) {
-            $tenant->run(function () {
-                $this->laravel->call([$this, 'handle']);
+            $result = $tenant->run(function () {
+                return $this->laravel->call([$this, 'handle']);
             });
+
+            if ($result !== 0) {
+                $exitCode = $result;
+            }
         }
 
-        return 1;
+        return $exitCode;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stancl:
I assume this will break your tests @devonmather. I'll wait for CI and then update them.

morloderex:
@stancl Whouldn't it be better to break off the loop if 

```php
$result = (int) $tenant->run(function () {
    return $this->laravel->call([$this, 'handle']);
});
```
returns a status that's not `0`. 

Let me explain my believe why:
Let's assume we have 3 tenants in our loop, and the second one inside the loop returned with a status of `1`. Then it might be better to hold off any futher execution of the script because i would assume that this is because of an exception has been thrown. 

Because even though the first tenant finished correctly the second one throw an exception?

src/Traits/TenantAwareCommand.php
```diff
@@ -21,13 +21,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $exitCode = 0;
         foreach ($tenants as $tenant) {
-            $tenant->run(function () {
-                $this->laravel->call([$this, 'handle']);
+            $result = (int) $tenant->run(function () {
+                return $this->laravel->call([$this, 'handle']);
             });
+
+            if ($result !== 0) {
+                $exitCode = $result;
```
morloderex:
```suggestion
if($result !== 0) {
    $exitCode = $result;
    break;
}
```

src/Traits/TenantAwareCommand.php
```diff
@@ -21,13 +21,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            }
```
morloderex, line 32:
```suggestion
            if ($result !== 0) {
                $exitCode = $result;
                break;
            }
```

See https://github.com/stancl/tenancy/pull/336#issuecomment-603430012

stancl:
I think exceptions would "bubble up". This is just exit codes, for things like unsuccessful DB inserts I'd say. And the command can fail for some tenants but succeed for others. It's tricky to find the ideal solution when dealing with multiple tenants in a loop.

I don't know too much about commands and exit codes in Laravel, so if anyone can give me some examples I'd be grateful.

devonmather:
What you have there looks like a more complete version of what my PR should have looked like. I've ran the tests on this branch, looks good to me 👌🏻

stancl:
I guess we can always improve the exit code logic later.

PR #348, opened 2020-04-10
Make config more understandable

stancl:
Originally, I wanted to keep the config relatively comment-free, to avoid creating the impression that there's a lot to configure by the long scrolling.

However, as the feature set grew, some things in the config are really unclear unless you take the time to understand the concepts in the documentation.

Ideally, people would read the docs, and then be reminded of how things work by these simplified explanations of the features.

I'd appreciate any feedback with regards to grammar and how well I documented these things.

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -13,11 +13,11 @@ return [
          * The majority of applications will want to use this storage driver.
          * The information about tenants is persisted in a relational DB
          * like MySQL or PostgreSQL. The only downside is performance.
-         * 
+         *
          * A database connection to the central database has to be established on each
          * request, to identify the tenant based on the domain. This takes three DB
          * queries. Then, the connection to the tenant database is established.
-         * 
+         *
          * Note: From v2.3.0, the performance of the DB storage driver can be improved
          * by a lot by using Cached Tenant Lookup. Be sure to enable that if you're
          * using this storage driver. Enabling that feature can completely avoid
@@ -32,7 +32,7 @@ return [
 
             /**
              * Your central database connection. Set to null to use the default one.
-             * 
+             *
              * Note: It's recommended to create a designated central connection,
              * to let you easily use it in your app, e.g. via the DB facade.
              */
@@ -45,7 +45,7 @@ return [
 
             /**
              * Here you can enable the Cached Tenant Lookup.
-             * 
+             *
              * You can specify what cache store should be used to cache the tenant resolution.
              * Set to string with a specific cache store name, or to null to disable cache.
              */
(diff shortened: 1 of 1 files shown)


assets/config.php
```diff
@@ -3,113 +3,287 @@
...
     'features' => [
-        // Features are classes that provide additional functionality
-        // not needed for tenancy to be bootstrapped. They are run
-        // regardless of whether tenancy has been initialized.
-
-        // Stancl\Tenancy\Features\Timestamps::class,
-        // Stancl\Tenancy\Features\TenantConfig::class,
-        // Stancl\Tenancy\Features\TelescopeTags::class,
-        // Stancl\Tenancy\Features\TenantRedirect::class,
+        // Stancl\Tenancy\Features\Timestamps::class, // https://tenancy.samuelstancl.me/docs/v2/features/timestamps/
```
rmsj, line 226: There is a hidden setting on the code base `timestamp_format` that could be mentioned here - the thing is that if only makes sense if you enable the Timestams feature.

stancl: Don't want to bloat the config too much, so I think I'll just mention this on the docs page. https://tenancy.samuelstancl.me/docs/v2/features/timestamps/#timestamps-redirect

drfraker: I agree that the config file being "clean" like you suggest here is great. But having config options scattered throughout the docs has its downsides as well. I sometimes wish documentation sites had a config example with every option and clear explanations of what each option does. This has the benefit of seeing all options available with a quick explanation (and maybe a link to the section of the docs for more detail) while still having a clean config file in the codebase.

How would you feel about something like that?

drfraker:
Also just to throw my latest hidden config setting I found yesterday into the ring: `tenancy.migration_parameters` is available for us to set migration parameters when automatic migrations run after tenant creation. 

I wrestled with a bug for a few hour on my production site because I have the option to run migrations after tenant creation set to true, but in production you need the `--force` option in order to migrate.

stancl:
@drfraker That would probably be a better default for production, but to maintain backwards compatibility I don't want to change that.

I'll add a note that it should be set to `true` in production.

drfraker:
> @drfraker That would probably be a better default for production, but to maintain backwards compatibility I don't want to change that.
> 
> I'll add a note that it should be set to `true` in production.

Hey Samuel. I wasn’t suggesting that you set that to true by default, I have an older version of the config file and it didn’t include that setting. Therefore I didn’t even know it existed. My fault for not reading release notes probably. But I still think a config file with all settings and explanations could be helpful in the docs.

PR #377, opened 2020-04-28
[WIP] Database user per tenant - original draft

stancl:
WIP. Needs to be reviewed and logic for deleting users needs to be added. Also some minor style changes.

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -108,8 +108,6 @@ return [
          */
         'prefix' => 'tenant',
         'suffix' => '',
-
-        'separate_by' => 'database', // database or schema (only supported by pgsql)
     ],
 
     /**
--- a/src/Contracts/ManagesDatabaseUsers.php
+++ b/src/Contracts/ManagesDatabaseUsers.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Stancl\Tenancy\Contracts;
 
 use Stancl\Tenancy\Contracts\Future\CanSetConnection;
--- a/src/Contracts/ManagesDatabaseUsers.php
+++ b/src/Contracts/ManagesDatabaseUsers.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Stancl\Tenancy\Contracts;
+
+use Stancl\Tenancy\Contracts\Future\CanSetConnection;
+use Stancl\Tenancy\Tenant;
+
+interface ManagesDatabaseUsers extends CanSetConnection
+{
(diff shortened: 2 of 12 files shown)


viezel:
@stancl I would recommend to add a `name` string column to the Tenant db table and use this as a human identifier throughout the app.

Imagine you have 200 customers, each having a `tenant` ´a UUID in the database name. This is super hard to find. see here:

image

Instead having a human readable name is much easier. 

* tenant_cocacola
* tenant_google
* tenant_dell
* tenant_apple


# How to

1. Create a new tenant with name

```php
Tenant::new()
    ->withDomains(['cocacola.myapp.test'])
    ->withName('Coca Cola')
    ->save();
```

2. generate the db name with random chars in the end

```php
// Tenant line 279
public function getDatabaseName(): string
{
    return $this->data['_tenancy_db_name'] ?? 
        (
            $this->config->get('tenancy.database.prefix') . 
            $this->name . 
            $this->config->get('tenancy.database.suffix') . 
            Str::random(4)
        );
}
```


This will have a bonus effect on all your Artisan commands as well

Much easier to do: `php artisan tenants:migrate --tenants=apple`

stancl:
@viezel Do you think the current way of doing this is okay too? You can specify some tenant `id` that will be used in place of the uuid (uuids are generated just when no specific ID is passed). So you'd do

```php
Tenant::new()->withData(['id' => 'apple'])->save();
```

And their DB would be named `tenantapple` with the default config.

https://github.com/stancl/tenancy/blob/4ff82a950d88eff72edf598ff0244eff6bb229b1/assets/config.php#L105-L110

stancl:
But since we're adding a lot of DB specific keys, I guess we might as well add a DB name key too, that would override the entire `prefix + id + suffix` logic? Yeah, probably makes more sense.

viezel:
oh, didn't know of `Tenant::new()->withData(['id' => 'apple'])->save();` will that also work with artisan commands?

viezel:
I would still vote for a dedicated db table column for Name. It can be used for many things, like Admin panels, transactional emails etc. 

What we currently do is: 

```php
'tenant_' . Str::snake($tenant->getAttribute('name')) . Str::random(4)
```

This way you can have a Tenant Company name like "Coca Cola Inc".

stancl:
Keeping this open for reference, but rewrote this from ground up because there's some refactoring to be done & breaking changes are very hard to avoid. So we're aiming for 3.x.

#382

goodmartian:
Hello. When will this feature be added?

stancl:
https://github.com/stancl/tenancy/blob/3.x/assets/config.php#L65

PR #382, opened 2020-04-30
[3.x] DB users

stancl:
Ground-up rewrite of #377


src/Contracts/TenantDatabaseManager.php
```diff
@@ -6,6 +6,13 @@
 
 interface TenantDatabaseManager
 {
+    /**
+     * Return the config key that separates databases (e.g. 'database' or 'schema').
+     *
+     * @return string
+     */
+    public function getSeparator(): string;
```
stancl, line 16: We don't need a weird `separate_by` config now. We just use this to know what key to change in the connection config array.

src/DatabaseConfig.php
```diff
@@ -0,0 +1,169 @@
...
+
+    public static function generateUsernameUsing(callable $usernameGenerator): void
+    {
+        static::$usernameGenerator = $usernameGenerator;
+    }
+
+    public static function generatePasswordUsing(callable $passwordGenerator): void
+    {
+        static::$passwordGenerator = $passwordGenerator;
+    }
```
stancl: These methods are used to customize the logic used for generating DB names, usernames and passwords.

They should probably be pluralized? E.g. `generateDatabaseNamesUsing()`

src/DatabaseConfig.php
```diff
@@ -0,0 +1,169 @@
...
+        if (isset($dbConfig['_tenancy_db_name'])) {
+            unset($dbConfig['_tenancy_db_name']);
+        }
+
+        return array_reduce($dbConfig, function ($config, $key) {
+            return array_merge($config, [
+                Str::substr($key, 0, strlen('_tenancy_db_')) => $this->tenant[$key],
+            ]);
+        }, []);
+    }
```
stancl: This lets you store (for whatever reason), e.g.

```php
'_tenancy_db_charset' => 'foo'
```
in the tenant storage and it will be part of the config.

src/DatabaseManager.php
```diff
@@ -89,53 +80,14 @@ public function setDefaultConnection(string $connection)
...
-
-    /**
-     * Get the driver of a database connection.
-     *
-     * @param string $connectionName
-     * @return string|null
-     */
-    public function getDriver(string $connectionName): ?string
-    {
-        return $this->app['config']["database.connections.$connectionName.driver"];
```
stancl, line 131: This class was the absolute biggest mess of the package and I'm glad we refactored a lot of this logic into the `DatabaseConfig` class. Still doesn't feel perfect to me, but a lot better than what we had before in my opinion.

src/DatabaseManager.php
```diff
@@ -170,101 +120,75 @@ public function ensureTenantCanBeCreated(Tenant $tenant): void
...
+            $this->createDatabaseAsynchronously($tenant, $afterCreating);
         } else {
-            $manager->createDatabase($database);
-            foreach ($afterCreating as $item) {
-                if (is_object($item) && ! $item instanceof Closure) {
-                    $item->handle($tenant);
-                } else {
-                    $item($tenant);
-                }
-            }
```
stancl: Some more cleanup for more readable methods.

src/Tenant.php
```diff
@@ -272,23 +272,13 @@ public function softDelete(): self
...
-     * Get the tenant's database's name.
+     * Get database config.
      *
-     * @return string
+     * @return DatabaseConfig
      */
-    public function getDatabaseName(): string
+    public function database(): DatabaseConfig
     {
-        return $this->data['_tenancy_db_name'] ?? ($this->config->get('tenancy.database.prefix') . $this->id . $this->config->get('tenancy.database.suffix'));
```
stancl, line 281: This should technically be `_tenancy_db_database` (to match the other keys with relation to the DB connection config arrays), but that would be a lot less memorable and break BC. So I kept the logic, but moved the method to `DatabaseConfig`.

stancl:
Ping @viezel.

Added comments for some relevant changes.

src/DatabaseManager.php
```diff
@@ -170,101 +120,75 @@ public function ensureTenantCanBeCreated(Tenant $tenant): void
...
+            QueuedTenantDatabaseDeleter::dispatch($manager, $tenant);
+        } else {
+            $manager->deleteDatabase($database);
+            if ($manager instanceof ManagesDatabaseUsers) {
+                $manager->deleteUser($tenant->database());
+            }
         }
 
-        return 'database';
+        $this->tenancy->event('database.deleted', $database, $tenant);
```
viezel, line 182: Don’t you need to wait for the queued job to be completed before dispatching this event?

viezel (commented):
Really nice work. Great structure. Clean code 👏🏻

I did not see any transactions in the diff. Do we have a db traction of creating db and user, so it’s possible to roll back if error occurs

src/DatabaseConfig.php
viezel: The code makes it possible to use hooks instead of creating a complete new connection in the Laravel config/database.php config file. 
Do I understand this code correct?

src/TenantDatabaseManagers/PermissionControlledMySQLDatabaseManager.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\TenantDatabaseManagers;
+
+use Stancl\Tenancy\Contracts\ManagesDatabaseUsers;
+use Stancl\Tenancy\DatabaseConfig;
+
+class PermissionControlledMySQLDatabaseManager extends MySQLDatabaseManager implements ManagesDatabaseUsers
+{
+    public static $grants = [
```
viezel, line 16: I think it should be possible for a developer to overwrite grants. Some might to give even less grants and thereby more restrictive

stancl:
Added transactions here: https://github.com/stancl/tenancy/pull/382/commits/65ebc043dc68f8c7f7ee82aee9a2a116b3783924

src/TenantDatabaseManagers/PermissionControlledMySQLDatabaseManager.php
stancl: Since it's `public static`, it should be possible to do this, no?

```php
PermissionControlledMySQLDatabaseManager::$grants = [...];
```

viezel: yes and no. If its not a singleton, then that will not work. 
I would suggest it to be in the config

stancl: It's a static property, so it will work exactly like singletons.

viezel: oh its static. sure that will work :D

viezel: but.. is that intuitive ? 
I mean, if I don't want to go through the entire source code to understand that I can overwrite this. At least, that needs to be documented for people to get that

stancl: Yeah, I'll add it to the docs. I also mentioned it in the config: https://github.com/stancl/tenancy/pull/382/files#diff-dd3b63ca90b9a9aac58961a8913d1f8fR199

Since these classes are optional, and you can use anything in place of them, I really don't like the idea of adding config keys for this stuff. The config should have things that are always in effect, rather than config for some classes that may not even be used. I will make a similar change to other things since I'm doing v3.

src/DatabaseConfig.php
stancl: That was possible before too, with hooks. Now you can have custom connection settings without hooks.

Before, you'd have a e.g. `tenant` connection and you'd override specific keys in that config by doing something like this:

```php
tenancy()->hook('bootstrapping', function ($tm, $tenant) {
    config(['database.connections.tenant.charset' => $tenant->data['_tenancy_db_charset']]);
});
```

src/DatabaseManager.php
stancl: That's a fair point. This got me thinking that I can maybe get rid of this clumsy `afterCreating` logic, and just make people use the `created` event for stuff that should be executed after the database is actually created? I'd trigger this event inside the queued job, and it would even make more sense because now it's triggered before the job is executed (after it's dispatched):

https://github.com/stancl/tenancy/blob/a39d961cc3e21721fe99c6410cccb9c213675ed4/src/DatabaseManager.php#L129-L136

The logic currently looks like this:

https://github.com/stancl/tenancy/blob/a39d961cc3e21721fe99c6410cccb9c213675ed4/src/DatabaseManager.php#L138-L164

So instead of dealing with queue chains and other complex stuff like that, I'm thinking that I could just dispatch this event inside the `QueuedTenantDatabaseCreater` job.

And of course, the same would apply for deleting DBs.

stancl:
I'll focus on getting the tests passing now, then I'll refactor the event stuff mentioned above, and then I'll write tests for the new behavior (DB users, grants, `DatabaseConfig` data structure).

Then we can merge.

stancl:
Okay, it's 4 AM and old tests are passing 🙃. Will write new tests tomorrow & merge.

src/DatabaseManager.php
stancl: Will do this as part of another PR, to avoid making this one too large.

viezel: Great idea. Stay green with new test and merge in 👍🏻

stancl:
@viezel Please review, this should be the final version.

Edit: Damn that's a large diff, 42 changed files 😄. If you want to review it quickly, just look at the `DatabaseUsersTest`, `PermissionControlledMySQLManager` and `DatabaseConfig`.

PR #392, opened 2020-05-04
[3.x] Add gitattributes

viezel:
Improve the distribution size of this package, as there is no reason to ship tests, github workflow and various test files with composer.


viezel:
@stancl no idea why it thinks code coverage went down, as I did not add any executable code in this PR. 🤷🏻‍♂️

stancl:
Makes sense, good idea!

PR #394, opened 2020-05-06
[3.x] - Add a hook to console commands that should run before trying to fetch/initialize tenants

morloderex:
# Introduction
This pull request adds the ability to run arbitary code before any tenants have been located and initialized. 

## Why
Imagine you have to create a confirmation dialog/ or some other condition which states if the command should run. Given the current implementation of `TenantAwareCommand` trait. If you implement that logic with the `handle` method inside the command that condition would previously run for every single tenant. 
This can in some situations be the desired behaviour and in other cases it's a termination of the command entirely.

I hope the point of this is getting across, if not i'll be happy to provide extra any extra information needed.

### DIFF
--- a/src/Traits/TenantAwareCommand.php
+++ b/src/Traits/TenantAwareCommand.php
@@ -13,6 +13,12 @@ trait TenantAwareCommand
     /** @return int */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $callback = $this->runBeforeRunningTenantsHook();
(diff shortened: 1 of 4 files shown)


stancl:
Name it `beforeExecuting()` or something like that, make it part of the trait with `return 0;` as the normal behavior, removing the need for a `method_exists()` check.

stancl:
See suggestion above.

PR #409, opened 2020-05-22
Updated some docker stuff

ohnotnow:
Changed the dockerfile to use the official Docker php image and installed
the pcov extension instead of xdebug for code coverage.

Also updated the docker-compose file to make the main 'test' service not start up
before all the DB containers are running ok.


stancl:
Beautiful, thank you very much.

PR #452, opened 2020-07-06
IDE Error Fix

CristianTudorache:


stancl:
The annotations were there for a reason (Intelephense support).

Can you fix their signature to match `Model`? Instead of removing them.

Thanks!

stancl:
Thanks!

Is this ready to be merged now?

CristianTudorache:
I think it's ok, everything is green and code inspection couldn't find a problem.
Thank you.

PR #457, opened 2020-07-10
Improved query for resolving domain

JorgenSolli:
The tenant relation is instantly called 100% of the times this method is called.
Including the tenant relation, we are saving an additional database query.


stancl:
What's the performance benefit of this?

I understand the reasoning, since we're always accessing the tenant, but `with()` makes sense when there are multiple records. In this case, it just executes another query to get the tenant, just like the `$domain->tenant` would.

JorgenSolli:
My mistake. I thought the `with()` method had some inner join magic.
You can safely just deny this PR. Sorry for wasting your time with this :)

stancl:
No worries, appreciate the effort to help!

PR #461, opened 2020-07-14
Use the Tenant models' getTenantKeyName() value

timothydc:
Don't assume the Tenant can be found with the `id` attribute. The Tenant model allows for setting a different Tenant key name through `getTenantKeyName()`, so lets use it to find the Tenant.

### DIFF
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -80,7 +80,7 @@ class Tenancy
 
     public function find($id): ?Tenant
     {
-        return $this->model()->find($id);
+        return $this->model()->where($this->model()->getTenantKeyName(), $id)->first();
     }
 
     /**


stancl:
Hey,

Appreciate the PR.

Note that `$id` is just the name of the variable. It's still using `find()` which uses the model's primary key.

I think this is correct. Am I missing something?

Thanks

timothydc:
In the default setup the `$tenants` variable in the `runForMultiple()` methods is an array of my tenant IDs. For example `['tenantFoo']`. Because the package is using the `find`-query, it will try and query the `Tenant` with the `id` attribute.

Now my setup is using path identification for tenants and adds an actual primary key to my `Tenant`-table. I moved the tenant ID to a `slug` column, so I am not using the primary key anymore. But now it will still try and query the `Tenant` with the `id` attribute. Because the package is using `$model->find($id)`.

To solve that issue I added this to my `BaseTenant` model:

```php
public function getTenantKeyName(): string
{
    return 'slug';
}
```

The `find`-query should use the custom `getTenantKeyName()`-value. Otherwise you can't customize the tenant identification column.

Hopefully my setup changes make sense. :)

stancl:
Ah that makes sense. So rather than using a custom primary key, you just want to use a custom **tenant identifier column**.

This makes sense, it's a separate concept from the Model's primary key.

Good catch, merging. Will tag a release once I fix the thing I've just broken in 3.x 😬

PR #467, opened 2020-07-27
Qualify scoped column

binaryk:


stancl:
What issue does this solve?

binaryk:
If you have for example 2 tables, and both have `tenant_id`, and you have some leftJoins, it doesn't know which belongs to which one. It says: `ambiguous tenant_id column`.

stancl:
Makes sense, thanks!

PR #473, opened 2020-08-04
Allows tenants to be initalized with their IDs

JorgenSolli:


stancl:
Thanks for the PR! Good quality, even a custom exception.

PR #479, opened 2020-08-18
reverse middlewares order before prepend

bepsvpt:
`prependToMiddlewarePriority` will add the middleware to first one. If it just iterates the `$tenancyMiddleware` array. The final middleware order will be incorrect.


stancl:
Thanks!

PR #488, opened 2020-09-04
Fix initialize() method

### LINKED ISSUE #481: Calls to initialize with a non object value only works the first time it's called
#### Describe the bug
The new feature to initialize the tenant manuall,  passing a string or int, only works the first time it's called. Subsequent calls will throw an error as $tenant->getTenantKey() does not exist.

https://github.com/stancl/tenancy/blob/1ff82fb2a09dd0ee545fd13679aa739ea589dfed/src/Tenancy.php#L34-L36

#### Steps to reproduce
Access the laravel console and try to initialize the tenancy more than once for the same or different tenant by its id:

```php
tenancy()->initialize('tenant1');
tenancy()->initialize('tenant2');
```

#### Expected behavior
The new tenant context would be loaded without throwing an error. In case of the id belongs to the current tenant, nothing should happen.

#### Your setup
 - Laravel version: 7.25.0
 - stancl/tenancy version: 3.1.7

stancl:
Fixes #481

### DIFF
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -31,6 +31,16 @@ class Tenancy
      */
     public function initialize($tenant): void
     {
+        if (! is_object($tenant)) {
+            $tenant = $tenant;
+            $tenantId = $tenant;
+            $tenant = $this->find($tenantId);
+
+            if (! $tenant) {
+                throw new TenantCountNotBeIdentifiedById($tenantId);
+            }
+        }
+
         if ($this->initialized && $this->tenant->getTenantKey() === $tenant->getTenantKey()) {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/Tenancy.php
```diff
@@ -31,6 +31,16 @@ class Tenancy
      */
     public function initialize($tenant): void
     {
+        if (! is_object($tenant)) {
+            $tenant = $tenant;
```
lcdss, line 35: This line it's not doing nothing and can be removed.

lcdss: This line is doing nothing and can be removed.

lcdss (commented):
I updated the package in my project and didn't find anything wrong. It's only missing a test for this feature.

src/Tenancy.php
stancl: Nice catch. So this all works now?

PR #489, opened 2020-09-06
Test Laravel 8 support

stancl:


stancl:
Dependencies won't install because virtualcolumn & jobpipeline require Laravel 6|7.

When L8 is stable, I'll add a version matrix to virtualcolumn & jobpipeline CI testing setup to make sure all 6, 7, and 8 are supported.

Once new versions of those packages are tagged, I can try re-running these tests.

So I'll get Laravel 8 support done as soon as it comes out.

stancl:
Closing in favor of #491

PR #491, opened 2020-09-08
First try at Laravel 8

andresayej:


stancl:
Verified that this works on a 7.x app.

PR #492, opened 2020-09-08
Support for Laravel 8

elkebirmed:


stancl:
See the other PR. We're working on this with @andresayej right now.

PR #493, opened 2020-09-15
Only bind UUID generator if one is specified in config

domkrm:
I would using autoincrement IDs for tenants. I changed the migration and set the `tenancy.id_generator` config to `null`. But when I tried to create a new tenant I got the error message:

`Illuminate/Contracts/Container/BindingResolutionException with message 'Target [Stancl/Tenancy/Contracts/UniqueIdentifierGenerator] is not instantiable.'`

This is because the `GeneratesIds` trait checks if an UUID generator is bound to determine if autoincrement should be used. But even if you bind a class with null Laravel interprets this as 'bound'.

This PR only binds the generator class if a generator is specified in the config.


stancl:
Thanks!

PR #499, opened 2020-09-22
Fix: The database name is displayed when the TenantDatabaseAlreadyExistsException exception is thrown.

GENL:
When TenantDatabaseAlreadyExistsException  was thrown, the reason did not display the concerned database name. This PR fixes that behavior.


stancl:
Thanks

PR #501, opened 2020-09-25
Fix Database Sessions

CristianTudorache:
Fix Database Sessions

### DIFF
--- a/src/Database/DatabaseManager.php
+++ b/src/Database/DatabaseManager.php
@@ -38,7 +38,7 @@ class DatabaseManager
      */
     public function connectToTenant(TenantWithDatabase $tenant)
     {
-        $this->database->purge('tenant');
+        $this->database->disconnect('tenant');
         $this->createTenantConnection($tenant);
         $this->setDefaultConnection('tenant');
     }
@@ -49,7 +49,7 @@ class DatabaseManager
     public function reconnectToCentral()
     {
         if (tenancy()->initialized) {
-            $this->database->purge('tenant');
+            $this->database->disconnect('tenant');
         }
 
         $this->setDefaultConnection($this->config->get('tenancy.database.central_connection'));


CristianTudorache:
This pull fixes the following scenario:
If session driver used is database than when you use $tenant->run() in central/tenant context you will get an error "Call to a member function prepare() on null".
This is the case for multi db scenario.
This happens because initial db connection is purged and database session manager is not initialized again.

stancl:
If you think there's a bug with the session driver, please open an issue.

Seems like there's a reason why `purge()` is used - judging by the failing tests.

Thanks!

dligero-mdmsystems:
Hi!

I faced a similar issue integrating Jetstream.
The session manager was storing the session data on the "default" database connection instead in the tenant connection.
I was forced to create a "tenant" config on the database config file, but then, the error "Call to a member function prepare() on null" was appearing when the framework tried to store the session.

The code from this pr solves the problem.

zek:
I also faced with same issue.

Session driver always uses central connection. This is a serious bug if we're using database driver.

zek:
My workaround is recreating `SessionManager` with a custom bootstrapper.

```php
namespace App\Bootstrapper;


use Illuminate\Session\SessionManager;
use Stancl\Tenancy\Contracts\TenancyBootstrapper;
use Stancl\Tenancy\Contracts\Tenant;

class SessionBootstrapper implements TenancyBootstrapper
{

    public function bootstrap(Tenant $tenant)
    {
        if ($tenant) {
            app()->extend(SessionManager::class, function ($service, $app) {
                return new SessionManager($app);
            });
        }
    }

    public function revert()
    {

    }
}
```

bliveinhack:
We also having same problem.

When using database as session driver can not run tenant context queries.

@stancl  Problem still persist in 3.4.0 .  Am I missing something here ?

Any help would be appreciate.

stancl:
> If you think there's a bug with the session driver, please open an issue.
> 
> Seems like there's a reason why `purge()` is used - judging by the failing tests.
> 
> Thanks!

zek:
I was using
https://laravelcollective.com/docs/5.4/html
and registering custom elements in app service provider. It somehow
accesses session driver before tenancy package. So I've removed it and it
works.

I wonder if we can determine tenant in service loading phase without using
Middleware. This can solve all the problems.

5 Ara 2020 Cmt, saat 17:55 tarihinde Samuel Štancl <notifications@github.com>
şunu yazdı:

stancl:
> I wonder if we can determine tenant in service loading phase without using middleware

Only with domain/subdomain identification.

You can access `request()->getHost()` inside a SP but you can't access any of the Route-related information.

The bottom part of the "early identification" page of docs explains this a bit.

ettiennelouw:
I get the following error no matter what I do, reinstalled package, clear all cache, set db_connection=central/mysql/default none work. It worked before when I initially installed the package and not I was at the port to change the 3rd party package settings, livewire/permissions etc now it gives me this error and I just cannot get passe it.

SQLSTATE[42S02]: Base table or view not found: 1146 Table 'tenancy.sessions' doesn't exist (SQL: select * from `sessions` where `id` = 0kh6si2RaP0MoPbDQIvjnhKiiJJc6ZOfiZe0zorf limit 1)

PR #504, opened 2020-10-01
Fix tenants:list when not using multi-domain tenancy

petecoop:
This is to fix the below when not using the `HasDomains` trait.

```
tenants:list
Listing all tenants.

   Error

  Call to a member function pluck() on null

  at vendor/stancl/tenancy/src/Commands/TenantList.php:38
     34▕         tenancy()
     35▕             ->query()
     36▕             ->cursor()
     37▕             ->each(function (Tenant $tenant) {
  ➜  38▕                 $this->line("[Tenant] id: {$tenant['id']} @ " . implode('; ', $tenant->domains->pluck('domain')->toArray() ?? []));
     39▕             });
     40▕     }
     41▕ }
     42▕
```

### DIFF
--- a/src/Commands/TenantList.php
+++ b/src/Commands/TenantList.php
@@ -35,7 +35,10 @@ class TenantList extends Command
             ->query()
             ->cursor()
             ->each(function (Tenant $tenant) {
-                $this->line("[Tenant] id: {$tenant['id']} @ " . implode('; ', $tenant->domains->pluck('domain')->toArray() ?? []));
+                $this->line(
+                    "[Tenant] id: {$tenant['id']}" .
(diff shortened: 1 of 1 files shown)


stancl (changes requested):
Hey,

Thanks for contributing. Please see the suggested changes.

petecoop:
I don't think I saw any tests for this command, but I'm happy to add some

stancl:
GitHub is weird, it doesn't show the change I requested 🤔 

Anyway, please make it like:
```php
if ($tenant->domains) {
    // old code
} else {
    // new code without domains
}
```

Thanks!

PR #512, opened 2020-10-24
[3.x] Add method to run a callback on the central application

### LINKED ISSUE #487: Helper for global context
I suggest adding an helper to run some part of the code in global context. More or less an opposite to the `run()`-Method on tenant model.

**Example:**
```
global()->run(function() use($subscription) {
  $subscription('xx')->swap();
});
```

There are cases without access to the e.g. model so that I can't add the `CentralConnection` Trait to it. Specifically I have that problem with the _cashier-mollie_ package which does not yet provide a functionality to customize the models (https://github.com/laravel/cashier-mollie/issues/146).

I have to run the swap method in global context to work.

morloderex:
Closes: https://github.com/stancl/tenancy/issues/487

I found some free time to create a helper for when a tenant has been identified and you want to run a callback on the central application.

I'm not quite sure about the naming tho: 

Other naming possiblities considered: 
- runAsCentral($callback)
- onCentral($callback)
- runCentral($callback)

### DIFF
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -85,6 +85,27 @@ class Tenancy
         return array_map('app', $resolve($this->tenant));
     }
 
+    /**
+     * @param callable $callback
+     * @return mixed
+     */
+    public function runGlobal(callable $callback)
+    {
+        $oldTenant = $this->tenant;
+
+        if ($this->initialized) {
+            $this->end();
+        }
+
+        $result = $callback();
+
+        if ($oldTenant) {
+            $this->initialize($oldTenant);
+        }
+
+        return $result;
+    }
+
     public function query(): Builder
     {
         return $this->model()->query();
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -86,7 +86,7 @@ class Tenancy
     }
 
     /**
-     * Run a callback using the central environment.
+     * Run a callback inside the central environment.
      *
      * @param callable $callback
      * @return mixed
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -86,6 +86,8 @@ class Tenancy
     }
 
     /**
+     * Run a callback using the central environment.
+     *
      * @param callable $callback
      * @return mixed
      */
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -89,7 +89,7 @@ class Tenancy
      * @param callable $callback
      * @return mixed
      */
-    public function runGlobal(callable $callback)
+    public function central(callable $callback)
     {
         $oldTenant = $this->tenant;
 
--- a/tests/AutomaticModeTest.php
+++ b/tests/AutomaticModeTest.php
@@ -71,10 +71,58 @@ class AutomaticModeTest extends TestCase
 
         $this->assertSame('foobar', app('tenancy_initialized_for_tenant'));
     }
+
+    /** @test */
+    public function running_the_global_tenancy_helper_with_tenant_already_initialized()
+    {
+        MyBootstrapper::$revertedCallCount = 0;
+        GlobalRun::$count = 0;
+
+        config(['tenancy.bootstrappers' => [
+            MyBootstrapper::class,
+        ]]);
+
+        $tenant = Tenant::create([
+            'id' => 'acme',
+        ]);
+
+        tenancy()->initialize($tenant);
+
+        $this->assertSame('acme', app('tenancy_initialized_for_tenant'));
+
+        tenancy()->runGlobal(function () {
+            GlobalRun::$count = 1;
+        });
+
+        $this->assertSame(1, MyBootstrapper::$revertedCallCount);
+        $this->assertSame(1, GlobalRun::$count);
+        $this->assertSame('acme', app('tenancy_initialized_for_tenant'));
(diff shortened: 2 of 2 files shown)


tests/AutomaticModeTest.php
```diff
@@ -71,17 +71,71 @@ public function context_is_switched_when_tenancy_is_reinitialized()
...
+        $tenant = Tenant::create([
+            'id' => 'acme',
+        ]);
+
+        tenancy()->initialize($tenant);
+
+        $this->assertSame('acme', app('tenancy_initialized_for_tenant'));
+
+        tenancy()->runGlobal(function () {
+            GlobalRun::$count = 1;
```
stancl: I think you could just assert that the connection was `tenant` before this call and is not `tenant` inside it.

src/Tenancy.php
```diff
@@ -85,6 +85,27 @@ public function getBootstrappers(): array
         return array_map('app', $resolve($this->tenant));
     }
 
+    /**
```
stancl, line 88: Add a method description

src/Tenancy.php
```diff
@@ -85,6 +85,27 @@ public function getBootstrappers(): array
...
+     * @param callable $callback
+     * @return mixed
+     */
+    public function runGlobal(callable $callback)
```
stancl: I like to use `global` for things that are genuinely global — things that are the same in the central & tenant parts of the app.

So `central` is more appropriate here.

I think `tenancy()->central(callable $callback)` is the ideal solution.

stancl (changes requested):
Hey. Thanks for submitting this.

I suggested a few changes.

tests/AutomaticModeTest.php
morloderex: This is to make sure that the callback is indeed being called.

src/Tenancy.php
morloderex: As I mentioned in the description of this pr i went back and furth on this.. We could settle on global if you prefer.. 👍

tests/AutomaticModeTest.php
stancl: Count is not really an intuitive way of testing things. I'd make assertions that:

1. The connection **was** `tenant` but is **not** anymore, *testing that we're in the central context*
2. Setting a value in `app()`, *testing that the callback is being called*

morloderex:
Hey @stancl I believe your comments have been addressed now. 

Can you take a second look?

src/Tenancy.php
morloderex: method description added

src/Tenancy.php
morloderex: Method name has been renamed

tests/AutomaticModeTest.php
morloderex: Can you take a look at the changes to the tests and se if you like this better?

src/Tenancy.php
```diff
@@ -86,10 +86,12 @@ public function getBootstrappers(): array
     }
 
     /**
+     * Run a callback using the central environment.
```
stancl:
```suggestion
     * Run a callback inside the central environment.
```

tests/AutomaticModeTest.php
```diff
@@ -90,52 +87,53 @@ public function running_the_global_tenancy_helper_with_tenant_already_initialize
...
+        $this->assertFalse(tenancy()->initialized);
+        $this->assertFalse(app()->bound('tenancy_ended'));
+        $this->assertFalse(app()->bound('tenancy_initialized_for_tenant'));
+
+        tenancy()->central(function () use (&$runned) {
+            $this->assertFalse(app()->bound('tenancy_initialized_for_tenant'));
+            $this->assertFalse(app()->bound('tenancy_ended'));
+            $this->assertFalse(tenancy()->initialized);
+
+            $runned = 1;
```
stancl, line 116: What I meant is to:

1. Check that tenancy was initialized by seeing that `config('database.default')` **was** `tenant` and is not anymore.
2. Do `app()->instance('central_callback_executed', true)` inside this closure.

I can just rewrite the whole thing if you want.

It should
- set it to false at the beginning of the test
- check that it's false right before the central() call
- set it to true inside the central call
- check that it's true after the central call

stancl (changes requested):

tests/AutomaticModeTest.php
morloderex: @stancl We have 2 different tests here which tests different scenarios seperatly.  

One where the tenant is already initialized and one where a tenant is not already initialized.

The test you are commenting on is the use case where a tenant has not been initialized so i want to make sure that the TenancyManager does not call the `end` method before running the callback. 

However, am not quite sure why you don't see this logic as sufficient? I am not quite sure why we would need to go through the entire Laravel container class to check that a veriable is changed within a callback? Using a variable passed in by reference does the same thing here and the assertions called within the callback tests wether or not the bootstrapper provided has been executed. Can you ellabrorate on throughs here?

stancl:
there are some other changes that I'd want here, but this is a small amount of code so it's literally easier to just write it than to explain the changes

appreciate the effort, closing in favor of #526

PR #534, opened 2020-11-19
Add --only-selected option for tenants:migrate command

xdimension:
This PR is a follow-up of this issue https://github.com/stancl/tenancy-docs/issues/73#issue-689454245
In that issue, we were talking about how to skip the failing tenant migration because an exception was raised, and somehow our solution was to suppress any potential exceptions. After some consideration, I think the better solution is we provide the list of 'valid' tenants to `tenants:migrate` command (yes, as you said earlier) but we should be able to provide the list of tenants automatically instead of specifying one by one using `--tenants` option. Therefore, I added a new option `--only-selected` (suggest me if the words aren't okay); when the `tenants:migrate` runs with this option, developer can create a custom method that returns the tenant IDs that becomes a filter which tenants  are going to be migrated. For flexibility the class and method name can be configured in tenancy config `migration_filter_tenants_method`. I need this functionality to postpone the creation of the tenant database as it might be waiting for approval, and I think this method can be applied to some other cases too.

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -179,6 +179,11 @@ return [
     ],
 
     /**
+     * The class and method name used by tenants:migrate command with --only-selected option.
+     */
+    'migration_filter_tenants_method' => [\App\Repositories\TenantRepository::class, 'filterBy'],
+
+    /**
      * Parameters used by the tenants:seed command.
      */
     'seeder_parameters' => [
--- a/src/Commands/Migrate.php
+++ b/src/Commands/Migrate.php
@@ -17,6 +17,20 @@ class Migrate extends MigrateCommand
     use HasATenantsOption, DealsWithMigrations;
 
     /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'migrate {--database= : The database connection to use}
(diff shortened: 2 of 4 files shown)


assets/config.php
```diff
@@ -178,6 +178,11 @@
         '--realpath' => true,
     ],
 
+    /**
+     * The class and method name used by tenants:migrate command with --only-selected option.
+     */
+    'migration_filter_tenants_method' => [\App\Repositories\TenantRepository::class, 'filterBy'],
```
stancl, line 184: This puts a burden on the developer by having to create this class.

stancl (commented):
Hey.

Thanks for the contribution, but I don't think this is the right way to go about it.

This adds significant complexity at little gain.

> but we should be able to provide the list of tenants automatically instead of specifying one by one using --tenants option

Why not create your own command that uses your own logic and calls `tenants:migrate` for the right tenants?

This seems like something that belongs into your application rather than the package.

assets/config.php
xdimension: It's optional actually.

xdimension:
Ok, I can make my own command instead of package's. Thanks for your review.

PR #536, opened 2020-11-23
ensure tenant will be rollbacked when calling central helper

bepsvpt:
When using `central` helper, if there is something wrong in user's callback, the tenancy will not rollback to origin tenant. 

This pull request ensures it will be rollbacked to origin tenant even when the callback throws exceptions.

### DIFF
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Stancl\Tenancy;
 
+use Exception;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Traits\Macroable;
@@ -109,6 +110,8 @@ class Tenancy
      *
      * @param callable $callback
      * @return mixed
+     *
+     * @throw Exception
      */
     public function central(callable $callback)
     {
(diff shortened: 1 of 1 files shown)


stancl:
Hey.

Can you give an example use case for that?

Thanks!

bepsvpt:
For example, assume the current scope is in tenant 1 and the app needs to do something in central scope.

If there is some unexpected errors in central callback, the scoping will be broken.

```php
// tenant 1 scope

try {
  tenancy()->central(function () {
    // business logic (central scope)
  });
} catch (Exception $e) {
  // log error (here should be tenant 1 scope)
}

// here should also be tenant 1 scope
```

stancl:
I mean, what would be the example of that logic.

It seems weird to do this implicitly, especially when `run()` doesn't do it.

If you're running some operation that might fail, why not do the exception handling inside the callback?

bepsvpt:
> If you're running some operation that might fail, why not do the exception handling inside the callback?

Our application handles exceptions in central handler. If the tenant scope isn't rollbacked correctly, there will cause another issue when processing the exceptions.

stancl:
> If the tenant scope isn't rollbacked correctly

Why would this happen? And if it were an issue, the right place to handle it wouldn't be the central helper in my opinion. There are many more things that change the app context (`run()`, `central()`, `initialize()`, ...)

PR #540, opened 2020-11-28
Improve PHPUnit assertions

peter279k:
# Changed log

- Using the `assertTrue` to assert expected is `true`.
- Using the `assertFalse` to assert expected is `false`.
- Using the `assertNull` to assert expected is `null`.


stancl:
https://github.com/stancl/virtualcolumn/pull/5#event-4048382634

PR #544, opened 2020-12-01
Feature: New PreventAccessFromTenants Middleware

robertdrakedennis:
As discussed in #help you were looking for an additional solution to the ``domain`` functionality in routes, with this middleware which is the inverse of ``PreventAccessFromCentralDomains`` should suffice. Let me know your thoughts.


stancl:
(I don't have time to deal with this right now, but for future reference: The reason why this is a draft is that using a different way than `domain()` to register central routes leads to the issue where you can't use `/` for both central and tenant routes)

PR #550, opened 2020-12-07
[FIX] Storage base path.

sirrom-ekim:
Was previously labeled as a suffix when in fact it was a prefix

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -96,9 +96,9 @@ return [
      */
     'filesystem' => [
         /**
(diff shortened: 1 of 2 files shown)


stancl:
The file bootstrapper is a bit complex.

I hope this graphic clarifies the logic a bit:

Screen Shot 2020-12-07 at 08 57 36

It's called a *suffix* because it's *suffixed* to the original *root* (= prefix).

It's not a prefix, because if you had a disk with the `/foo` root, and you configured this configuration key to be `/bar`, it would set the disk's root to `/foo`, even though it should *suffix* it with `/bar`, resulting in `/foo/bar`.

PR #598, opened 2021-02-06
Check DB existsance  as config

oniz93:
I have a project with an high I/O (20/30 API requests per second) and using all the caches remains a query to the DB: the one  that check DB existence.
I would like to make optional that query since I'm the admin of the DB and I know when a DB exists or not.

I hope that it's a good feature :)

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -39,6 +39,7 @@ return [
      * Database tenancy config. Used by DatabaseTenancyBootstrapper.
      */
     'database' => [
+        'check_db_exists' => true,
         'central_connection' => env('DB_CONNECTION', 'central'),
 
         /**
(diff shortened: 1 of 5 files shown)


oniz93:
I don't know why codeconv says that the coverage is less. I just added a bunch of "if"...

stancl:
Where is the method called? DB existence should only be checked on tenant *creation* in production, not on requests.

https://github.com/stancl/tenancy/blob/411d486282517956ae32976cabe2cda07ad887ab/src/Bootstrappers/DatabaseTenancyBootstrapper.php#L27-L32

oniz93:
In my case I use PostgreSQLSchemaManager and it's checked at each bootstrap of the application, not only on tenant creation.
I modified all the DB Managers because I thought that the behaviour it's always the same.
Let me know if this is the behaviour only for PostgreSQLSchemaManager and I'll change only there

stancl:
See the code I sent. Do you have the correct environment configured in production?

oniz93:
You are right. The problem was my APP_ENV.
Thank you so much!
I close the PR

PR #611, opened 2021-03-09
Fix arguments being initialized twice for tenants:seed command

morloderex:
Laravel >= 8.32 introduced a new argument to the seed command see https://github.com/laravel/framework/pull/36513.

This is causing our extended tenants:seed command to initialize the options and arguments twice which throws an exception because at the point we call the `specifyParameters` method in the constructor it has already instanciated the arguments once so the new `class` argument is already present.

This pr fixes that by removing our call to the `specifyParameters` method on the constructor sense it's already being called in the parent constructor anyway and the only thing we are really doing is specifying one additional option to the command anyway.

### DIFF
--- a/src/Commands/Seed.php
+++ b/src/Commands/Seed.php
@@ -21,6 +21,8 @@ class Seed extends SeedCommand
      */
     protected $description = 'Seed tenant database(s).';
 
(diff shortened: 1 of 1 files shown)


stancl:
Thanks!

PR #619, opened 2021-03-22
Update 2019_09_15_000010_create_tenants_table.php

AbrahamBrookes:
just added a comment to clarify how to use these custom columns on the tenant model


stancl:
Hey. I don't think this is needed. Feels a bit wrong to refer to models from migrations, and you can technically use the columns even without changes to your model. And using custom columns is well documented.

PR #621, opened 2021-03-22
Adds a new tenant resolver using request origin

jonerickson:
Introduces a new tenant resolver using the request origin. This tenant resolver is lightweight and the perfect addition for an SPA that uses AJAX requests to a Laravel backend. By using the request origin for resolving, requests can be made without additional config on the frontend to include paths or extra request data.

### DIFF
--- a/src/Middleware/InitializeTenancyByRequestOrigin.php
+++ b/src/Middleware/InitializeTenancyByRequestOrigin.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use Stancl\Tenancy\Tenancy;
+use Stancl\Tenancy\Resolvers\RequestOriginTenantResolver;
+
+class InitializeTenancyByRequestOrigin extends IdentificationMiddleware
+{
+	/** @var callable|null */
+    public static $onFail;
+
+    /** @var Tenancy */
+    protected $tenancy;
+
+    /** @var TenantResolver */
+    protected $resolver;
+
+    public function __construct(Tenancy $tenancy, RequestOriginTenantResolver $resolver)
+    {
+        $this->tenancy = $tenancy;
+        $this->resolver = $resolver;
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+	    return $this->initializeTenancy($request, $next, $this->getPayload($request));
+    }
+
+	protected function getPayload(Request $request): ?string
+	{
+		$tenant = null;
+		if ($request->hasHeader('origin')) {
+			$tenant = optional(parse_url($request->headers->get('origin')))['host'];
+		}
+
+		return $tenant;
+	}
+}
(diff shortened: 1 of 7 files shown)


src/Middleware/InitializeTenancyByRequestOrigin.php
```diff
@@ -0,0 +1,50 @@
...
+    public function handle($request, Closure $next)
+    {
+	    return $this->initializeTenancy($request, $next, $this->getPayload($request));
+    }
+
+	protected function getPayload(Request $request): ?string
+	{
+		$tenant = null;
+		if ($request->hasHeader('origin')) {
+			$tenant = optional(parse_url($request->headers->get('origin')))['host'];
```
stancl: Does the resolver check the headers basically?

If so, what's the difference between this class and the request data resolver? https://github.com/stancl/tenancy/blob/3.x/src/Middleware/InitializeTenancyByRequestData.php#L55

Thanks!

jonerickson: Yes it does accept the request data resolver is expecting the ID of the tenant in the X-Tenant header (or whatever header you set) in order to properly resolve the tenant. This header must be explicitly set and sent from the client application. 

This new resolver uses the URL the request is originating from to resolve the tenant. The Origin header is automatically sent with any Ajax request in any modern browser thus preventing the need to send any additional data to resolve the tenant. The domain is then matched with the tenant to properly resolve. Browsers are in control of the Origin header thus making the resolver inherently safer than using a client generated header that any attacker could spoof (the request data resolver).

Thanks for the question.

morloderex: I think @stancl means what is the difference between using the request data helper with it's lookup header customized to being Origin with this new resolver?

stancl: I see. The request data middleware expects an ID, whereas the origin is the domain I think?

Still, you probably don't need another resolver because it's using the domain to resolve the tenant. The only difference is in the middleware. You could probably extend the domain MW and change the line that calls `request()->getHost()` to `request()->header('Origin')`.

jonerickson: @stancl Yes, the domain is needed to resolve rather than the ID.

I would suggest adding a separate MW that specifically is checking the request origin rather than the host as it currently stands and not requiring people to extend the the domain MW. The reason is, the Host header is where the request is being sent. The Origin header is where the request is coming from. 

An example use case:

A SPA running on tenant1.application.com is sending a request to a Laravel backend on api.application.com. In the current setup, request()->getHost() will return api.application.com and tenancy will fail to resolve because of the cross origin request and no tenant has a domain that matched api.application.com. However if we have tenancy resolving using the Origin header, request()->header('Origin') will return tenant1.application.com and tenancy will be properly resolved. The origin header is always sent on cross origin requests.

I would say the above scenario is the most common use case for enterprise level SPAs. I'd be happy to update the PR and remove the additional resolver, but add the request origin MW that uses the current domain resolver if you're interested, and if not, you can absolutely close this. Thanks.

stancl:
I'll take a second look at this if more people request the feature. Closing for now as there doesn't seem to be interest in it. Thanks for the PR either way

darkons:
I did my custom implementation of this. I'm using a SPA frontend too and setting X-Tenant header dynamically is headache.
I just create domains/subdomains for my tenants (like tenant1.app.com) that are used to make tenant identification by Origin header. This method is more transparent and easy that setting a header manually.

Maybe we can re-use current InitializeTenancyByDomain class just adding a new property like **useOriginHeader**:

```
use Stancl\Tenancy\Middleware\InitializeTenancyByDomain;

InitializeTenancyByDomain::$useOriginHeader = true;
```

PR #633, opened 2021-04-12
Fix issue 632: cached lookup

### LINKED ISSUE #632: $currentDomain is null when using cached lookup
#### Describe the bug
When using cached tenant lookup with subdomain identification, the `DomainTenantResolver::$currentDomain` always is `null` for cached lookups.

#### Steps to reproduce
1) Install `stancl/tenancy`
2) Create one tenant and its subdomain
3) Use subdomain identification
4) Enable cached lookup (set `DomainTenantResolver::$shouldCache = true;`)
5) Create Route, which displays `DomainTenantResolver::$currentDomain`
In ***`routes/tenant.php`***
```php
Route::get('/test', function () {
    dd(\Stancl\Tenancy\Resolvers\DomainTenantResolver::$currentDomain);
});
```
6) Access the route at least **twice**.
6.1. it displays Domain model (because it caches tenant on first call)
6.2. from now on it displays `null`

#### Expected behavior
Every time accessing this route (getting `DomainTenantResolver::$currentDomain`), you should get current domain instance.

##### Possible solution?
In ***`CachedTenantResolver.php`***
```php
  // define domain property
  public static $currentDomain;

  // update resolve method
  public function resolve(...$args): Tenant
  {
      // rest of the code...

      //  (line 39) the problem - currently the current domain is not set when using cached lookup
      if ($this->cache->has($key)) {
          $tenant = $this->cache->get($key);
          // I guess, it could be done like that, because it caches only one domain per key
          static::$currentDomain = $tenant->domains->first();
          return $tenant;
      }

      // rest of the code...
  }
```

#### Your setup
 - Laravel version: 8.36.0
 - stancl/tenancy version: 3.4.2
 - Cache driver: redis

ralfinsh:
Fixes #632 

#### What I changed
https://github.com/stancl/tenancy/blob/27e9fb4a692a15ca2fd320799d97fc657cb14b68/src/Resolvers/DomainTenantResolver.php#L31-L38

Instead of querying Domain model, find Tenant and eager load its domains via Tenant model. This way, you get `$tenant`, which already has loaded `domains` relationship ~~, **with only the current domain** (Tenant can have many domains, but `$tenant->domains` will have **only the identified domain!**)~~. This reduces SQL queries and memory usage - when using cache, Domain models are loaded from cache, not from the central database. Now there is no need to access central database, if you need to get current Domain or Tenant models, or any of Tenant's domains.

~~Added a new method to `CachedTenantResolver` - `tenantIdentifiedFromCache(Tenant $tenant, ...$args)`, which is called everytime, when Tenant is identified from cache. Then it's used in resolver, to set the current domain.~~

Added a new method to `CachedTenantResolver` - `tenantIdentified(Tenant $tenant, ...$args)`, which is used to set `$currentDomain` in `DomainTenantResolver`.

### DIFF
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -43,7 +43,7 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
             ->first();
 
         /** @var Domain|null $domain */
-        $domain = $tenant?->domains->first();
+        $domain = $tenant ? $tenant->domains->first() : null;
 
         if ($domain) {
             static::$currentDomain = $domain;
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -28,16 +28,35 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
 
     public function resolveWithoutCache(...$args): Tenant
     {
-        /** @var Domain $domain */
-        $domain = config('tenancy.domain_model')::where('domain', $args[0])->first();
+        $domain = $args[0];
+
+        /** @var Tenant|null $tenant */
+        $tenant = config('tenancy.tenant_model')::query()
+            ->whereHas('domains', function ($query) use ($domain) {
+                $query->select(['tenant_id', 'domain'])->where('domain', $domain);
+            })
+            ->with([
+                'domains' => function ($query) use ($domain) {
+                    $query->where('domain', $domain);
+                },
+            ])
+            ->first();
+
+        /** @var Domain|null $domain */
+        $domain = $tenant?->domains->first();
 
         if ($domain) {
             static::$currentDomain = $domain;
 
-            return $domain->tenant;
+            return $tenant;
         }
 
         throw new TenantCouldNotBeIdentifiedOnDomainException($args[0]);
+    }
+
+    public function tenantIdentifiedFromCache(Tenant $tenant, ...$args): void
+    {
+        static::$currentDomain = $tenant->domains->first();
     }
 
     public function getArgsForTenant(Tenant $tenant): array
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -35,15 +35,11 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
             ->whereHas('domains', function ($query) use ($domain) {
                 $query->select(['tenant_id', 'domain'])->where('domain', $domain);
             })
-            ->with([
-                'domains' => function ($query) use ($domain) {
-                    $query->where('domain', $domain);
-                },
-            ])
+            ->with('domains')
             ->first();
 
         if ($tenant) {
-            static::$currentDomain = $tenant->domains->first();
+            $this->tenantIdentified($tenant, ...$args);
 
             return $tenant;
         }
@@ -51,9 +47,9 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
         throw new TenantCouldNotBeIdentifiedOnDomainException($args[0]);
     }
 
-    public function tenantIdentifiedFromCache(Tenant $tenant, ...$args): void
+    public function tenantIdentified(Tenant $tenant, ...$args): void
     {
-        static::$currentDomain = $tenant->domains->first();
+        static::$currentDomain = $tenant->domains->where('domain', $args[0])->first();
     }
 
     public function getArgsForTenant(Tenant $tenant): array
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -33,13 +33,13 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
         /** @var Tenant|null $tenant */
         $tenant = config('tenancy.tenant_model')::query()
             ->whereHas('domains', function ($query) use ($domain) {
-                $query->select(['tenant_id', 'domain'])->where('domain', $domain);
+                $query->where('domain', $domain);
             })
             ->with('domains')
             ->first();
 
         if ($tenant) {
-            $this->tenantIdentified($tenant, ...$args);
+            $this->setCurrentDomain($tenant, ...$args);
 
             return $tenant;
         }
@@ -48,6 +48,11 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
     }
 
     public function tenantIdentified(Tenant $tenant, ...$args): void
+    {
+        $this->setCurrentDomain($tenant, ...$args);
+    }
+
+    protected function setCurrentDomain(Tenant $tenant, ...$args): void
     {
         static::$currentDomain = $tenant->domains->where('domain', $args[0])->first();
     }
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -7,6 +7,7 @@ namespace Stancl\Tenancy\Resolvers;
 use Stancl\Tenancy\Contracts\Domain;
 use Stancl\Tenancy\Contracts\Tenant;
 use Stancl\Tenancy\Exceptions\TenantCouldNotBeIdentifiedOnDomainException;
+use Illuminate\Database\Eloquent\Builder;
 
 class DomainTenantResolver extends Contracts\CachedTenantResolver
 {
@@ -32,14 +33,14 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
 
         /** @var Tenant|null $tenant */
         $tenant = config('tenancy.tenant_model')::query()
-            ->whereHas('domains', function ($query) use ($domain) {
+            ->whereHas('domains', function (Builder $query) use ($domain) {
                 $query->where('domain', $domain);
             })
             ->with('domains')
             ->first();
 
         if ($tenant) {
-            $this->setCurrentDomain($tenant, ...$args);
+            $this->setCurrentDomain($tenant, $domain);
 
             return $tenant;
         }
@@ -47,14 +48,14 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
         throw new TenantCouldNotBeIdentifiedOnDomainException($args[0]);
     }
 
-    public function tenantIdentified(Tenant $tenant, ...$args): void
+    public function resolved(Tenant $tenant, ...$args): void
     {
-        $this->setCurrentDomain($tenant, ...$args);
+        $this->setCurrentDomain($tenant, $args[0]);
     }
 
-    protected function setCurrentDomain(Tenant $tenant, ...$args): void
+    protected function setCurrentDomain(Tenant $tenant, string $domain): void
     {
-        static::$currentDomain = $tenant->domains->where('domain', $args[0])->first();
+        static::$currentDomain = $tenant->domains->where('domain', $domain)->first();
     }
 
     public function getArgsForTenant(Tenant $tenant): array
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -42,11 +42,8 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
             ])
             ->first();
 
-        /** @var Domain|null $domain */
-        $domain = $tenant ? $tenant->domains->first() : null;
-
-        if ($domain) {
-            static::$currentDomain = $domain;
+        if ($tenant) {
+            static::$currentDomain = $tenant->domains->first();
 
             return $tenant;
         }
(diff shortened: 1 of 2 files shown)


ralfinsh:
I can't reproduce these failures locally with running tests on docker, with PHP 8.0 and trying both Laravel ^6.0 and ^8.0.
Feels like, there is something weird happening on the CI tests. 
Could You, @stancl, please, chek it?

src/Resolvers/DomainTenantResolver.php
```diff
@@ -28,18 +28,34 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
...
+
+            return $tenant;
         }
 
         throw new TenantCouldNotBeIdentifiedOnDomainException($args[0]);
     }
 
+    public function tenantIdentifiedFromCache(Tenant $tenant, ...$args): void
+    {
+        static::$currentDomain = $tenant->domains->first();
```
stancl: As I mentioned, you cannot use the first domain here. **Tenants can have multiple domains**. See what I suggested in the original thread.

src/Resolvers/DomainTenantResolver.php
```diff
@@ -28,18 +28,34 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
...
+        if ($tenant) {
+            static::$currentDomain = $tenant->domains->first();
+
+            return $tenant;
         }
 
         throw new TenantCouldNotBeIdentifiedOnDomainException($args[0]);
     }
 
+    public function tenantIdentifiedFromCache(Tenant $tenant, ...$args): void
```
stancl: This should be called from `resolveWithoutCache()` and as such renamed to something unrelated to cache. Right now this logic is duplicated between `resolveWithoutCache()` and this method.

src/Resolvers/DomainTenantResolver.php
```diff
@@ -28,18 +28,34 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
...
-        /** @var Domain $domain */
-        $domain = config('tenancy.domain_model')::where('domain', $args[0])->first();
+        $domain = $args[0];
 
-        if ($domain) {
-            static::$currentDomain = $domain;
+        /** @var Tenant|null $tenant */
+        $tenant = config('tenancy.tenant_model')::query()
+            ->whereHas('domains', function ($query) use ($domain) {
+                $query->select(['tenant_id', 'domain'])->where('domain', $domain);
```
stancl: What's the select for?

src/Resolvers/DomainTenantResolver.php
ralfinsh: No, in this context you can, because the $tenant comes from DomainTenantResolver, and you can have multiple domains, but the **$tenant will have only the current domain loaded on the relationship**. Look at DomainTenantResolver line 35, where $tenant is fetched from database.

src/Resolvers/DomainTenantResolver.php
ralfinsh: It's my habit to select only used columns.

To explain, why **you can** use `$tenant->domains->first()`:
The queries executed on line 34 looks like this:
```sql
select * from `tenants` where exists (select `tenant_id`, `domain` from `domains` where `tenants`.`id` = `domains`.`tenant_id` and `domain` = DOMAIN) limit 1
select * from `domains` where `domains`.`tenant_id` in (TENANT_ID) and `domain` = DOMAIN
```
`$tenant` **always** will have **only one** related domain (as we can see in the raw sql query), and it always will be the current domain (because we are finding tenants id by domain, then eager loading the current domain). I also checked that manually, by creating multiple tenants with multiple domains, and with cache enabled - it always returns correct domain model **without calling database**.

src/Resolvers/DomainTenantResolver.php
ralfinsh: [Explanation](https://github.com/stancl/tenancy/pull/633#discussion_r611895731).

src/Resolvers/DomainTenantResolver.php
ralfinsh: I just thought, that a method like `tenantIdentifiedFromCache()` makes this package more flexible, when you are writing your own resolvers and need to implement some custom logic, as name implies, when tenant is identified from cache. It's like an event (firing an event and listening for it would be an overkill and only would make more problems, for a simple task like this). Ofcourse, if it is needed, I can update code like this:
```php

public function resolveWithoutCache(...$args)
{
    // rest of the code
    if ($tenant) {
        $this->setCurrentDomain($tenant->domains->first());
    
        return $tenant;
    }

    // rest of the code
}

public function tenantIdentifiedFromCache(Tenant $tenant, ...$args)
{
    $this->setCurrentDomain($tenant);
}

protected function setCurrentDomain(Tenant $tenant)
{
    self::$currentDomain = $tenant->domains->first();
}
```

ralfinsh: But on the other hand, it could be simple `tenantIdentified()` method, which is called whenever Tenant is identified (from `resolve()` in *CachedTenantIdentifier* and `resolveWithoutCache()` in *DomainTenantResolver*), hmmm. Please, make final decision about this, and I will make the necessary changes.

```php
public function resolveWithoutCache(...$args)
{
    // rest of the code
    if ($tenant) {
        $this->tenantIdentified($tenant->domains->first());
    
        return $tenant;
    }

    // rest of the code
}

public function tenantIdentified(Tenant $tenant)
{
    self::$currentDomain = $tenant->domains->first();
}
```

src/Resolvers/DomainTenantResolver.php
ralfinsh: Can I resolve this conversation or you don't like this approach? @stancl

src/Resolvers/DomainTenantResolver.php
stancl: Okay, so the tenant instance will have incorrect data loaded in the $domains property? What if someone wants to use tenant()->domains to fetch all domains?

ralfinsh: You can fetch them from database with `tenant()->domains()->get()`, but I get your point on this.
So, you want me to refactor the code, so `tenant()->domains` still contains all domains, and when using cache, there are no calls to database, right? (I guess, when setting current domain, find it in `$tenant->domains->where('domain', $domain)`)
I am just trying to get rid of unnecessary calls to central database, because, the suggested method queries the domain from DB, but it is not necessary, as the Domain model already exists in cache.

ralfinsh:
Now, with this pull request:
- Issue #632 is fixed
- When accessing `tenant()` or `tenant()->domains` (also, it shows **all** domains) with cached lookup enabled, there are **no calls made to central db** (so this fix does not affect existing functionality)

src/Resolvers/DomainTenantResolver.php
```diff
@@ -28,18 +28,30 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
...
+            })
+            ->with('domains')
```
stancl, line 39: Is the `with()` needed? Afaik it makes a separate query.

src/Resolvers/DomainTenantResolver.php
```diff
@@ -28,18 +28,30 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
...
-        /** @var Domain $domain */
-        $domain = config('tenancy.domain_model')::where('domain', $args[0])->first();
+        $domain = $args[0];
 
-        if ($domain) {
-            static::$currentDomain = $domain;
+        /** @var Tenant|null $tenant */
+        $tenant = config('tenancy.tenant_model')::query()
+            ->whereHas('domains', function ($query) use ($domain) {
+                $query->select(['tenant_id', 'domain'])->where('domain', $domain);
```
stancl: To be fair I'd remove the `select()` because any performance gains will be negligible, and it makes the code harder to read. It seems like a microoptimization that will ultimately just lead to issues down the line. For example, someone might use a different column than `tenant_id` — for example, `team_id`. The point of `whereHas()` is that it abstracts these things, and we can safely make as few assumptions as needed, namely that there's a `domain` column, and that there's a defined relationship.

ralfinsh: Okay, will remove that.

src/Resolvers/DomainTenantResolver.php
ralfinsh: Yes, it makes separate query, to resolve the issue #632 with cached lookup. It eager loads the Domains, and when cache is enabled, there are no calls to Central DB when accessing DomainTenantResolver::$currentDomain or tenant()->domains. If I drop that line, then I would need to modify `tenantIdentified()` method, and from there it would always call central database to get Domain model, which, in my opinion, makes caching pointless.

ralfinsh: Also, with the original approach, if tenant had multiple domains, when accessing `tenant()->domains` it fetches all domains from database, including the current domain, which already is in memory, resulting in one model loading in memory and from DB twice.

src/Resolvers/DomainTenantResolver.php
```diff
@@ -28,18 +28,30 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
...
+            })
+            ->with('domains')
+            ->first();
 
-            return $domain->tenant;
+        if ($tenant) {
+            $this->tenantIdentified($tenant, ...$args);
```
stancl: I think that rather than calling `tenantIdentified()` from this method, the logic that sets the current domain should be in a separate method, one that's called both from here and from `tenantIdentified()`.

Seems weird to never call "tenant identified" when cache is not used in the other resolvers, but use it in this one. The methods should have more specific roles.

src/Resolvers/DomainTenantResolver.php
stancl: Hmm fair. I guess there are two approaches.

1. $domain = Domain where id $arg[0], $tenant = $domain->tenant, use $domain for $currentDomain
2. Tenant = Tenant where has domain for $arg[0], with the domains relation loaded.

I think both make the same amount of queries, so yeah, the `with()` is probably fine. If Laravel caches the relation too, it will have the benefit of `tenant()->domains` not making a central query.

stancl:
I made some changes to the code, can you please check what test started failing?

ralfinsh:
> 
> 
> I made some changes to the code, can you please check what test started failing?

@stancl Github Actions sometimes fail for no reason. Thanks for the cleanup! I pushed a commit (to get the tests run again) and now everything passed.

stancl:
Thanks!

shimsag:
I'm getting below error after upgrade from 3.4.2 to 3.4.3
Can you please assist ?

`SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer LINE 1: select * from "domains" where "domains"."tenant_id" in (0) ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. (SQL: select * from "domains" where "domains"."tenant_id" in (0))`

aman-rohiman:
> 2 to

Same to me, I getting error like that after upgrade from 3.4.2 to 3.4.3

aman-rohiman:
> I'm getting below error after upgrade from 3.4.2 to 3.4.3
> Can you please assist ?
> 
> `SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer LINE 1: select * from "domains" where "domains"."tenant_id" in (0) ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. (SQL: select * from "domains" where "domains"."tenant_id" in (0))`

Same to me, I getting error like that after upgrade from 3.4.2 to 3.4.3

ralfinsh:
@shimsag could you please open an issue with details and steps to reproduce. I think the problem is in your code. Can't assist without details.

aman-rohiman:
> @shimsag could you please open an issue with details and steps to reproduce. I think the problem is in your code. Can't assist without details.

In my case, I'm using Postgres 11, and everything is normal, but after i run composer update, when I try to open tenant domain, I'm getting those error.
Step to reproduce:
1. Use postgres database
2. Open tenant before upgrade, everything went normal
3. Composer update
4. Open tenant after upgrade
5. Error "SQLSTATE[42883]" shown up
To make sure the source of problem, I replace vendor stancl/tenancy with previous version (3.4.2) everything went normal, but after replace it again with stancl/tenancy 3.4.3, the error show up again.
I hope it helps.

stancl:
What's your data type for `tenant_id` on `domains` table and the id on the tenants table?

stancl:
The only change is that the domain is not getting fetched via `Domain::find()` but via the relation (`$tenant->domains()->where()`) so the issue must be a mismatching type for those two columns.

aman-rohiman:
> What's your data type for `tenant_id` on `domains` table and the id on the tenants table?

Data type for tenant_id column on domains table is varchar(255), for example "cbab4d88-5b45-4c91-8cf7-37accf642960", this is same data type with id column on tenants table, i'm using saas boilerplate and using postgresql database, and this problem occured after composer update.

stancl:
Can you try running `composer update` and checking if this is fixed now?

aman-rohiman:
> Can you try running `composer update` and checking if this is fixed now?

Nice, it works now, thank you @stancl for the fast response.

stancl:
Great, thanks!

PR #636, opened 2021-04-13
Adds ability to disable tenant asset routes

hackeresq:
In some cases, it may not be necessary to have a tenant asset route exposed. In such cases, it would be nice to disable the default `/tenancy/asset/{file}` route that is generated. 

This is not a breaking change as the config defaults to "true" -- as in the route is registered by default. 

```php
config('tenancy.filesystem.serve_tenant_assets', true)
```

I do not believe this needs additional testing. If it does, happy to write a test.

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -135,6 +135,13 @@ return [
          * where you want to use tenant-specific assets (product images, avatars, etc).
          */
         'asset_helper_tenancy' => true,
+        
+        /**
+         * Should the tenant asset route be registered. This route enables you to access files
+         * in the tenant-specific storage. By default, this route is enabled. But it may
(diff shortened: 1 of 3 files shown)


stancl:
I think that it might be better to add a conditional here, rather than creating a config property specific to the route.

https://github.com/stancl/tenancy/blob/27e9fb4a692a15ca2fd320799d97fc657cb14b68/src/TenancyServiceProvider.php#L113

Something like `tenancy.routes` being the config key. True by default, and if it's false, don't register any routes.

hackeresq:
Makes sense. Updated.

stancl:
Thanks!

PR #639, opened 2021-04-20
Set keyType to String for Tenant

Jaspur:
Got issues since last composer update with PostgreSQL and the Tenant model:

` LINE 1: select * from "domains" where "domains"."tenant_id" in (0) ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. (SQL: select * from "domains" where "domains"."tenant_id" in (0)) {"exception":"[object] (Illuminate\\Database\\QueryException(code: 42883): SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer LINE 1: select * from \"domains\" where \"domains\".\"tenant_id\" in (0) ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. (SQL: select * from \"domains\" where \"domains\".\"tenant_id\" in (0)) at `

Solved by setting the `keyType` to `string`.

### DIFF
--- a/src/Database/Models/Tenant.php
+++ b/src/Database/Models/Tenant.php
@@ -30,6 +30,7 @@ class Tenant extends Model implements Contracts\Tenant
 
     protected $table = 'tenants';
     protected $primaryKey = 'id';
+    protected $keyType = 'string';
     protected $guarded = [];
 
     public function getTenantKeyName(): string
--- a/tests/TenantModelTest.php
+++ b/tests/TenantModelTest.php
@@ -1,6 +1,6 @@
 <?php
 
-declare(strict_types=1);
+declare (strict_types = 1);
 
 namespace Stancl\Tenancy\Tests;
 
@@ -76,8 +76,8 @@ class TenantModelTest extends TestCase
         $tenant1 = Tenant::create();
         $tenant2 = Tenant::create();
 
-        $this->assertSame(1, $tenant1->id);
-        $this->assertSame(2, $tenant2->id);
+        $this->assertSame((string) 1, $tenant1->id);
+        $this->assertSame((string) 2, $tenant2->id);
     }
 
     /** @test */
--- a/tests/TenantModelTest.php
+++ b/tests/TenantModelTest.php
@@ -1,6 +1,6 @@
 <?php
 
-declare (strict_types = 1);
+declare(strict_types=1);
 
 namespace Stancl\Tenancy\Tests;
 


stancl:
Seems like the problem is somewhere else. The key type should definitely not be a string by default because it can just be a numeric id. Maybe you need to do extra config with postgres when using uuids?

Still, weird that the package's tests would be passing even with postgres 🤔

Jaspur:
@stancl but you are setting it as a string yourselves?
`$table->string('id')->primary()`
See the error I get on postgresql. This error didn't happened before. Different project which worked fine, also got this same issue after a composer update. 

But setting the keyType to string doesn't say that you cannot use ID's. It's just handled as a string in the query parameter. Seem correct while the column is a string type too. 

Strange that no one else opened an issue about this. In total I have four projects now with the issue after composing update. Some different versions of Postgres database driver too. 

See also this post: 
https://kim-jangwook.medium.com/use-uuid-as-primary-key-of-laravel-eloquent-orm-82e3db36cb62
There is also reference to use keyType string when using uuid's

ralfinsh:
I think that there is a need to implement `getKeyType()` method on `GeneratesIds` trait. If Tenant model generates ID's, I believe it always will be some kind of string, if not - use default key type. 

It could look like this (in **`Stancl\Tenancy\Database\Concerns\GeneratesIds`**):
```php
public function getKeyType()
{
    return $this->shouldGenerateId() ? 'string' : $this->keyType;
}
```

ralfinsh:
From Laravel [docs](https://laravel.com/docs/8.x/eloquent#primary-keys):
>  If your model's primary key is not an integer, you should define a protected `$keyType` property on your model.

So the issue right now is, that Eloquent assumes that Tenant model primary key type is `integer`. Basic queris are not affected by this assumption, but when working with relationships, it fails, because the primary key type in reality isn't `integer`, and when querying relationships this makes a difference.

I think, as I mentioned in last [comment](https://github.com/stancl/tenancy/pull/639#issuecomment-824603847), it can be solved with these lines, and it would work in both cases - when using autoincrement (`integer`) or generating IDs (`string`):

> ```
> public function getKeyType()
> {
>     return $this->shouldGenerateId() ? 'string' : $this->keyType;
> }
> ```

stancl:
> I think that there is a need to implement `getKeyType()` method on `GeneratesIds` trait. If Tenant model generates ID's, I believe it always will be some kind of string, if not - use default key type.
> 
> It could look like this (in **`Stancl\Tenancy\Database\Concerns\GeneratesIds`**):
> 
> ```
> public function getKeyType()
> {
>     return $this->shouldGenerateId() ? 'string' : $this->keyType;
> }
> ```

This seems like a good solution yeah, I'll do that

stancl:
Can you try running `composer update` and checking if this is fixed now?

Jaspur:
@stancl  Fixed!

PR #644, opened 2021-04-25
Support for soft delete events

savchukoleksii:
@stancl

### DIFF
--- a/assets/TenancyServiceProvider.stub.php
+++ b/assets/TenancyServiceProvider.stub.php
@@ -48,6 +48,9 @@ class TenancyServiceProvider extends ServiceProvider
                     return $event->tenant;
                 })->shouldBeQueued(false), // `false` by default, but you probably want to make this `true` for production.
             ],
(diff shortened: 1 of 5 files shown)


hyquality:
@stancl Can you review PR, please?
This pull request add tenant soft delete events system support.

savchukoleksii:
@stancl Can you review PR please?
This PR add ability to support soft/force deleting tenants.

stancl:
I'm not sure if this is needed. Generally people use custom mechanisms for disabling tenants, and soft deletes aren't IMO a good idea here.

That said, nothing is stopping the developer from adding soft deletes to the tenant model in their application.

savchukoleksii:
@stancl
My idea was to create build-in solution for this.

stancl:
I don't think it'd make much sense to be honest. It's very easy to add in the app, and chances are that you'd be using a more custom setup than this. I also don't want to encourage using soft deletes with tenants, as people should use custom logic that will handle everything they need.

PR #679, opened 2021-06-22
feat(UniversalRoutes): Stop overwriting the (maybe) customized onFail…

bemigho95:
With this change we still can customize the onFail method of the middleware and use the UniversalRoutes feature.


bemigho95:
@stancl No comment on this one?

stancl:
Ah sorry seems like I missed the notification. I'll review & merge this in a few weeks when I work on v4, thanks!

src/Features/UniversalRoutes.php
```diff
@@ -23,12 +23,14 @@ class UniversalRoutes implements Feature
     public function bootstrap(Tenancy $tenancy): void
     {
         foreach (static::$identificationMiddlewares as $middleware) {
-            $middleware::$onFail = function ($exception, $request, $next) {
+            $originalOnFail = $middleware::$onFail;
```
stancl, line 26: I believe this can be `null` sometimes? Which would make the `$originalOnFail()` call result in an error.

Though tests seem to be passing

abrardev99: You are right. When the Universal Route feature is enabled, and `onFail` logic is not provided, we get the `Value of type null is not callable` error instead of the `TenantCouldNotBeIdentifiedOnDomainException` exception.

abrardev99:
[Regarding this review](https://github.com/archtechx/tenancy/pull/679#discussion_r775169980), you are right. See my comment on the review. 

So, resolved the mentioned review and also added a test. Now it throws a correct exception when `$originalOnFail()` is `null` instead of the generic `Value of type null is not callable` error.

stancl:
The solution looks good. Regarding the test, I'd come up with a better name, the current one is long and a bit hard to understand.

Also, I'm thinking that I'll merge the pest PR first and then you can edit this to be a pest test.

stancl:
You can now adjust the DomainTest to use pest

abrardev99:
> You can now adjust the DomainTest to use pest

Done. 

> I'd come up with a better name, the current one is long and a bit hard to understand.

I've renamed the test to `throw correct exception when onFail is null and universal route is enabled`.

tests/DomainTest.php
```diff
@@ -90,25 +89,19 @@
         ->assertSee('foo');
 });
 
-    /** @test */
-    public function not_defining_onfail_customized_logic_throw_correct_exception_when_universal_route_feature_enabled()
-    {
-        $this->expectException(TenantCouldNotBeIdentifiedOnDomainException::class);
+test('throw correct exception when onFail is null and universal route is enabled', function () {
```
stancl:
```suggestion
test('throw correct exception when onFail is null and universal routes are enabled', function () {
```

stancl:
Is this ready to be merged now?

abrardev99:
> Is this ready to be merged now?

Yes. I don't see any reviews/comments to resolve.

Issue #658, opened 2021-05-18
Synced resources between tenants throws error when inserting new model on tenant


kr-apps:
#### Describe the bug
When using *Synced resources between tenants*, the method `updateResourceInCentralDatabaseAndGetTenants` on the listener `UpdateSyncedResource` trows the following error:

```
  SQLSTATE[42S22]: 
Column not found: 1054 Unknown column 'code' in 'field list' (SQL: insert into 
`users` (`name`, `email`,  `password`, `remember_token`, `code`, `global_id`) 
values (Hiram Hermann, yschumm@example.com, secret, fd9ca932-23c3-4e2f-abb6-38bc82cd92c4))
```

#### Steps to reproduce
Assuming we want to sync users:

1. Set up *Synced resources between tenants* following the documentation
2. Add some extra columns on the tenant's users migration, for example `code`
3. Try to add a NEW user on the tenant that doesn't exist on the central database

#### Expected behavior
Since the user doesn't exists on the central database the `updateResourceInCentralDatabaseAndGetTenants` method will try to insert the new user on the central database using all the attributes from the model on the tenant's database, but like the columns from the tenants are not the same the insert can't happen.

#### Your setup
 - Laravel version: 8.41.0
 - stancl/tenancy version:  3.4.4

stancl:
Does `getSyncedAttributeNames()` not work?

kr-apps:
> Does `getSyncedAttributeNames()` not work?

To make it work I override the method `updateResourceInCentralDatabaseAndGetTenants` in this specific part:

BEFORE
```
// If the resource doesn't exist at all in the central DB, we create
// the record with all attributes, not just the synced ones.
$centralModel = $event->model->getCentralModelName()::create($event->model->getAttributes());
event(new SyncedResourceChangedInForeignDatabase($event->model, null));
```

NOW
```
// If the resource doesn't exist at all in the central DB, we create
// the record with all attributes, not just the synced ones.
$centralModel = $event->model->getCentralModelName()::create($syncedAttributes); <----CHANGED THIS LINE
event(new SyncedResourceChangedInForeignDatabase($event->model, null));
```

Also I have to add `global_id` in the fields to be synced.

PiotrekPKP:
I have same error :c

stancl:
I guess there should be more configuration for how synced resources should be created, i.e. whether it should use all attributes, just the synced one, or something else. I'll add this in v4.

amjeed-ay:
[...]

PR #680, opened 2021-06-26
prevent sync of non-existent column and primary key collisions

backstageel:
When syncing resources, this line $centralModel = $event->model->getCentralModelName()::create($event->model->getAttributes()); fails in two situations:

1. when there are column(s) in tenant table that are not in central table.
2. when there is already a row in central table with same id as the id generated while creating the model on tenant table(With ID i mean primary key which my differ from Global ID column).

With this change, only syncedAttributes will be used while creating the model on central table. unfortunatelly, with this we lose the hability to populate the row with default values and all required columns must be on $syncedAttributes, but is the only way I could get it to work.

### DIFF
--- a/src/Listeners/UpdateSyncedResource.php
+++ b/src/Listeners/UpdateSyncedResource.php
@@ -59,7 +59,7 @@ class UpdateSyncedResource extends QueueableListener
                 event(new SyncedResourceChangedInForeignDatabase($event->model, null));
             } else {
                 // If the resource doesn't exist at all in the central DB,we create
-                // the record with all attributes, not just the synced ones.
+                // the record with only synced ones.
                 $centralModel = $event->model->getCentralModelName()::create($syncedAttributes);
                 event(new SyncedResourceChangedInForeignDatabase($event->model, null));
             }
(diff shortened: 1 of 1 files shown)


stancl:
See this https://github.com/archtechx/tenancy/blob/3.x/tests/ResourceSyncingTest.php#L409

There's no right or wrong implementation, it just depends on what you need. In v4 I want to make this more configurable.

I guess you could have a `creating` hook and unset the properties that you don't want.

backstageel:
Yes, the `creating` approach should solve my current problem.... I will close this PR

Thanks

backstageel:
Sorry seems like the creating approach will not work since events are not fired on this listener:
https://github.com/backstageel/tenancy/blob/0d423ecdddee767f3db269ff05b576c011e84cff/src/Listeners/UpdateSyncedResource.php#L56

But i will try another solutions that does not require this PR to be merged.

morloderex:
One solution is to add a second method and split up how synced columns are configured? 

Basically we could do 2 specific methods on the interface one for creating entirely new records and one for updating?

This would of cause be a major breaking change but it would work a bit better in v4 by given the developer complete control over just how the resources is being synced?

stancl:
Does this related to #658?

morloderex:
@stancl I believe it does sense i have hit the exact same issue once described in #658 already.

stancl:
So is this a duplicate of #658?

PR #682, opened 2021-06-30
Create AccessFromCentralDomains.php

webactor724:
add middleware run only central domain . invert PreventAccessFromCentralDomains  . i have module cms we need run cms route only in central domain


stancl:
What does this solve? And does it need to be part of the package? You can just use your own middleware. This would also need to be documented if it's added.

PR #683, opened 2021-07-08
Check cache store support for tags

MohamedGamil:
The issue arises when using a cache driver which doesn't support cache tags, therefore if the `tags` method wasn't explicitly called and the current cache driver doesn't support tags it shouldn't be invoked by default on all calls to mitigate any unwanted "This cache store does not support tagging." exception. 

This solves the issue for cache drivers which doesn't support tags such as the "file" driver.

### DIFF
--- a/src/CacheManager.php
+++ b/src/CacheManager.php
@@ -18,6 +18,7 @@ class CacheManager extends BaseCacheManager
     public function __call($method, $parameters)
     {
         $tags = [config('tenancy.cache.tag_base') . tenant()->getTenantKey()];
+        $supportTags = $this->store()->getStore() instanceof \Illuminate\Cache\TaggableStore;
 
         if ($method === 'tags') {
             if (count($parameters) !== 1) {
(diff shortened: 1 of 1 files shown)


stancl:
But this would result in the cache calls not being scoped at all (on non-redis drivers) when users expect them to be scoped, wouldn't it?

MohamedGamil:
It would if the driver supports it. It doesn't change the default behavior of the package, the changes only mitigate the exception if the driver doesn't support it and I don't think it'd be beneficial to restrict developers to drivers that support tags only. 

It should be documented to let developers make an educated decision since the modified `CacheManager` intercepts all cache calls by default.

stancl:
>  I don't think it'd be beneficial to restrict developers to drivers that support tags only

Simply disable the CacheTenancyBootstrapper in the config.

If I'm not misunderstanding, this PR would lead to developers expecting the scoping to take place (cache bootstrapper enabled), but it wouldn't do anything if a non-Redis driver is being used.

MohamedGamil:
> > I don't think it'd be beneficial to restrict developers to drivers that support tags only
> 
> Simply disable the CacheTenancyBootstrapper in the config.
> 

Sorry, I didn't know about this, I will apply the recommended solution instead.

PR #688, opened 2021-07-15
Add --force option to tenants:seed Artisan call

chillbram:
When creating a tenant using the SeedDatabase job in production I've been getting a `Undefined constant "STDIN"` error. Adding the --force flag to the tenants:seed resolves the issue of artisan trying to interact with the CLI when running in production (as per [Mohamed Said](https://divinglaravel.com/the-problem-behind-undefined-constant-stdin-when-running-laravel-artisan-command-programmatically)).

### DIFF
--- a/src/Jobs/SeedDatabase.php
+++ b/src/Jobs/SeedDatabase.php
@@ -33,6 +33,7 @@ class SeedDatabase implements ShouldQueue
     {
         Artisan::call('tenants:seed', [
             '--tenants' => [$this->tenant->getTenantKey()],
+            '--force' => true,
         ]);
     }
 }


stancl:
https://github.com/archtechx/tenancy/blob/3.x/assets/config.php#L195

chillbram:
Ah didn't see that in the config. But is there a reason this should not be turned on by default? Seems like it won't work otherwise on production and you won't really have an indication that it wouldn't work until after you deploy and start working in production.

stancl:
Seeders running in production by default would be dangerous. Easy to make a mistake and have your DB get filled with lots of made up data.

In general, even the practice of using seeders in prod is wrong.

PR #693, opened 2021-07-30
Setting $deleteWhenCannotIdentify property deletes the job when it could not be identified

morloderex:
Similiar to defining `deleteWhenMissingModels` when using regular models on a job class we can now specify that the job should not fail when tenancy could not be initialized for a given job when setting the property `deleteWhenCannotIdentify` on the job.

This is great when a job is being processed but for what ever reason the tenant has been deleted prior to processing the job.

See [here](https://laravel.com/docs/8.x/queues#ignoring-missing-models) for reference to what i am trying to implement for tenancy initialization

### DIFF
--- a/src/Bootstrappers/QueueTenancyBootstrapper.php
+++ b/src/Bootstrappers/QueueTenancyBootstrapper.php
@@ -7,6 +7,7 @@ namespace Stancl\Tenancy\Bootstrappers;
 use Illuminate\Support\Str;
 use Illuminate\Config\Repository;
 use Illuminate\Queue\QueueManager;
+use ReflectionClass;
 use Stancl\Tenancy\Contracts\Tenant;
 use Illuminate\Queue\Events\JobFailed;
 use Illuminate\Queue\Events\JobProcessed;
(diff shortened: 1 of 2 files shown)


preecedavid (approved):

preecedavid (approved):

stancl:
I can merge this in v3 if you could make one more addition. Are you able to work on this now?

morloderex:
> I can merge this in v3 if you could make one more addition. Are you able to work on this now?

What kind of addition do you want? :)

stancl:
I believe our current logic listens to `JobProcessing` but not `JobRetryRequested`. We should have a listener for that as well — similar to [this](https://github.com/spatie/laravel-multitenancy/blob/804ad0f1e41a5d6d9d306d383b08a054e790d9de/src/Actions/MakeQueueTenantAwareAction.php#L56).

I think that'd solve #762

morloderex:
@stancl Sure, i will tackle and get back to you :)

stancl:
Actually no need to do that now - I've just pushed a change here https://github.com/archtechx/tenancy/commit/73a4a3018cadca2ba0fb5f2130fca1718a2b3670

So you can just get your PR to work and I'll merge 👍🏻

stancl:
Any progress on this? Seems like the tests are still failing

thanhtoan1196:
any updates?

PR #707, opened 2021-08-19
Remove default table override

pet1330:
By setting a default table value, Eloquent cannot automatically detect the table name based on the model name. Unless the developer actively sets a table property. This can cause migration issues when using anything other than a model calledTenant. This PR removes the default value and lets eloquent detect the correct table (which, if using the default Tenant model, will default to the current value anyway).

primaryKey isn't an issue, but it defaults to eloquents set value anyway, so has also be removed

### DIFF
--- a/src/Database/Models/Tenant.php
+++ b/src/Database/Models/Tenant.php
@@ -28,8 +28,6 @@ class Tenant extends Model implements Contracts\Tenant
         Concerns\TenantRun,
         Concerns\InvalidatesResolverCache;
 
-    protected $table = 'tenants';
-    protected $primaryKey = 'id';
     protected $guarded = [];
 
     public function getTenantKeyName(): string


stancl:
I think there was some reason for this, maybe the fact that some people name tenants a bit differently but it's still best to use the `tenant` terminology on the database level

Closing this because the change is easy to do in app land.

pet1330:
@stancl, thanks for taking the time to review this! I've had a look and I can't see why this would be needed as the defaults have the same value. It just seems to disable the default Laravel behaviour, which can be quite confusing for newcomers.

There may be edge cases that I'm missing that would require these to be set, such as naming tenants a bit differently, but in those cases, developers would need to set the table value anyway, so for them, it would make no difference as this keeps that behaviour while also leaving the default Laravel behaviour in place?

If I still haven't managed to convince you, let me know, and I'll send a PR to update the docs with a warning to explicitly set the table attribute when using a custom model :smiley:

stancl:
> There may be edge cases that I'm missing that would require these to be set, such as naming tenants a bit differently, **but in those cases, developers would need to set the table value anyway**, so for them, it would make no difference as this keeps that behaviour while also leaving the default Laravel behaviour in place?

Not necessarily. You can keep the table name `tenants` but then use e.g. `Team` or `Seller` in your code to make the business logic more clear. Yes, the model layer and DB layer would use slightly different terms, but as long as you know that e.g. the `Team` is the tenant, it won't be confusing, and the benefit of using `Team` in code instead of `Tenant` is there when your app has a lot of business logic and this makes things a lot cleaner.

I believe the reason why it's hardcoded is that you'd have to change all of the `tenant` references in the codebase. So, `BelongsToTenant` column static property, other models which reference `tenant_id`, migrations which reference `tenant_id`, and things like that.

If someone wants to use a custom table name, it should be a very explicit change. Using a custom model name while keeping everything else default is supported with the current behavior.

Another thing is that the PR would be a breaking change, so it'd have to be a v4 change anyway.

I think the issue is easy to debug, but if you want to send a docs PR, that'll be appreciated :)

PR #715, opened 2021-09-06
[4.x] Added support for Microsoft Sql Server

nickakitch:
SQL Server is currently supported within Laravel, and this pull request provides support for users with this type of database


stancl:
Hey, thanks a lot for the PR. The code looks good. I'll merge & release this in a few weeks when I get v4 done.

Adding a note to self to add the reference to the config file and everywhere else where it needs to be.

stancl:
Wanted to say that this PR should change the build so that we're using this TenantDatabaseManager in tests, but noticed that you actually made the change already and it should be sufficient 👍🏻

stancl:
Could you resolve the conflicts in the test file? I can look into merging this before v4

stancl:
Any idea why the tests are failing?

stancl:
Seems like the `Dockerfile` is fine now. Only thing remaining is to add a MSSQL service since the current error is `Microsoft ODBC Driver 17 for SQL Server : Login timeout expired.`

stancl:
Still seems to be unable to connect to the service.

Note to myself:
- [x] Rename all `mssql` references to `sqlsrv` for consistency

The next step in debugging is using `docker-compose up -d` locally and trying to connect to the SQL server instance/checking its status using `docker-compose`.

stancl:
Thanks a lot!

PR #729, opened 2021-09-29
Prevent missing and duplicate field

MrJmpl3:
The documentation says a custom column needs to be declared in getCustomColumns function, that include 'id' (per default in Virtual Column package), 'created_at', 'updated_at'.

If  'created_at', 'updated_at' are not declare, that make a empty 'created_at' field and duplicate 'update_at' field, when created a tenant.

### DIFF
--- a/src/Database/Models/Tenant.php
+++ b/src/Database/Models/Tenant.php
@@ -46,6 +46,15 @@ class Tenant extends Model implements Contracts\Tenant
     {
         return new TenantCollection($models);
     }
+    
+    public static function getCustomColumns(): array
+    {
+        return [
+            'id',
+            'created_at',
+            'updated_at',
+        ];
+    }
 
     protected $dispatchesEvents = [
         'saving' => Events\SavingTenant::class,


stancl:
- Hardcoding `id` will break this when someone changes the column name
- `id` should be automatically included (I think)
- What is the actual issue here? That tenants have timestamps inside the virtual data?

MrJmpl3:
> 
> 
>     * Hardcoding `id` will break this when someone changes the column name
> 
>     * `id` should be automatically included (I think)
> 
>     * What is the actual issue here? That tenants have timestamps inside the virtual data?

I don't know if is a problem of my config or system, but when I create a tenant in the DB, the fields created_at and updated_at are null, and in the json of the virtual data have a update_at field

MrJmpl3:
@stancl Hi! A question about this PR. 

That tenants have timestamps inside the virtual data, that is correct?. 

Perhaps it is necessary to add a notice to the documentation, in Discord, some peoples also asked about this issue of the timestamps in the virtualcolumns.

stancl:
> That tenants have timestamps inside the virtual data, that is correct?

It's not: https://github.com/archtechx/tenancy/blob/3.x/assets/migrations/2019_09_15_000010_create_tenants_table.php#L23

When does it get set in `data`? I don't think that was happening in the past anyway. Maybe you're trying to set the timestamp directly? The automatically set data should be going into the right columns. They're not nullable so the model would fail without them anyway iirc

chrillep:
having similar issue. 'created_at', 'updated_at' gets set in data column and 'created_at', 'updated_at' columns

PR #740, opened 2021-10-15
Set Internal DB Connection Name

bogdankharchenko:
@stancl so I played around with it and it really seems like setting a DB connection name perhaps solves the issue at hand.  

Is there a reason why this was not included as part of the package initially?

Seems to me like it would be ideal to be able to "switch" the `template_tenant_connection` and have the previous tenants depend on the connection they were setup for.

I also may be missing the point that the #656 PR was trying to solve.

### DIFF
--- a/src/DatabaseConfig.php
+++ b/src/DatabaseConfig.php
@@ -78,6 +78,11 @@ class DatabaseConfig
         return $this->tenant->getInternal('db_password') ?? null;
     }
 
+    public function getDbConnectionName(): ?string
+    {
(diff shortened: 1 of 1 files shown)


bogdankharchenko:
@stancl thoughts?

stancl:
The point of setting db_connection is overriding the default template connection. Basically all these internal keys are meant for overrides

When it's just the default, then the tenant shouldn't need this

What exactly does this solve and if it's that linked PR, why is this better than the originally proposed PR?

PR #745, opened 2021-11-02
Add missing import for Domain model

JaZo:
I noticed there's a binding in the container for the current domain, but it's not working because the Domain class import is missing.


stancl:
Ah, nice catch. Thanks a lot for the PR!

PR #755, opened 2021-11-17
Use GitHub forms for issues template.

abrardev99:
[Updated]
The final output looks like the attached SS. 

image


stancl:
Looks great!

Issue #757, opened 2021-11-18
Tenant Maintenance Secret


aeq-dev:
#### Description
Same as the main domain where we can enable maintenance with secret key in order to access to the app for admins for example,
is there a way to do that for tenant's admins

#### Why this should be added
Give the ability to tenant's admins to access to the app

stein-j:
From what I can understand: The package doesn't uses the Laravel default maintenance mode, which doesn't surprises me because the way Laravel handle it is not tenant compatible.
This means this package would have to fully recreate Laravel maintenance mode, which could work... but not all features would be doable (like the --render option because this allows the render a page before Laravel boots, but for tenancy you need to boot Laravel).

_Small idea/lead:_ 
When Laravel is in maintenance mode, it can "except" routes defined in `PreventRequestsDuringMaintenance` middleware. Maybe the package could utilise this to use Laravel down feature and still allow some tenant. Though, I'm not convinced this is the best idea.

stancl:
In which case would some tenants be in maintenance mode, but others not?

stein-j:
@stancl to answer your question, I think there are many use case :

- A tenant that is used as a demo which is often reset, it needs to be put in maintenance mode temporally and alone
- You have a big update and want to put then into maintenance mode one by once instead of denying access to all tenants too long
- You have a tenant used for testing and do not want people to access it while you are working on it
- etc.

stancl:
Right, I expressed myself poorly. I meant, in which case would some tenants be in maintenance mode, but others not, with a need for access via the URL secret.

For example the "_You have a big update and want to put then into maintenance mode one by once instead of denying access to all tenants too long_" point is just selective tenant maintenance mode. And the first point probably doesn't need a secret.

I can see why having some tenant accessible only for testing is good though.

PR #761, opened 2021-12-02
[4.x] Add command to bring the tenants up and down from maintenance and remove deprecated exception

### LINKED ISSUE #533: Disabling maintenance mode, and CLI commands for maintenance mode

stein-j:
This PR changes the maintenance mode to comply to the changes made in Laravel 8.

**New**
- Adds two CLI commands to the package: `tenancy:up` and `tenancy:down` with the same parameters as the default Laravel command except for the option `render` because it is not possible to render a page before Laravel is booted for a specific tenant.

- Adds the function `bringUpFromMaintenance` in the trait `MaintenanceMode`.

**Changes**

A tenant in maintenance mode now throws an `HttpException` exception instead of the deprecated `MaintenanceModeException` exception. 

Closes #533.


stancl:
Hey. This looks good, thanks for the contribution! I think I had some maintenance mode related tasks written in my backlog for v4, so I'll probably merge this when working on the feature. Should be in January I think.

src/Commands/Down.php
```diff
@@ -0,0 +1,52 @@
...
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+
+    protected $signature = 'tenancy:down
+                                 {--time= : The time when the app has been set to maintenance mode}
+                                 {--message= : Message to display}
+                                 {--retry= : The number of seconds after which the request may be retried}
+                                 {--allowed=* : List of IPs allowed}';
```
stancl: This doesn't implement the same parameters as the default `artisan` down command, right?

For instance, the `secret` — https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Console/DownCommand.php#L23

stancl:
Closing #757 and will finish that here. We could probably add `--secret` but I'm not sure if there's a point in having a different secret than the central app.

stein-j:
Laravel now simply throws a `HttpException` exception. https://github.com/laravel/framework/blob/8f3b411969ffdd733a0ba54098e08f8ca7c1369e/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php#L78

If you give me a few days, I can update the code and comply to the new maintenance mode used by Laravel and then update the commands so that we can use the same parameters as [Laravel](https://github.com/laravel/framework/blob/5398fbbbf81ed842cf0430b278f07eb0f869cf8f/src/Illuminate/Foundation/Console/DownCommand.php#L19).
I think all parameters can be recreated except for `render` because this allows to return a response even before Laravel is initialised (which means, before the package is also initialised).

Note: Naturally this would allow #757

stancl:
Sure, thanks a lot for the work on this! FWIW if there's a signature change, you can use the version in Laravel 9. And probably change the target to the `master` branch.

stein-j:
I'm glade you mentioned Laravel 9 because coincidently two PRs ([1](https://github.com/laravel/framework/pull/40070), [2](https://github.com/laravel/framework/pull/40102)) have been accepted during those last few days and it changed the maintenance system.

L9 will now have a configurable driver for the maintenance storage system (which will determine how and where the data is stored), at the time of writing this comment there are two drivers/managers **file**: Which is the old way to put the application to maintenance ; **cache** : Which will store the data inside the cache.

The reason they added drivers, is because when using a load balancer and you set the application into maintenance mode it would only work on the application the command has been typed on... so a _cache_ driver was needed. But this package doesn't need such implementation because even if you would use a load balancer, there should be only one database connection **per tenant** for all applications (at least in most use cases).

So I think to keep things simple, this PR will keep its own feature and keep storing the data in the database. It might be important to mention in the documentation that the maintenance mode doesn't uses the driver from the config file.

Maybe in another PR we can try to create our custom driver, tenant compatible.

stancl:
> this PR will keep its own feature and keep storing the data in the database

That is in the `data` column in our case, right?

stein-j:
Yes

stancl:
Just a note: the tests will fail because `master` has some failing tests due to dependencies that need new syntax. But we can look at this feature's tests in isolation

stein-j:
@stancl the PR is done and all the tests have been added.
As you said there are conflict versions with composer, so the tests fails. Do you want me to look into it ? Or should we wait that the master to be updated ?

stancl:
Probably wait until master is updated 👍🏻 

When we add L9 support, I'll merge 3.x into master

stein-j:
Hello @stancl,

I just realized that the changes from this PR are not compatible with Laravel 6 & 7. I've put this PR back to draft, ping me when you start working on the V4 and I'll see what can be done at that point.

stancl:
@lukinovec I think we should finish this PR before the pending tenants since it's smaller and it interacts with the trait adding a `--tenants` option. That way we'll make the conflict resolution simpler.

Also, the tests are showing as failing, but I think they'll run differently when the CI is triggered again since I simplified the CI matrix in the `master` branch. So if you could resolve the conflicts, that would trigger a CI build and we'll see how the feature works now.

stancl:
Only php-cs-fixer is failing, tests seem to be passing, so CI-wise this PR is ready now.

I'll look at the code now and see if this is ready to be merged.

src/Commands/Down.php
```diff
@@ -0,0 +1,53 @@
...
+    protected $description = 'Put tenants into maintenance';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        tenancy()->runForMultiple($this->option('tenants'), function ($tenant) {
```
stancl: @lukinovec Does the `HasATenantsOption` trait *require* `--tenants` or make it optional?

If it's optional, then it's possible that `$this->option('tenants')` will return `null`. Which would run this for **all** tenants.

Which can be the desired behavior, and probably is the desired behavior, but it's not immediately clear from looking at the code. So if you can confirm what I wrote above, we probably want to add a comment like this:
```suggestion
    {
        // This runs for all tenants if no --tenants are specified
        tenancy()->runForMultiple($this->option('tenants'), function ($tenant) {
```

stancl: And you can probably do the same in `Up.php`

stancl:
@stein-j Is anything unimplemented yet or is the PR fully ready? Is the code made to work with Laravel 9 (the changes you made to the way tenants are put into maintenance mode)?

stein-j:
@stancl I want to go over it again, it's been a while and a file was changed in another PR since (https://github.com/archtechx/tenancy/pull/802/files#diff-40487242039e9842dc87ba96285282e42d0cc5b9a7dbeac6f115c4ee2bb8db59). Let me get back to you in a few hours.

stein-j:
@stancl what will be the minimum requirements for Laravel and PHP for V4 ?

stancl:
Just that the syntax is updated for Laravel 9. I *think* you already did this since the PR is recent. But I'm just checking in case anything is missing.

stein-j:
Ok, it's all good.

src/Commands/Down.php
lukinovec: Yeah, the option is optional, and if not specified, the code runs for all tenants

lukinovec: I added the comments

stancl:
FYI I merged `master` into this so you should delete the `3.x` branch after we merge this since it isn't compatible with our `3.x` (in case you plan on making PRs to `3.x` in the future)

stancl:
Thanks for your work on this PR!

PR #764, opened 2021-12-14
Change tenant connection name in DatabaseManager

davidrushton:
I'm integrating with an existing project that requires the tenant connection to be called something else. I can get this working in most parts of the code, including by changing the `tenancy.database.template_tenant_connection` config value.

However the `Stancl\Tenancy\Database\DatabaseManager` class seems to hard-code the connection name, causing `SQLSTATE[3D000]: Invalid catalog name: 1046 No database selected exceptions on migration, etc.

This PR replaces the hard-coded tenant connection name with `$tenant->database()->getTemplateConnectionName()`. 

Hope that is acceptable :)

### DIFF
--- a/src/Database/DatabaseManager.php
+++ b/src/Database/DatabaseManager.php
@@ -38,9 +38,9 @@ class DatabaseManager
      */
     public function connectToTenant(TenantWithDatabase $tenant)
     {
-        $this->database->purge('tenant');
+        $this->database->purge($tenant->database()->getTemplateConnectionName());
         $this->createTenantConnection($tenant);
-        $this->setDefaultConnection('tenant');
+        $this->setDefaultConnection($tenant->database()->getTemplateConnectionName());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stancl:
I think the template connection name and tenant connection name are intentionally separate things. The tenant connection shouldn't modify the template.

Why do you need the connection to have a different name?

davidrushton:
@stancl Thanks for the quick reply! Our existing project doesn't refer to tenants as a `tenant` and there are external third party scripts that would need updated and tested to match that new database connection, so having that cutomisable would be ideal. If it's core to the package however, I understand. Thanks :)

stancl:
I think there's another issue which suggested making the name customizable, so it might be a feature in v4. But we can't use the template connection name — see the failing tests.

With regards to your issue, I think renaming a DB connection inside Laravel shouldn't be a problem. If it's repetitive, you can make it a trait, and I don't think 3rd party scripts should have any effect on this? Since the name of one DB connection can differ from the standard terminology throughout your app. Obviously I don't know the details of your app, but these are the suggestions I can give. If the connection name is the only issue, see if you really can't work around having it named `tenant`.

davidrushton:
Ok, thanks for the review anyway!

PR #768, opened 2021-12-23
Use tenant key on console commands instead of id

sort72:
In some cases 'id' would not be used (For example, assume one has a key named slug and deleted id column, commands would not recognize tenant on console comment, this simple PR fixes that).


stancl:
Thanks!

PR #775, opened 2022-01-06
Remove redundant initialization

stein-j:
I noticed the run command initializes the tenant, but the `runForMultiple` function already takes care of that which means when running `tenants:run` each tenant is initialized twice.

Please double check that I'm not changing something else by removing this.


stancl:
Ah, nice catch. Seems like that was left over from before we had the `runForMultiple()` method.

PR #780, opened 2022-01-14
Add --tenant option to the make:migration command

viicslen:
Add `--tenant` option to the default MigrateMake command provided by Laravel. This option will create the migration file inside of the tenant migrations directory

### DIFF
--- a/src/Commands/MigrateMake
+++ b/src/Commands/MigrateMake
@@ -1,0 +1,39 @@
+<?php
+
+namespace Stancl\Tenancy\Commands;
+
+use Illuminate\Database\Console\Migrations\MigrateMakeCommand;
+use Illuminate\Database\Migrations\MigrationCreator;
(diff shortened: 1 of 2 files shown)


stancl:
Hey, thanks for wanting to contribute to Tenancy.

I think I'm not going to merge this, because it adds significant overhead and potentially breaks support of other packages that replace the migrator for a very minor syntax change.

This currently works:
```bash
$ art make:migration --path=database/migrations/tenant create_foo_table
Created Migration: 2022_01_14_203655_create_foo_table
$ ls database/migrations/tenant
2022_01_14_203655_create_foo_table.php
```

And the difference between `--tenant` and `--path=...` isn't big. Further, the `tenant/` directory is just a default. Tenant migrations [can be elsewhere as well](https://github.com/archtechx/tenancy/blob/3.x/assets/config.php#L187).

PR #782, opened 2022-01-17
[3.x] Add pending tenants

stein-j:
As discussed here #779, this PR adds a new feature to this package: **pending tenants**.

This adds:
- Four new config settings: `pending.include_in_scope`, `pending.count`, `pending.older_than_days` and `pending.older_than_hours`
- Two new commands (`tenants:pending` and `tenants:pending-clear`)
- A `hasPending` trait that contains a pendingScope which was inspired by the `SoftDelete` trait.

All of this can be combined with the scheduler to automatically have x amount of tenants ready to be used and link to a new client/domain and skip the time of creating and migrating (and seeding) the client database.


stein-j:
I don't understand why this includes other changes, did I mess up something ?

stein-j:
Ok, I simply created my branch from the wrong original branch, my bad.

This shouldn't lead to any breaking changes, so either I change the target to 3.x or I redo a PR to target the master

stancl:
Yeah, v4 follows v3 so it should be fine

stein-j:
Hello @stancl, what do I need to change to run the tests with Laravel 6 with your docker-compose ? This will help me understand why the test failed.

stancl:
You can see this on GH:
CleanShot 2022-01-18 at 22 49 12@2x

https://github.com/archtechx/tenancy/runs/4857115025?check_suite_focus=true#step:5:20

And locally, you only need to change your `composer.json` to require Laravel 6. `composer require --no-interaction "laravel/framework:^6.0"` would probably work. Just make sure you don't commit that file afterwards.

And if you wanted to change the PHP version too, you could do it similarly to our CI workflow: https://github.com/archtechx/tenancy/blob/f08e33afd80d14d99eb0595261cb100f981d4b52/.github/workflows/ci.yml#L23-L26

stein-j:
I did check the logs of the CI, but it only show what test has failed so I can't debug with that.
And locally I did try your solution, even with `--with-all-dependencies` but I keep getting composer conflicts, and I don't know how to resolves those.

stein-j:
Anyhow, this PR is pretty much ready. You can start checking the code if you want to make sure this complies with your expectations for your package.

The only thing I'm not sure, and I need your feedback on. Is if you set the config `include_in_scope` (might rename to `include_in_queries`) to `false` then all of the commands (e.g migrate even with option `--tenants=xx`) will never run on the readied tenants.

I would like to give an option to bypass and include those tenants but the only solution I could think of is adding a third parameter on this [function](https://github.com/archtechx/tenancy/blob/f08e33afd80d14d99eb0595261cb100f981d4b52/src/Tenancy.php#L138) `$withReadied`. But this seems king of weird to do it like that.

stancl:
Hey, I'll do an in-depth review later this week 👍🏻  Feel free to push any more changes in the meantime if there are things you'd still like to change

stancl:
I'm thinking what'd be a good term for this feature. "readied" is probably not it, and neither are "standby" or "premade" tenants. There might be some standard term, I'll see if I can find it

stein-j:
Yeah, I don't like either the word _readied_ because it's not attractive/sexy. But I haven't found another good term. How about "on demand" or "deploy margin" ?

stein-j:
Hey @stancl,
I like the the term _standby_, I can rename everything if you're ok with it. 

Did you have time to have a look at the PR ? No rush, simply if any changes are needed I want to make sure I have time to work on it before v4.
I know the tests under PHP 6.0 are failing, I'll give it another try soon to debug  but other that that everything works

stancl:
So what all needs to be finished here?
- terminology
- failing tests
Anything else?

I'll probably review the code in depth in the next few weeks when I'll be working on Tenancy

stein-j:
Exactly, I also have two questions:

1) You can, by default exclude or include pending tenants on the query (global scope). When enabled, when running a command with `tenants:run` or `tenants:migrate`, pending tenants will be ignored. I wanted to add an option _--with-pending_ but because the query is not made in the commande, it causes issues. Do you have any suggestion for the best approche ?

2) The attribute to identify a pending tenant is currently in the json _data_ column. But this means the query is not optimized, should this be something to worry about or should I create another column ?

stancl:
> I wanted to add an option --with-pending but because the query is not made in the commande, it causes issues. Do you have any suggestion for the best approach

What do you mean by this?

> The attribute to identify a pending tenant is currently in the json data column

We'll keep it in the data column, since it's an extra feature.

stein-j:
Ok all done with this PR. I've renamed everything with the the term "pending", if for whatever reason you find another word you seem more fitting, let me know and I'll change it, it's an easy change.

I've added the possibility to include pending tenants when running `tenants:migrate` or `tenants:run` etc.

It was very annoying to understand why it didn't work under Laravel 6, there was a bug (fixed in L7) that prevented the query to detect null in a json column. I has to do a workaround but I'm not sure how much it affect the query performance cd70b57b83eb83a8cd8fcf2f3d9ee8b0841831f6

stancl:
See #869

PR #792, opened 2022-02-07
Run tests on PHP 8.1

erikgaal:
Currently the tests are not yet being run on PHP 8.1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


stancl:
This change will fail because Laravel 6 doesn't support 8.1 iirc. And it creates a large amount of jobs.

Perhaps we can just change the `8.0` to `8.1`? That way we'll know that it works on both PHP 7 and PHP 8.

erikgaal:
We can exclude the PHP 8.1 and Laravel 6 combination to make sure it won't run.

PR #793, opened 2022-02-08
[WIP] Use a lighter Docker image, use it in CI

erikgaal:
As discussed sometime on Discord I'm opening a PR to use Github Actions native services and `shivammathur/setup-php`. This way we don't have to maintain a Dockerfile for this project anymore and immediately allows us to test on PHP 8.1 (so #792 is redundant).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,12 +66,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Setup PHP
-        uses: shivammathur/setup-php@v2
-        with:
-          php-version: 8.1
-          coverage: pcov
-
       - name: Get composer cache directory
         id: composer-cache
         run: echo "::set-output name=dir::$(composer config cache-files-dir)"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,9 @@ env:
 
 on:
   push:
-    branches: [ 3.x, 2.x, master ]
+    branches: [ 3.x, 2.x ]
   pull_request:
-    branches: [ 3.x, 2.x, master ]
+    branches: [ 3.x, 2.x ]
 
 jobs:
   tests:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
 
     strategy:
       matrix:
-        php: ['8.1']
         laravel: ['^8.0', '^9.0']
 
     services:
@@ -70,8 +69,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, sqlsrv, pdo_sqlsrv, redis, bcmath, soap, intl, gd, exif, iconv, imagick, fileinfo
+          php-version: 8.1
           coverage: pcov
 
       - name: Get composer cache directory
@@ -82,13 +80,12 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ steps.composer-cache.outputs.dir }}
-          key: ${{ runner.os }}-${{ matrix.php }}-${{ matrix.laravel }}-${{ hashFiles('**/composer.lock') }}
+          key: ${{ runner.os }}-${{ matrix.laravel }}-${{ hashFiles('**/composer.lock') }}
 
       - name: Install Composer dependencies
         run: |
           composer require "laravel/framework:${{ matrix.laravel }}" --no-interaction --no-update
           composer update --prefer-dist --no-interaction
-
       - name: Run tests
         run: ./vendor/bin/pest
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,7 @@ jobs:
         env:
           ACCEPT_EULA: Y
           SA_PASSWORD: P@ssword
+        options: --health-cmd "echo quit | /opt/mssql-tools/bin/sqlcmd -S 127.0.0.1 -l 1 -U sa -P P@ssword"
 
       redis:
         image: redis
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           composer require "laravel/framework:${{ matrix.laravel }}" --no-interaction --no-update
           composer update --prefer-dist --no-interaction
       - name: Run tests
-        run: ./vendor/bin/pest
+        run: composer test
         env:
           DB_PASSWORD: password
           DB_USERNAME: root
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,16 +21,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Get composer cache directory
-        id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-
-      - name: Cache composer dependencies
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.composer-cache.outputs.dir }}
-          key: ${{ runner.os }}-${{ matrix.laravel }}-${{ hashFiles('**/composer.lock') }}
-
       - name: Install Composer dependencies
         run: |
           composer require "laravel/framework:${{ matrix.laravel }}" --no-interaction --no-update
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,10 +85,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ steps.composer-cache.outputs.dir }}
-          key: ${{ runner.os }}-${{ matrix.php }}-${{ matrix.laravel }}--${{ hashFiles('**/composer.lock') }}
+          key: ${{ runner.os }}-${{ matrix.php }}-${{ matrix.laravel }}-${{ hashFiles('**/composer.lock') }}
 
-      -if: steps.composer-cache.outputs.cache-hit != 'true'
-        name: Install Composer dependencies
+      - name: Install Composer dependencies
         run: |
           composer require "laravel/framework:${{ matrix.laravel }}" --no-interaction --no-update
           composer update --prefer-dist --no-interaction
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,9 @@ name: CI
 
 on:
   push:
-  pull_request:
-    branches: [ 3.x, 2.x, master ]
+env:
+  COMPOSE_INTERACTIVE_NO_CLI: 1
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   tests:
@@ -107,3 +108,19 @@ jobs:
         uses: codecov/codecov-action@v2
         with:
           token: 24382d15-84e7-4a55-bea4-c4df96a24a9b
+
+  php-cs-fixer:
+    name: Code style (php-cs-fixer)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install php-cs-fixer
+        run: composer global require friendsofphp/php-cs-fixer
+      - name: Run php-cs-fixer
+        run: $HOME/.composer/vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php
+      - name: Commit changes from php-cs-fixer
+        uses: EndBug/add-and-commit@v5
+        with:
+          author_name: "PHP CS Fixer"
+          author_email: "phpcsfixer@example.com"
+          message: Fix code style (php-cs-fixer)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+  pull_request:
+    branches: [ 3.x, 2.x, master ]
 env:
   COMPOSE_INTERACTIVE_NO_CLI: 1
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -94,7 +96,7 @@ jobs:
           composer update --prefer-dist --no-interaction
 
       - name: Run tests
-        run: ./vendor/bin/phpunit
+        run: ./vendor/bin/pest
         env:
           DB_PASSWORD: password
           DB_USERNAME: root
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on:
   pull_request:
     branches: [ 3.x, 2.x, master ]
 env:
-  COMPOSE_INTERACTIVE_NO_CLI: 1
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           composer require "laravel/framework:${{ matrix.laravel }}" --no-interaction --no-update
           composer update --prefer-dist --no-interaction
       - name: Run tests
-        run: composer test
+        run: ./vendor/bin/pest
         env:
           DB_PASSWORD: password
           DB_USERNAME: root
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,13 +14,8 @@ jobs:
 
     strategy:
       matrix:
-        php: ['7.4', '8.0', '8.1']
-        laravel: ['^6.0', '^8.0', '^9.0']
-        exclude:
-          - php: '8.1'
-            laravel: '^6.0'
-          - php: '7.4'
-            laravel: '^9.0'
+        php: ['8.1']
+        laravel: ['^8.0', '^9.0']
 
     services:
       postgres:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,12 @@
 name: CI
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 on:
   push:
   pull_request:
-    branches: [ 3.x, 2.x, master ]
-env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    branches: [ master ]
 
 jobs:
   tests:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
         TENANCY_TEST_PGSQL_HOST: postgres
         TENANCY_TEST_REDIS_HOST: redis
 
-    - name: Send code coverage to codecov
-      env:
-        CODECOV_TOKEN: 24382d15-84e7-4a55-bea4-c4df96a24a9b
-      run: bash <(curl -s https://codecov.io/bash)
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v2
+      with:
+        token: 24382d15-84e7-4a55-bea4-c4df96a24a9b
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,31 +1,97 @@
 name: CI
 
-env:
-  COMPOSE_INTERACTIVE_NO_CLI: 1
-
 on:
   push:
-    branches: [ 3.x, 2.x, master ]
   pull_request:
     branches: [ 3.x, 2.x, master ]
 
 jobs:
   tests:
     runs-on: ubuntu-latest
+    container: shivammathur/node:latest
 
     strategy:
       matrix:
-        php: ["7.4", "8.0"]
-        laravel: ["^6.0", "^8.0"]
+        php: ['7.4', '8.0', '8.1']
+        laravel: ['^6.0', '^8.0']
+        exclude:
+          - php: '8.1'
+            laravel: '^6.0'
+
+    services:
+      postgres:
+        image: postgres:latest
+        env:
+          POSTGRES_PASSWORD: password
+          POSTGRES_USER: root
+          POSTGRES_DB: main
+        ports:
+          - 5432/tcp
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 3
+
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: false
+          MYSQL_ROOT_PASSWORD: password
+          MYSQL_DATABASE: main
+        ports:
+          - 3306/tcp
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+      mysql2:
+        image: mysql:5.7
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: false
+          MYSQL_ROOT_PASSWORD: password
+          MYSQL_DATABASE: main
+        ports:
+          - 3306/tcp
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+      redis:
+        image: redis:alpine
+        ports:
+          - 6379/tcp
+        options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Start docker containers
-      run: PHP_VERSION=${{ matrix.php }} docker-compose up -d
-    - name: Install dependencies
-      run: docker-compose exec -T test composer require --no-interaction "laravel/framework:${{ matrix.laravel }}"
+    - name: Checkout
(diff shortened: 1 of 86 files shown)


stancl:
The Dockerfile is primarily for local development, and the CI setup uses it as well, to keep the dev & CI environments consistent.

That said, I'm open to changing these. I'm looking at the original Dockerfile and trying to find what was the main purpose of it.

Seems like:
- variable PHP version (it's fine to only have this in CI using `shivammathur/setup-php`)
- phpredis extension
- pcov for coverage — not sure if we're still using this. I think there are now better ways to generate coverage reports? I'm fine with moving to something else. But I think the original purpose was to make coverage reports from phpunit faster 
- pgsql & other DB extensions, I assume these are in `shivammathur/setup-php` as well

If these can be replaced with `shivammathur/setup-php`, we can make the switch and get rid of the Dockerfile altogether.

Seems like someone has an issue with building it as well (#794), so moving to this setup is probably the right decision.

Locally, there'd only be a `docker-compose.yml` file for starting all of the services used by tests.

The only downside of removing the `Dockerfile` is that we'd have *some* demands for the local setup of people who clone this package. Namely, that they have phpredis set up. But with a good CI setup, it's probably irrelevant for minor contributions.

So, in this PR I'm wondering if:
1. the CI file could use `docker-compose.yml`, instead of defining the services manually
2. `shivammathur/setup-php` can replace all of the dependencies we have in our `Dockerfile`: phpredis, possibly pcov, and the DB extensions

rawmarius:
This setup is nice and convenient (also great performance) but if you allow me I think it's a great advantage to support a docker or a sail setup for developers who want to keep their local installation "clean" (just not install anything locally). It would also help to test with more databases more easily, for example testing with mysql 8.

stancl:
I think it's not *that* important because it's only relevant for people contributing to the package. And the system requirements are common enough. But ideally, there'd be a simple Dockerfile shipped, yeah.

erikgaal:
Unfortunately there's no Sail support for package development, that would be cool to have.

As far as I can see there's no way to easily run `setup-php` in your own Dockerfile, so you can mirror the Github Action environment. I would suggest to update the Dockerfile to use the `shivammathur/node` image and enable the extensions that we require manually.

Would you like that to be a separate PR, so we can merge this? @stancl

stancl:
@erikgaal Does `setup-php` support the PHP extensions we had in our original Dockerfile?

erikgaal:
Yep, `redis` and `pcov` are enabled. :)

.github/workflows/ci.yml
```diff
@@ -1,32 +1,98 @@
...
+    - name: Get composer cache directory
+      id: composer-cache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+    - name: Cache composer dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+        restore-keys: ${{ runner.os }}-composer-
```
stancl: What's this step for?

.github/workflows/ci.yml
```diff
@@ -1,32 +1,98 @@
...
+
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: false
+          MYSQL_ROOT_PASSWORD: password
+          MYSQL_DATABASE: main
+        ports:
+          - 3306/tcp
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
```
stancl, line 63: What are these pings used for?

stancl:
Cool, in that case we can simplify the Dockerfile to use `node` as you mentioned. And use `setup-php` in GitHub Actions since they're similar enough.

Ideally, I'd like to do this all in one PR.

Aside from the Dockerfile, it'd be great if we could reuse the `docker-compose.yml` in GH actions as I mentioned above. And as for PHP 8.1, I'd go with a matrix that runs:
- L6 @ 7.4
- L8 @ 8.0
- L9 @ 8.1

This way, all three PHP versions are covered, and the most relevant Laravel versions are covered.

.github/workflows/ci.yml
erikgaal: Checks the health of the services so that the "main action" only starts after all services are healthy.

.github/workflows/ci.yml
erikgaal: This saves the composer dependencies to the cache to speed up next runs.

.github/workflows/ci.yml
stancl: Makes sense

.github/workflows/ci.yml
stancl: Is the cache PHP/Laravel version-specific?

erikgaal:
Unfortunately, I don't really have time to finish this PR shortly. If someone wants to continue, feel free to do so!

stancl:
Remaining tasks here:
- [x] 1. See if the `Dockerfile` can be rewritten **from scratch** to use `shivammathur/node` as mentioned here https://github.com/archtechx/tenancy/pull/793#issuecomment-1033577726. **Note: I recently added *more* dependencies to the Dockerfile here https://github.com/archtechx/tenancy/commit/cc6d4fe0ddcec405f645fbd2d8e90ab71afb1bec#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557**. I'm not 100% sure if we can migrate to `node`. It'd be ideal if we could, though.
- [x] 2. Verify that the variables used here https://github.com/archtechx/tenancy/pull/793#discussion_r805854785 will work even if we have a matrix of multiple PHP/Laravel versions. Either the variables should not be shared between different GH Actions environments, or the keys should be modified to include the versions in them
- [x] 3. Update the matrix to run the 3 jobs described here https://github.com/archtechx/tenancy/pull/793#issuecomment-1034026507

abrardev99:
Updated

>  **Note**: I merged the `3.x` branch into this PR to resolve the conflicts.

## Dockerfile
I've rewritten the Dockerfile and used the **node** container as the base and enabled some extensions manually like redis and sqlsrv. **node** installed PHP versions from `5.6` to `8.1`. To enable extensions, we had to use the PHP version with them like `/etc/php/8.1/mods-available/redis.ini`; otherwise, the build failed. 

This image was published here https://hub.docker.com/repository/docker/abrardev/tenancy, and I'm pulling the image in CI. 

## CI
I've improved the CI to run to multiple Laravel and PHP versions. Also, added the sqlsrv service. 

## Todo (Done)
I think we still can improve the Dockerfile by spitting the PHP version extensions enabling part. Also, it Should accept the PHP_VERSION as the container argument. 

## Todo
Published Docker image under ArchTech docker hub account.

Dockerfile
```diff
@@ -1,9 +1,7 @@
-ARG PHP_VERSION=7.4
-ARG PHP_TARGET=php:${PHP_VERSION}-cli
+# add amd64 platform to support Mac M1
+FROM --platform=linux/amd64 shivammathur/node:latest-amd64
```
stancl, line 2: If we're using `latest`, does this even respect `PHP_VERSION`?

stancl:
Are all of these three points solved? https://github.com/archtechx/tenancy/pull/793#issuecomment-1167829755

stancl:
@abrardev99 could you also fix the missing health check in `docker-compose.yml`?

Dockerfile
abrardev99: > If we're using latest, does this even respect PHP_VERSION?

No. 

I don't think we can pass the PHP_VERSION to `shivammathur/node`. It installs PHP from `5.6` to `8.1` automatically. But we need `PHP_VERSION` to enable extensions.

abrardev99:
> Are all of these three points solved? https://github.com/archtechx/tenancy/pull/793#issuecomment-1167829755

All of the points are completed, but I'm not sure about `2`. I think the cache key is unique based on composer.json. 
`key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}`.

abrardev99:
> could you also fix the missing health check in docker-compose.yml?

I don't think `docker-compose.yml` has a missing health check in this branch.

Dockerfile
stancl: > It installs PHP from 5.6 to 8.1 automatically

So it installs **all** PHP versions? That seems weird. I thought there are version-specific images.

stancl:
The missing healthcheck is in `3.x` I think.

stancl:
But if that part of the code isn't affected by this PR, you can just ignore it and resolve it later. Will probably be easier and result in fewer conflicts.

Dockerfile
abrardev99: > So it installs all PHP versions?

I even checked the container source code, [You can see it is installing all of the PHP versions. ](https://github.com/shivammathur/node-docker/blob/main/Dockerfile_bionic#L88)

stancl: I see. Aren't there version-specific images?

abrardev99:
> The missing healthcheck is in 3.x I think.

It's not in `3.x` either. It's in the `master` branch where `mssql` changes got merged, and I think only `mssql` has missing health checks.

Dockerfile
abrardev99: > Aren't there version-specific images?

I couldn't find any.

stancl: Okay, that's fine then. We can use `node` with this approach

abrardev99: Resolving the review then.

.github/workflows/ci.yml
stancl: @abrardev99 I think you addressed this, but the review is still open. Is this resolved now? If so - how?

stancl:
> I think the cache key is unique based on composer.json.

This would mean that it's not unique on each PHP/Laravel version.

.github/workflows/ci.yml
abrardev99: Putting the [below comments conversation](https://github.com/archtechx/tenancy/pull/793#issuecomment-1197044280) here. 

>me: I think the cache key is unique based on `composer.json. key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}`.

>you: This would mean that it's not unique on each PHP/Laravel version.

I think it is unique based on the `composer.lock` file because each Laravel installation would change the `composer.lock` so a new cache is automatically created. 
I'm getting this point from the [docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-contexts-to-create-cache-keys).
image

If you still think this is not working correctly, we can use the `matrxi.key` combination.

stancl:
Actually, if it's based on composer.**lock**, it might be unique since different PHP & Laravel versions will install different packages. Though I'd still like to be sure about this, or that there is some completely unique random value in each `composer.lock`. Which I don't think there is.

abrardev99:
Did you see my reply to the review? I explained there.

https://github.com/archtechx/tenancy/pull/793#discussion_r931897332

stancl:
I did see it. My comment addresses the potential issues with using `composer.lock` which is what that reply mentioned.

abrardev99:
I see. So then, we can use the "matrix" config with different Laravel and PHP versions.

stancl:
Not 100% sure now, do we want this in `3.x` for any specific reason @abrardev99? Or can we change the target to `master`?

I'd rather make this change as part of v4.

stancl:
> So then, we can use the "matrix" config with different Laravel and PHP versions.

Is this implemented now?

abrardev99:
> Is this implemented now?

Not yet.

> can we change the target to master? I'd rather make this change as part of v4.

Yes, we can make it to `v4`.

abrardev99:
> Is this implemented now?

It's got implemented.

stancl:
Can you resolve the conflicts? For `CONTRIBUTING.md` you can just use what's in `master`, for `ci.yml` you can use what's in this PR (though **do** check if there are any things in `master` that we may want to add back).

abrardev99:
@stancl The checks are failing. Looks like `master` got some [changes](https://github.com/archtechx/tenancy/blob/master/src/Enums/LogMode.php) like `enum`, which works only on PHP `8.1`.  So we can't really test below `8.1`. 

So now, checks pass only when the PHP version is `8.1` on `L8` and `L9`. We can't test `L6` because of PHP version limitations. 

Let me know what you want now to make it compatible with previous PHP versions, Or you might want to keep it as it is now.

stancl:
Yeah, `master` is meant to only run on 8.1 :) v4 will use the latest version of Laravel and PHP.

abrardev99:
I see. Then checks are good now.

.github/workflows/ci.yml
```diff
@@ -1,32 +1,109 @@
...
-      run: ./test
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, sqlsrv, pdo_sqlsrv, redis, bcmath, soap, intl, gd, exif, iconv, imagick, fileinfo
+          coverage: pcov
```
stancl: Can't this be done directly inside the docker container? That's sort of the point of this PR, that we will be using a Docker container made exactly for our CI. We shouldn't be reconfiguring it in a separate step, it should be already fully configured in the Dockerfile.

stancl:
Seems like this improves the CI speed by ~30 seconds.

Previous:
CleanShot 2022-08-08 at 6 09 15@2x

Current:
CleanShot 2022-08-08 at 6 09 29@2x

I think if we moved all of the php setup into the container, as I just wrote in the review above, the time should be cut down to around a half.

.github/workflows/ci.yml
abrardev99: You are right. We shouldn't need these extensions here when they are already enabled in Dockerfile.

abrardev99:
> I think if we moved all of the php setup into the container, as I just wrote in the review above, the time should be cut down to around a half.

What! we are down to 7s
Screenshot 2022-08-09 at 3 15 46 PM

stancl:
Do we need the setup PHP step at all? If we are using the Docker container? What does the step actually do?

abrardev99:
> Do we need the setup PHP step at all?

Yes, otherwise, it does not find a composer. Also, I think we need these for coverage.

stancl:
Can you expand on that? I'm not sure what you mean.

I think the php version should come from an env var passed to the docker container, I don't see what the setup php step would be needed for.

What does the step **actually do**?

abrardev99:
> I don't see what the setup php step would be needed for.

I tried removing it, and I got a [composer not found error](https://github.com/abrardev99/tenancy/runs/7760897781?check_suite_focus=true). 

Can we install composer in Dockerfile and simply use it in CI?

stancl:
Very likely we can. That's why I asked you what the step is exactly for. You should find that. If it only sets up composer, then we can simply set up composer in the docker image, properly handle the PHP version variable, and simplify the CI steps here and **make the Dockerfile completely usable for local tests as well**.

The bold part is also a point of this PR, we don't just want to make CI work, we want the image to work great for local tests.

Right now, we have a Dockerfile that makes the tests always run in the same context. It's used in the CI and it's used locally. In this PR, we are improving the Dockerfile and optimizing the CI for speed, but we still want the Dockerfile to be the authoritative environment in which the tests can be run locally.

.github/workflows/ci.yml
```diff
@@ -1,32 +1,100 @@
...
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: [ 3.x, 2.x, master ]
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
stancl: @abrardev99

- Revert the `branches` change, only use `master`
- Move `env` back to the top of the file

.github/workflows/ci.yml
```diff
@@ -1,32 +1,100 @@
...
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-${{ matrix.laravel }}-${{ hashFiles('**/composer.lock') }}
+
+      - name: Install Composer dependencies
+        run: |
+          composer require "laravel/framework:${{ matrix.laravel }}" --no-interaction --no-update
+          composer update --prefer-dist --no-interaction
+      - name: Run tests
+        run: ./vendor/bin/pest
```
stancl: Try changing this to `composer test` and see if it works

.github/workflows/ci.yml
```diff
@@ -1,32 +1,100 @@
...
+
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache composer dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-${{ matrix.laravel }}-${{ hashFiles('**/composer.lock') }}
```
stancl: Just to confirm: is it tested that this definitely speeds up the CI?

.github/workflows/ci.yml
```diff
@@ -1,32 +1,100 @@
...
+          SA_PASSWORD: P@ssword
+        options: --health-cmd "echo quit | /opt/mssql-tools/bin/sqlcmd -S 127.0.0.1 -l 1 -U sa -P P@ssword"
+
+      redis:
+        image: redis
+        ports:
+          - 6379/tcp
+        options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
 
     steps:
```
stancl: Can you put `steps` above `services`?

Dockerfile
```diff
@@ -1,9 +1,7 @@
...
 
-FROM --platform=linux/amd64 ${PHP_TARGET}
-
-ARG COMPOSER_TARGET=2.0.3
+ARG PHP_VERSION=8.1
```
stancl, line 4: We had another conversation about this, but does this now respect the PHP version? The value is hardcoded here, so I'm not sure if that's a default for an env var or if that's a fixed value

src/TenancyServiceProvider.php
```diff
@@ -84,10 +84,16 @@ public function boot(): void
...
-            Commands\TenantDump::class,
             Commands\MigrateFresh::class,
         ]);
 
+        if (version_compare(app()->version(), '8.0', '>=')) {
+            // DumpCommand only exists since Laravel 8.0 or above
+            $this->commands([
+                Commands\TenantDump::class,
+            ]);
+        }
```
stancl: You can probably revert all of the changes in this file since `master` is L8+

tests/CommandsTest.php
```diff
@@ -76,6 +76,11 @@
 });
 
 test('migrate command loads schema state', function () {
+    if (! isVersion8()) {
```
stancl: These changes can likewise be reverted. When exactly were they added @abrardev99?

In this commit? https://github.com/archtechx/tenancy/commit/f165fc58ba9b62b76798fdf0f9330319bceb7afd

Was that to support lower versions before we decided to send this to `4.x` (which will support fewer versions)?

You can probably just revert that one commit instead of making these changes manually.

.github/workflows/ci.yml
abrardev99: I've compared both versions. One with [cached](https://github.com/abrardev99/tenancy/runs/7947399069?check_suite_focus=true) composer dependencies and one [without caching](https://github.com/abrardev99/tenancy/runs/7947643982?check_suite_focus=true). Looks like both are taking the same time even when the cached version is found. Without cache, job finished a few (4-8) seconds early, but sometimes it finished a few seconds later. Not a major difference. 

> is it tested that this definitely speeds up the CI?

I don't think so. Plus, I found this comment from Graham
image

.github/workflows/ci.yml
abrardev99: `composer test` is not working. Error is `docker-compose: command not found`. So looks like we can't use the docker-compose command because we are not composing up in the CI.

tests/CommandsTest.php
abrardev99: > When exactly were they added @abrardev99?
> In this commit? https://github.com/archtechx/tenancy/commit/f165fc58ba9b62b76798fdf0f9330319bceb7afd

Yes. 

> Was that to support lower versions before we decided to send this to 4.x

Yes. [Here you said](https://github.com/archtechx/tenancy/pull/793#issuecomment-1034026507) support lower versions. The master branch got some changes that need `PHP 8.1`. So we [decided](https://github.com/archtechx/tenancy/pull/793#issuecomment-1205622528) to support only `L8` and `L9` on `PHP 8.1`.

Dockerfile
abrardev99: >  but does this now respect the PHP version?

As discussed, the `shivammathur/node` installed PHP versions from `5.6` to `8.1`. And there is no way to specify the version using `node`. 
If we pass the argument `PHP_VERSION=8.0 docker-compose ...` looks like it does respect and set up the extensions for the given versions. 
But when I do `docker-compose exec -T test php -v "$@"` it always gives the `8.1`. 

Also, its argument, which available only for Building image.

.github/workflows/ci.yml
stancl: Thanks for testing that. So if the cache has no real impact, you can remove it from the `ci.yml` to simplify it 👍🏻

Dockerfile
stancl: So what is the value used for exactly now? What does it affect?

.github/workflows/ci.yml
stancl: > because we are not composing up in the CI

What do you mean?

abrardev99: > What do you mean?

In CI, we are using the Dockerfile as our container, but we are not spinning up the `docker-compse.yml` services, that's why the `docker-compose` command is unavailable.

Dockerfile
abrardev99: So, I spent some time on this and tried digging it more. 

> So what is the value used for exactly now? What does it affect?

We are using Argument for enabling extensions using `phpenmod`. Note that without specifying the PHP version, extensions do not enable. So, for now, the only purpose of the variable is to enable correct extensions. 

However, if want to enable the given php version we can do that too. 
I dug into how to use the same PHP version as the argument. Following commands, switch the given php version, and later down, extensions enable for the same version. 
```
# set PHP version
RUN sudo update-alternatives --set php /usr/bin/php$PHP_VERSION \
    && sudo update-alternatives --set phar /usr/bin/phar$PHP_VERSION \
    && sudo update-alternatives --set phar.phar /usr/bin/phar.phar$PHP_VERSION \
    && sudo update-alternatives --set phpize /usr/bin/phpize$PHP_VERSION \
    && sudo update-alternatives --set php-config /usr/bin/php-config$PHP_VERSION
```

Then we can set php version `PHP_VERSION=8.0 docker-compose build --no-cache && docker-compose up -d --force-recreate`

I tried the above code. Here is the `php -v` output in the container. 
image

abrardev99: One more thing, As in this branch, we support only php `8.1`. If we run tests `./test`, I get the error. 
image

So what's the point of specifying the PHP version when we can't run the package below `8.1`? 

In fact, I was thinking of having a private variable (used only for extensions to avoid duplications) and stopping passing variables from outside.

.github/workflows/ci.yml
stancl: Ah I see, yeah

Dockerfile
stancl: I see, so:
- the variable gets used for configuring PHP extensions
- in `ARG PHP_VERSION=8.1`, the `8.1` is the default, but it **does** respect outside environment variables

Correct?

If so, that's probably good and we can keep it that way.

And I'd still accept it as an environment variable, because even though *now* we are only using PHP 8.1, in the future we'll be running the tests on a matrix that includes PHP 8.2 and any newer versions.

abrardev99: > the variable gets used for configuring PHP extensions

Yes, And we can use it for switching the PHP version. 

> in ARG PHP_VERSION=8.1, the 8.1 is the default, but it does respect outside environment variables

Yes, that's correct. 

> And I'd still accept it as an environment variable, because even though now we are only using PHP 8.1, in the future we'll be running the tests on a matrix that includes PHP 8.2 and any newer versions.

Sure. We can use the same variable for the environment as
```
ARG PHP_VERSION=8.1
ENV PHP_VERSION=${PHP_VERSION}
```

stancl: > We can use the same variable for the environment as

Not sure what you mean by that code example. Doesn't the env var get automatically passed as the value for `ARG PHP_VERSION`?

stancl: The 8.1 default should *still* be there. But we want to be able to override it using env vars.

abrardev99: > The 8.1 default should still be there. But we want to be able to override it using env vars.

So should I add the PHP version switching code now? 

> Doesn't the env var get automatically passed as the value for ARG PHP_VERSION?

I don't think so. Because when I do `ENV PHP_VERSION=8.1`, passing PHP_VERSION variable from outside is not working. So `ARG` is working only.

stancl:
@abrardev99 I addressed the open review on Basecamp. I'll merge this now and if any changes are needed you can open a new PR 👍🏻 

Thanks a lot for your work on this @abrardev99 @erikgaal!

Issue #654, opened 2021-05-14
v4 Roadmap


stancl:
## Update: Version 4 is now in early access, see the `#announcements` channel on [our Discord](https://archte.ch/discord)

Hey everyone,

Soon (in a few weeks) I'll start working on the next major version of Tenancy for Laravel.

v3 has worked extremely well, with me writing most of the code in spring 2020 and very few changes being needed for a year. The core is really solid.

That said, there's always room for improvement, which is what version 4 be focused on.

Currently, I have these changes planned:
- **Fix all of the unresolved bugs.** Most aren't fixed yet because they're low-severity in impact, and high-complexity in fixing. Some would require changes to the codebase that'd result in breaking changes, hence the major version.
- **Improve the documentation.** Both UI-wise and content-wise. I'll try to write more step-by-step guides for implementing common features and improve the structure & writing of the current pages.
- **More compatibility with the ecosystem.** Jetstream, route model binding, API auth packages, and more.

In short, the major version won't have too many changes, and most will be quality of life improvements.

But those are _my_ plans. If _you_ would like to see any new features or changes, please share them below. I'd love to hear any feature suggestions or other feedback, because I'll be working a lot on the package now, and I'll have a chance to work on features that would normally not be possible to add due to breaking changes.

Thank you!

kengraversen:
Samuel - thanks a lot for all the great work you put into this. I completely agree with your priorities. Security, stability and compatibility are for many the most important factors. New features are less important and historically a source to problems with the three mentioned areas.

dzymail:
Great news, Thank you for providing such a good package. 
looking forward to the new version of the SaaS template. I hope the template can fit the following situations out of the box:
1. In a tenant, users can belong to multiple Jetstream teams/groups and have different team roles and permissions in different teams/groups.
2. limit how many members a team can have by subscription. 
3. one user can have multiple profile types.
4. connecting Google account or Facebook/Microsoft.
5. more payment providers.

alexrififi:
It would be useful if the connection name used in trait **TenantConnection** is taken from the config

https://github.com/archtechx/tenancy/blob/54a33f93a85a5e0a4d53ac48fabacd8ebb716eec/src/Database/Concerns/TenantConnection.php#L11

stancl:
@medvinator What benefit would that bring?

alexrififi:
> @medvinator What benefit would that bring?

Convenience when working with code.
[...]

PR #795, opened 2022-02-10
[3.x] build: add illuminate support v9.x dependency

scaabel:
Added `illuminate/support: ^9.0` to `composer.json`


stancl:
See https://github.com/archtechx/tenancy/issues/654#issuecomment-1033825261

PR #796, opened 2022-02-10
Add schema dump command

viicslen:
Add schema dump command for the tenant context.

It accepts a tenant identifier as `tenant` option. If the option is not provided, it prompt the user for it or if there is no input, it will try to get the first tenant from the database.

Schema file will be created inside of the `database/schema` folder with the default name of `tenant-schema.sql`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ENV LANG=en_GB.UTF-8
 
 # install some OS packages we need
 RUN apt-get update
-RUN apt-get install -y --no-install-recommends libfreetype6-dev libjpeg62-turbo-dev libpng-dev libgmp-dev libldap2-dev netcat curl sqlite3 libsqlite3-dev libpq-dev libzip-dev unzip vim-tiny gosu git
+RUN apt-get install -y --no-install-recommends libfreetype6-dev libjpeg62-turbo-dev libpng-dev libgmp-dev libldap2-dev netcat curl sqlite3 libsqlite3-dev libpq-dev libzip-dev unzip vim-tiny gosu git default-mysql-client
     # install php extensions
 RUN docker-php-ext-configure pgsql -with-pgsql=/usr/local/pgsql \
     # && if [ "${PHP_VERSION}" = "7.4" ]; then docker-php-ext-configure gd --with-freetype --with-jpeg; else docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/; fi \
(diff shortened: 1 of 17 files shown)


viicslen:
This is just some initial concept, I will be happy to write tests and organize it a bit more if the PR gets approval

stancl:
Can you look into the failing check?

> it prompt the user for it or if there is no input, it will try to get the first tenant from the database.

And it uses just the tenant's *schema*, right? Not his data. Going for the first one is probably a safe default in that case. Tenancy makes this feature a bit more complex since different tenants can be at a different "stages" of migrations, but the first one should always be up to date.

viicslen:
> Can you look into the failing check?

Sure, I will look into it once I get home today

> And it uses just the tenant's schema, right?

Yes, just the schema along with the migrations table data so that it doens't try to run previous migrations once the schema is restored. The logic in the command is mainly for getting the tenant and initializing it. It uses Laravel's default behavior for creating the schema dump.

In theory since the migrate command uses Laravel's logic for migrating the database it should initialize the tenant's database with the tenant's schema and then run remaining migrations not included in the schema dump.

viicslen:
Added test for dumping and loading tenant schema and added conditional loading for Laravel 8 and up since the schema functionality was added in that version.

viicslen:
Sorry, I had removed the rollback command from the service provider by mistake when I tried to add conditional registering of the dump command based on larevel versions

stancl:
The imported `DumpCommand` doesn't seem to exist in Laravel 6

stancl:
I think it was added in Laravel 8 actually, so that'd make sense.

In that case, it'd probably be better to PR this to the master branch (which will be used for v4).

viicslen:
Yeah, I was afraid that might be the best solution. I did register the command conditionally if the installed laravel version is 8 and up but PHPUnit is still going to complain about it.

Alright, I will change the target to the master branch instead and remove the conditional registering code

stancl:
> I did register the command conditionally if the installed laravel version is 8 and up but PHPUnit is still going to complain about it.

Yeah, I wasn't sure what the behavior would be — if it'd only matter with actual usage of the class, or if the existence itself would break autoloading — but apparently it's the latter. So sending this to master would be best.

I'm thinking that if there are more features like this, we could have a semi-stable branch that requires Laravel 8/9, but is otherwise backwards compatible with v3. Like a step between v3 and v4. It wouldn't be guaranteed to be bug-free, but it should be stable enough for development, or staging/production if the version gets properly locked.

viicslen:
Closed in favor of #807.

PS: Sorry for messed up commit history on this one. This is actually my second PR ever to any GH project and I still haven't familiarized with the whole rebase option

stancl:
No worries, I squash-merge PRs anyway so the PR will only add one commit to the codebase 👍🏻

PR #799, opened 2022-02-11
Fix .env loading in development

erikgaal:
`Dotenv::create($paths)` was the syntax for releases before v4. As it's a dev dependency, I think it's reasonable to require the latest major version at minimum.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
         "stancl/virtualcolumn": "^1.0"
     },
     "require-dev": {
-        "vlucas/phpdotenv": "^5.0",
         "laravel/framework": "^6.0|^7.0|^8.0",
         "orchestra/testbench-browser-kit": "^4.0|^5.0|^6.0",
         "league/flysystem-aws-s3-v3": "~1.0",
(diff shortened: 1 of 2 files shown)


stancl:
> As it's a dev dependency, I think it's reasonable to require the latest major version at minimum.

Do they support older Laravel? The dev dependencies are used in the CI build with different Laravel versions

erikgaal:
Right, I see now that it's actually already a dependency of `laravel/framework`. In that case we don't have to include that dependency ourselves, right? I will update the PR to use the correct `create`/`createImmutable` method, based on the version instead so it will work on all version.

stancl:
Ah yeah, it's a dependency of laravel/framework, you can remove it.

I think I added it *before* we had full laravel/framework in the dev dependencies. But it might not be needed now

stancl:
Thanks Erik!

Is this change just a codebase cleanup, or were you contributing it because there was some issue with the package? Asking so that I know if you need this released now

erikgaal:
It's just a cleanup in preparation of a L9 compatibility PR I was planning to submit.

PR #800, opened 2022-02-12
[WIP] Laravel 9 support

CodeSkills:
Based on #654 I tried to make a compatible version with Laravel 9 with an updated Filesystem.

Only issue I have faced so far is the usage of `"facade/ignition-contracts"` since Laravel 9 uses `"spatie/laravel-ignition": "^1.0"` instead of `facade\ignition`. 

I tried to replace and update it too, but I realized that it would cause this version to be able to support only Laravel 8 & 9 with a required update to `spatie/laravel-ignition` for everyone. 

@stancl How could we solve this?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,15 +3,13 @@
     "description": "Automatic multi-tenancy for your Laravel application.",
     "keywords": ["laravel", "multi-tenancy", "multi-database", "tenancy"],
     "license": "MIT",
-    "authors": [
-        {
-            "name": "Samuel Štancl",
-            "email": "samuel.stancl@gmail.com"
-        }
-    ],
+    "authors": [{
+        "name": "Samuel Štancl",
+        "email": "samuel.stancl@gmail.com"
+    }],
     "require": {
         "ext-json": "*",
-        "illuminate/support": "^6.0|^7.0|^8.0",
+        "illuminate/support": "^6.0|^7.0|^8.0|^9.0",
         "facade/ignition-contracts": "^1.0",
         "ramsey/uuid": "^3.7|^4.0",
         "stancl/jobpipeline": "^1.0",
@@ -19,10 +17,10 @@
     },
     "require-dev": {
         "vlucas/phpdotenv": "^3.3|^4.0|^5.0",
-        "laravel/framework": "^6.0|^7.0|^8.0",
-        "orchestra/testbench-browser-kit": "^4.0|^5.0|^6.0",
-        "league/flysystem-aws-s3-v3": "~1.0",
-        "doctrine/dbal": "^2.10",
+        "laravel/framework": "^6.0|^7.0|^8.0|^9.0",
+        "orchestra/testbench-browser-kit": "^4.0|^5.0|^6.0|^7.0",
+        "league/flysystem-aws-s3-v3": "~3.0",
+        "doctrine/dbal": "^2.10|^3.1.4",
         "spatie/valuestore": "^1.2.5"
     },
     "autoload": {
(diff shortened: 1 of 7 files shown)


stancl:
Related: I believe I managed to add (composer-level) Laravel 9 support here: https://github.com/archtechx/tenancy/commit/98832195448eb781235d9e1abcfa1de4b44104c5

The dev dependencies would be ^9.0 now of course.

And we can try to change the Ignition dependencies, assuming they're backwards compatible. Not sure what would be the best way to do that specifically

CodeSkills:
I went through that commit and will try to update this PR based on it in a bit. 

I would also change the Ignition directly to `spatie/laravel-ignition` and would update the code when it is used to use that instead. Like they proposed to do here https://github.com/facade/ignition-contracts/pull/7/files , or we could help that get merged and then just use the latest version.

stancl:
If the changes in Ignition require the entire package to be installed just to use the contract, I'm fine with removing it from Tenancy.

CodeSkills:
So I have spent the last two days upgrading our codebase and all deps to L9, got it all working now. Here I updated path prefixing to currently match the original one, and it is tested and working. 

What I noticed that on L9 if I had disk root starting with `/` it would now make it double (I guess the new flysystem). But everything else works as before.

Also, I have removed the ignition, since even the contract libs will be requiring the entire package to be installed.

Let me know @stancl  if there is anything else for me to do.

stancl:
I think the Filesystem adapter changes won't work properly. I approved the CI build to see if the tests catch it

composer.json
```diff
@@ -3,26 +3,23 @@
...
     },
     "require-dev": {
         "vlucas/phpdotenv": "^3.3|^4.0|^5.0",
-        "laravel/framework": "^6.0|^7.0|^8.0",
-        "orchestra/testbench-browser-kit": "^4.0|^5.0|^6.0",
-        "league/flysystem-aws-s3-v3": "~1.0",
-        "doctrine/dbal": "^2.10",
+        "laravel/framework": "^6.0|^7.0|^8.0|^9.0",
+        "orchestra/testbench-browser-kit": "^7.0",
+        "league/flysystem-aws-s3-v3": "~3.0",
```
stancl: Why *only* 3.0? Tests run on all versions

composer.json
```diff
@@ -3,26 +3,23 @@
...
+        "doctrine/dbal": "^2.10|^3.1.4",
```
stancl: Does L9 use dbal 3.x?

CodeSkills: Not idea how did I update that here, reverted it back to 2.10. Laravel does not require it, but if it does I think it takes the latest.

composer.json
CodeSkills: Updated it so it is for 1.0 and 3.0.

CodeSkills:
You can try to approve the tests so we can see now how it goes.

CodeSkills:
Fixed the orchestra version issue

stancl:
strange that I need to re-approve the run 🤔 I thought GitHub only needs one approval per PR

erikgaal:
I would like to point out that the workflow hasn't been updated to also test on Laravel 9.

erikgaal:
Hi @CodeSkills, 

In extension to your original PR I've opened a PR myself that includes the necessary work to test on Laravel 9 and some an additional deprecation that was needed regarding maintenance mode. No hard feelings, I've added you as a co-author, but I wanted to speed up the process of releasing a version of this package compatible with L9.

CodeSkills:
Sure I don't mind, I did my best to make this fast as possible, so I could use this already in production L9 app :-)

stancl:
I'm closing this in favor of #802, thanks for the work on this 👍🏻

Issue #739, opened 2021-10-15
setting `root` parameter of S3 filesystem produces incorrect paths for some tenant-aware jobs.


plakhin:
#### Describe the bug
When I set `root` parameter of S3 filesystem, some "tenant-aware" jobs get incorrect path when queues are being processed by Laravel Horizon. It is hard to "catch" the error, however I'll try to provide all steps to reproduce. The strangest thing for me is that if I dispatch a bunch of jobs for 1 tenant - all works good. If I dispatch jobs for 2 tenants - 2nd tenant gets wrong paths. But if I dispatch jobs for 2 tenants and restart Horizon after all jobs for 1st tenant were processed but before jobs for 2nd tenant start to process - all works fine again for 2nd tenant. If I don't set `root` - all works fine for all tenants.

#### Steps to reproduce
filesystems.php:
```
        's3' => [
            'driver'                  => 's3',
            'key'                     => env('AWS_ACCESS_KEY_ID'),
            'secret'                  => env('AWS_SECRET_ACCESS_KEY'),
            'region'                  => env('AWS_DEFAULT_REGION'),
            'bucket'                  => env('AWS_BUCKET'),
            'url'                     => env('AWS_URL'),
            'endpoint'                => env('AWS_ENDPOINT'),
            'use_path_style_endpoint' => env('AWS_USE_PATH_STYLE_ENDPOINT', false),
            'visibility'              => 'public',
            'root'                    => 'media',
```
1. Create a simple queued job that will put a file to the `s3` disk: `Storage::cloud()->put('hello.json', '{"hello": "world"}');`
2. Dispatch that job for 2-3 tenants
3. Start horizon and wait till all jobs processed.
#### Expected behavior
There should be 3 files in S3 storage bucket:
```
/media/tenant1/hello.json
/media/tenant2/hello.json
/media/tenant3/hello.json
```

#### Actual behavior
There will be an error if using MinIO because the path containing one extra `/` symbol for Tenant2 (for Tenant 1 the path is correct).
The path for `put` command should be `/media/tenant2/hello.json` but the actual path is `/media//tenant2/hello.json`

There will be 3 files in S3 storage bucket (in case of Digital Ocean Spaces):
```
/media/tenant1/hello.json
/tenant2/hello.json
/tenant3/hello.json
```

#### Your setup
 - Laravel version: [e.g. 8.64.0]
 - stancl/tenancy version: [e.g. 3.4.4]

stancl:
[...]

PR #802, opened 2022-02-14
[3.x] Compatibility with Laravel 9

erikgaal:
There were some more changes required before getting the package to work with Laravel 9. I've added testing on L9 to the workflow so we can be sure it's compatible before we merge.

Thank you @CodeSkills for the work in your PR. I've made sure to add you as a co-author for the commit regarding the `FilesystemTenancyBootstrapper`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,16 +11,16 @@
     ],
     "require": {
         "ext-json": "*",
-        "illuminate/support": "^6.0|^7.0|^8.0",
+        "illuminate/support": "^6.0|^7.0|^8.0|^9.0",
         "facade/ignition-contracts": "^1.0",
         "ramsey/uuid": "^3.7|^4.0",
         "stancl/jobpipeline": "^1.0",
         "stancl/virtualcolumn": "^1.0"
     },
     "require-dev": {
-        "laravel/framework": "^6.0|^7.0|^8.0",
-        "orchestra/testbench-browser-kit": "^4.0|^5.0|^6.0",
-        "league/flysystem-aws-s3-v3": "~1.0",
+        "laravel/framework": "^6.0|^7.0|^8.0|^9.0",
+        "orchestra/testbench": "^4.0|^5.0|^6.0|^7.0",
+        "league/flysystem-aws-s3-v3": "^1.0|^3.0",
         "doctrine/dbal": "^2.10",
         "spatie/valuestore": "^1.2.5"
     },
--- a/src/Bootstrappers/FilesystemTenancyBootstrapper.php
+++ b/src/Bootstrappers/FilesystemTenancyBootstrapper.php
@@ -54,23 +54,22 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
         }
 
         // Storage facade
-        foreach ($this->app['config']['tenancy.filesystem.disks'] as $disk) {
-            /** @var FilesystemAdapter $filesystemDisk */
-            $filesystemDisk = Storage::disk($disk);
-            $this->originalPaths['disks'][$disk] = $filesystemDisk->getAdapter()->getPathPrefix();
+        Storage::forgetDisk($this->app['config']['tenancy.filesystem.disks']);
 
-            if ($root = str_replace(
+        foreach ($this->app['config']['tenancy.filesystem.disks'] as $disk) {
+            $root = $this->app['config']->get("filesystems.disks.{$disk}.root");
+            $this->originalPaths['disks'][$disk] = $root;
+
+            if (!($root = str_replace(
                 '%storage_path%',
                 storage_path(),
                 $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? ''
-            )) {
-                $filesystemDisk->getAdapter()->setPathPrefix($finalPrefix = $root);
-            } else {
-                $root = $this->app['config']["filesystems.disks.{$disk}.root"];
-                $filesystemDisk->getAdapter()->setPathPrefix($finalPrefix = $root . "/{$suffix}");
+            ))) {
+                $root = "{$root}/{$suffix}";
             }
 
-            $this->app['config']["filesystems.disks.{$disk}.root"] = $finalPrefix;
+            $this->app['config']["filesystems.disks.{$disk}.root"] = $root;
+
         }
     }
 
@@ -84,14 +83,9 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
         $this->app['url']->setAssetRoot($this->app['config']['app.asset_url']);
 
         // Storage facade
+        Storage::forgetDisk($this->app['config']['tenancy.filesystem.disks']);
         foreach ($this->app['config']['tenancy.filesystem.disks'] as $disk) {
-            /** @var FilesystemAdapter $filesystemDisk */
-            $filesystemDisk = Storage::disk($disk);
-
-            $root = $this->originalPaths['disks'][$disk];
-
-            $filesystemDisk->getAdapter()->setPathPrefix($root);
-            $this->app['config']["filesystems.disks.{$disk}.root"] = $root;
+            $this->app['config']["filesystems.disks.{$disk}.root"] = $this->originalPaths['disks'][$disk];
         }
     }
 }
--- a/src/Bootstrappers/FilesystemTenancyBootstrapper.php
+++ b/src/Bootstrappers/FilesystemTenancyBootstrapper.php
@@ -57,19 +57,20 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
         Storage::forgetDisk($this->app['config']['tenancy.filesystem.disks']);
 
         foreach ($this->app['config']['tenancy.filesystem.disks'] as $disk) {
-            $root = $this->app['config']->get("filesystems.disks.{$disk}.root");
-            $this->originalPaths['disks'][$disk] = $root;
+            $originalRoot = $this->app['config']["filesystems.disks.{$disk}.root"];
+            $this->originalPaths['disks'][$disk] = $originalRoot;
 
-            if (!($root = str_replace(
+            $finalPrefix = str_replace(
                 '%storage_path%',
                 storage_path(),
-                $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? ''
-            ))) {
-                $root = "{$root}/{$suffix}";
+                $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? '',
+            );
+
+            if (! $finalPrefix) {
+                $finalPrefix = $originalRoot . '/'. $suffix;
             }
 
-            $this->app['config']["filesystems.disks.{$disk}.root"] = $root;
-
+            $this->app['config']["filesystems.disks.{$disk}.root"] = $finalPrefix;
         }
     }
 
--- a/src/Middleware/CheckTenantForMaintenanceMode.php
+++ b/src/Middleware/CheckTenantForMaintenanceMode.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Stancl\Tenancy\Middleware;
 
 use Closure;
-use Illuminate\Foundation\Http\Exceptions\MaintenanceModeException;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode;
 use Stancl\Tenancy\Exceptions\TenancyNotInitializedException;
 use Symfony\Component\HttpFoundation\IpUtils;
@@ -29,7 +29,12 @@ class CheckTenantForMaintenanceMode extends CheckForMaintenanceMode
                 return $next($request);
             }
 
-            throw new MaintenanceModeException($data['time'], $data['retry'], $data['message']);
+            throw new HttpException(
+                503,
+                'Service Unavailable',
+                null,
+                isset($data['retry']) ? ['Retry-After' => $data['retry']] : []
+            );
         }
 
         return $next($request);
--- a/tests/BootstrapperTest.php
+++ b/tests/BootstrapperTest.php
@@ -194,7 +194,7 @@ class BootstrapperTest extends TestCase
         // Check that disk prefixes respect the root_override logic
         $this->assertSame($expected_storage_path . '/app/', $this->getDiskPrefix('local'));
         $this->assertSame($expected_storage_path . '/app/public/', $this->getDiskPrefix('public'));
-        $this->assertSame('/tenant' . tenant('id') . '/', $this->getDiskPrefix('s3'));
+        $this->assertSame('tenant' . tenant('id') . '/', ltrim($this->getDiskPrefix('s3'), '/'));
 
         // Check suffixing logic
         $new_storage_path = storage_path();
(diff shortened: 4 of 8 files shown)


stancl:
Thanks for the PRs! I'm happy to merge #800 or #802, whichever adds the L9 support fully

With regards to the workflow changes you made, I think it'll fail now since Laravel 9 requires PHP 8. And it's trying to use it with PHP 7.4.

It'd probably be best to merge #793 first. There are a few changes remaining there, so if you could try implementing them, we'll be able to see how this PR works.

I just responded to a few reviews there.

composer.json
```diff
@@ -11,16 +11,16 @@
     ],
     "require": {
         "ext-json": "*",
-        "illuminate/support": "^6.0|^7.0|^8.0",
+        "illuminate/support": "^6.0|^7.0|^8.0|^9.0",
         "facade/ignition-contracts": "^1.0",
```
stancl, line 15: Won't this have the same issue that @codeskills tried to address in his PR?

erikgaal: Yes, but I think we should keep the contracts so the solutions work for older versions. I don't think there is a way to support solutions in libraries at the moment without including `ignition` in the project.

stancl: If `facade/ignition-contracts` doesn't cause issues on L9, I agree that this makes sense. That seems to be the case, looking at the CI build.

If Spatie ends up making `spatie/ignition-contracts` **not** depend on entire ignition, we can try to see if there's a way to support both <L9 and L9

I'll keep this review open

src/Bootstrappers/FilesystemTenancyBootstrapper.php
```diff
@@ -54,23 +54,22 @@ public function bootstrap(Tenant $tenant)
...
+
         foreach ($this->app['config']['tenancy.filesystem.disks'] as $disk) {
-            /** @var FilesystemAdapter $filesystemDisk */
-            $filesystemDisk = Storage::disk($disk);
-            $this->originalPaths['disks'][$disk] = $filesystemDisk->getAdapter()->getPathPrefix();
+            $root = $this->app['config']->get("filesystems.disks.{$disk}.root");
+            $this->originalPaths['disks'][$disk] = $root;
 
-            if ($root = str_replace(
+            if (!($root = str_replace(
```
stancl: @codeskills could you explain the changes a bit?

tests/BootstrapperTest.php
```diff
@@ -187,27 +187,6 @@ public function filesystem_data_is_separated()
...
-            } else {
-                if ($base = $old_storage_facade_roots[$disk]) {
-                    $correct_path_prefix = $base . "/$suffix/";
-                } else {
-                    $correct_path_prefix = "$suffix/";
-                }
-            }
-
-            $this->assertSame($correct_path_prefix, $current_path_prefix);
-        }
```
stancl: Why was this removed?

src/Middleware/CheckTenantForMaintenanceMode.php
```diff
@@ -29,7 +29,12 @@ public function handle($request, Closure $next)
                 return $next($request);
             }
 
-            throw new MaintenanceModeException($data['time'], $data['retry'], $data['message']);
+            throw new HttpException(
+                503,
+                'Service Unavailable',
+                null,
+                isset($data['retry']) ? ['Retry-After' => $data['retry']] : []
+            );
```
stancl, line 37: Was `MaintenanceModeException` removed?

erikgaal: It's deprecated since L8, and with Symfony 6 using it it throws a TypeError.

src/Bootstrappers/FilesystemTenancyBootstrapper.php
```diff
@@ -84,14 +83,9 @@ public function revert()
         $this->app['url']->setAssetRoot($this->app['config']['app.asset_url']);
 
         // Storage facade
+        Storage::forgetDisk($this->app['config']['tenancy.filesystem.disks']);
```
stancl, line 87: Is `forgetDisk()` used to solve the prefix issue without having to manipulate the existing disk instances?

src/Bootstrappers/FilesystemTenancyBootstrapper.php
stancl: I'd like to make sure this uses the root overrides correctly, as described here https://tenancyforlaravel.com/docs/v2/filesystem-tenancy/

I think I might add some assertions to the `BootstrapperTest`, assuming the PR can be set up locally

erikgaal: I think this was an IDE suggestion to negate the condition and remove the original `if` part. Because the original `if` part of the conditional (the `setPathPrefix` call) was no longer needed.

src/Bootstrappers/FilesystemTenancyBootstrapper.php
erikgaal: Correct, it makes sure that the next time the disk is accessed, it is recreated from the new config.

tests/BootstrapperTest.php
erikgaal: I feel like this part is unnecessarily testing the internals of the `FilesystemBootstrapper`. On line 189 the storage path is already tested by checking the result of the `storage_path` function. The test also broke down because we can't do `getPathPrefix` anymore. Testing the end-user results (`storage_path()` and `Storage::facade()`) makes more sense to me.

src/Bootstrappers/FilesystemTenancyBootstrapper.php
CodeSkills: It is just a simplification, it works the same way as before, since there is `$this->app['config']["filesystems.disks.{$disk}.root"] = $root;` at the end of the function. You dont have to do this `$filesystemDisk->getAdapter()->setPathPrefix($finalPrefix = $root);` because it changes the config itself.

stancl: Pushed those test changes now

src/Middleware/CheckTenantForMaintenanceMode.php
stancl: And the code above is how Laravel implements maintenance mode in L8+?

I'd like to avoid making a breaking change here if people depend on the old exception in some way. But if the new implementation is similar to how Laravel does this, it's probably fine.

tests/BootstrapperTest.php
stancl: I believe the point of this was to test that the actual adapter instances use the right paths.

`storage_path()` is irrelevant in that context, and `Storage::facade()` doesn't allow fetching the prefix. I pushed some tests now to assert that the paths used by the instances in `Storage` are correct

src/Middleware/CheckTenantForMaintenanceMode.php
erikgaal: Yeah, I copied this from Laravel's codebase :)

stancl: Cool, we can go with this then

erikgaal:
Looks good to me now :)

tests/BootstrapperTest.php
erikgaal: Right, makes sense and should be fine this way now. I do believe a better abstraction of resolving the paths (and testing those) would result in a robuster implementation. Maybe something for v4.

stancl:
Do you think we can finish 793 as I wrote here? https://github.com/archtechx/tenancy/pull/802#issuecomment-1039104506

src/Bootstrappers/FilesystemTenancyBootstrapper.php
stancl: I'd rewrite the if to not include the negation and assignment in the condition. You can make them separate

colinmackinlay:
> Do you think we can finish 793 as I wrote here? [#802 (comment)](https://github.com/archtechx/tenancy/pull/802#issuecomment-1039104506)

@stancl - would be great if this could be merged now if it is working. Even if you don't tag a release until #793 is done we could still pull it in via `dev:master` and run Laravel 9 upgrades :)

tests/BootstrapperTest.php
```diff
@@ -184,30 +189,38 @@ public function filesystem_data_is_separated()
         $this->assertFalse(Storage::disk('public')->exists('foo'));
         $this->assertFalse(Storage::disk('public')->exists('abc'));
 
+        $expected_storage_path = $old_storage_path . '/tenant' . tenant('id'); // /tenant = suffix base
+
+        // Check that disk prefixes respect the root_override logic
+        $this->assertSame($expected_storage_path . '/app/', $this->getDiskPrefix('local'));
+        $this->assertSame($expected_storage_path . '/app/public/', $this->getDiskPrefix('public'));
+        $this->assertSame('tenant' . tenant('id') . '/', ltrim($this->getDiskPrefix('s3'), '/'));
```
stancl, line 197: I'm using `ltrim()` in the tests here, to make the strings identical, but per what @CodeSkills suggested in https://github.com/archtechx/tenancy/pull/800#issuecomment-1038433316 there might be a problem with the prefixing? So if this issue happens in the app and not just tests, the `/` prefix adjustment should be made in the bootstrapper rather than tests.

stancl: (For context, the tests don't use s3 at all. They just make some general config assertions.)

composer.json
stancl: Seems like the `ignition-contracts` dependency also makes the package incompatible with PHP 8.1? https://github.com/facade/ignition-contracts/blob/3c921a1cdba35b68a7f0ccffc6dffc1995b18267/composer.json#L20

stancl:
@colinmackinlay I can merge the PR sooner and tag a version after all of the other things are resolved, but there are still some reviews here that should be addressed before merging.

composer.json
arcdigital: they have it set to `^8.0`, which will allow PHP 8.1 to work fine

stancl: Ah, good catch. I think we'll stay with the old `ignition-contracts` package then, for now

SupianIDz:
> @colinmackinlay I can merge the PR sooner and tag a version after all of the other things are resolved, but there are still some reviews here that should be addressed before merging.

is there any progress ? We have entered the tenancy stage of our program. I hope it's merged soon.

Thank you all for your hard work ❤️ .

magarrent:
Hey! It's the package already compatible with L9? Or there are some pending issues? Maybe I can help

stancl:
Would appreciate @CodeSkills' feedback on the review where I mentioned him, and a refactor of this https://github.com/archtechx/tenancy/pull/802/files/f5269b2b025e24243cba78e06435f25bfff9b3fa#diff-a7131a9582938c1b642f3d517a9064cbacb7b83d75fb52fe2c6e52dbca26a584

src/Bootstrappers/FilesystemTenancyBootstrapper.php
CodeSkills: You can rewrite it to something like this, to make it more clear:
```php
$root = str_replace(
    '%storage_path%',
    storage_path(),
    $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? ''
);
// Default root
if (!$root) {
    $root = "{$root}/{$suffix}";
}
```

CodeSkills:
I have written there an update on how you could rewrite it. Hope it helps! (Someone should probably test it before since it has been changed from my PR, when somebody replaced the vars with {$root}, which I did not test.

tests/BootstrapperTest.php
stancl: @CodeSkills could you comment on the `/` prefix changes? It'd help a lot here

oxiginedev:
hey! @stancl , please what's the update on support for laravel 9. Any help needed?

src/Bootstrappers/FilesystemTenancyBootstrapper.php
erikgaal: I've refactored the proposed changes to prevent overwriting used variables. It took me a while to fully understand what the code was doing, but this should clear it up.

src/Bootstrappers/FilesystemTenancyBootstrapper.php
```diff
@@ -54,20 +54,20 @@ public function bootstrap(Tenant $tenant)
...
                 storage_path(),
-                $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? ''
-            )) {
-                $filesystemDisk->getAdapter()->setPathPrefix($finalPrefix = $root);
-            } else {
-                $root = $this->app['config']["filesystems.disks.{$disk}.root"];
-                $filesystemDisk->getAdapter()->setPathPrefix($finalPrefix = $root . "/{$suffix}");
+                $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? '',
+            );
+
```
vstyler96, line 68:
```suggestion

$finalPrefix = str_replace('%tenant%', $suffix, $finalPrefix);

```
I would like to suggest the following suggestion if this is accepted. let add a custom path with the tenant as argument.

stancl: If this is a new feature, like in the other PR, it's best to add it after merging this.

This PR should be just for the Laravel upgrade

stancl:
I'll probably finish and merge this later today 👍🏻

stancl:
Thanks for the $finalPrefix refactor (as well as your work on all this) @erikgaal! It looks great now.

The only thing I'm unsure about is the review re: S3, but I'll merge this into 3.x anyway. I'd appreciate if people who use S3 could test that branch and report if there are any issues. If there are not, I'll tag a release tomorrow. I'll also try to revisit the logic tomorrow (in code, not prod) to see if it's okay.

So to use the dev branch locally, you can just run:
```sh
composer require stancl/tenancy:3.x-dev
```

And then switch it back to `^3.0` once this is released.

tkuhnert:
I just updated locally to L9 using 3.x-dev. Using flysystem-aws-s3-v3 in combination with a minio server. Everything works as expected so far. 👍

stancl:
Thanks a lot for the testing!

jtomek:
Hi, I also tested it on Vapor. All worked fine except the filesystem root url. I have described the issue and a hotfix here: https://github.com/archtechx/tenancy/issues/739#issuecomment-1067763548

stancl:
@jtomek Can you confirm that this regression was introduced in the latest release (and therefore it all worked fine in 3.5.1)?

jtomek:
@stancl, I will try my best to find a slot for it. This week is rather busy. I can share that I have been using and updating your package regularly for a long time and the issue at hand started with the Laravel 9 update.

miketimeturner:
> @jtomek Can you confirm that this regression was introduced in the latest release (and therefore it all worked fine in 3.5.1)?

I can confirm this is the case. I have the same issue and therefor tenant directories now reside in directory "/" this is also having an impact retrieving the media and you are met with a permission denied, probably due to an incorrect url.

The issue no longer exist after reverting to 3.5.1

stancl:
Thanks for the testing! I'll try to get it fixed this week.

trippo:
Any news on this?
Are now production-ready v.3.5.2?

stancl:
See the release notes

trippo:
Thanks seem that v3.5.5 solves all bugs.

Thanks

PR #807, opened 2022-02-17
[4.x] Add tenant schema dump command

viicslen:
Add schema dump command for the tenant context.

It accepts a tenant identifier as tenant option. If the option is not provided, it prompt the user for it or if there is no input, it will try to get the first tenant from the database.

Schema file will be created inside of the database/schema folder with the default name of tenant-schema.sql

### DIFF
--- a/src/Commands/TenantDump.php
+++ b/src/Commands/TenantDump.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Database\ConnectionResolverInterface;
+use Illuminate\Database\Console\DumpCommand;
+use Stancl\Tenancy\Contracts\Tenant;
+use Symfony\Component\Console\Input\InputOption;
+
+class TenantDump extends DumpCommand
+{
+    /**
+     * Create a new command instance.
+     */
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->setName('tenants:dump');
+        $this->specifyParameters();
+    }
+
+    /**
+     * Execute the console command.
+     *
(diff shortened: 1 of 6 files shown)


src/Commands/TenantDump.php
```diff
@@ -0,0 +1,77 @@
...
+     * @return \Stancl\Tenancy\Contracts\Tenant
+     *
+     * @throws \Throwable
+     */
+    public function tenant(): Tenant
+    {
+        $tenant = $this->option('tenant')
+            ?? tenant()
+            ?? tenancy()->query()->first()
+            ?? $this->ask('What tenant do you want to dump the schema for?');
```
stancl: If `first()` will return null, the `ask()` can't work either. I think we could ask this question, with the default answer being the first tenant.

tests/Etc/tenant-schema.dump
```diff
@@ -0,0 +1,66 @@
+/*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
```
stancl, line 1: Can we gitignore this file? Or remove it in the test or tearDown()

viicslen: This file is used in the `migrate_command_loads_schema_state` test to check if it actually used the schema to migrate the tenants.  I contains a _schema_users_ table that is not included in the default test migrations.

The actual schema dump generated during the test will have the name of "tenant-schema-test.dump". I will gitignore it so that someone cannot commit it by mistake during local testing

src/Commands/TenantDump.php
viicslen: True 😅. Alright I will change it as soon as I can.

stancl:
Thanks for the PR!

therecluse26:
@stancl is there any documentation on how to actually use the schema dump file in a tenant context? I'm not seeing it referenced anywhere on your site and am struggling to get it working

stancl:
v4 isn't released yet, see `#announcements` on [our Discord](https://archte.ch/discord).

PR #812, opened 2022-02-22
Add drop of db views on migrate fresh command

### LINKED ISSUE #811: db:wipe does not wipe views
#### Description
It would be nice to be able to wipe views too of a tenant db, along with the real tables.
Right now the package takes care only of tables.

It should be a trivial job, since it can be accomplished passing a parameter (`'--drop-views' => true`) to db:wipe on 

`/stancl/tenancy/src/Commands/MigrateFresh.php`
#### Why this should be added
I think it can help in cases when you work with views, and it has no side effects in case no views are not present.

I can create a pull request once @stancl 's feedback is positive about that.

masiorama:
#811

### DIFF
--- a/src/Commands/MigrateFresh.php
+++ b/src/Commands/MigrateFresh.php
@@ -36,13 +36,11 @@ final class MigrateFresh extends Command
      */
     public function handle()
     {
-        $dropViewsFlag = $this->option('drop-views');
-
-        tenancy()->runForMultiple($this->option('tenants'), function ($tenant) use ($dropViewsFlag) {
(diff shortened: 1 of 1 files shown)


masiorama (commented):
Right, missed that one.

stancl:
Thanks! Will include this in the next release

masiorama:
> Thanks! Will include this in the next release

I just focused a bit more on your sentence: do you mean in the next minor of v3 or in v4? Thanks!

stancl:
v3 👍🏻

PR #817, opened 2022-03-04
[3.x][Filesystem] Provide an additional argument for tenant name path

vstyler96:
## Summary of changes
During the work of an application that uses Google Cloud Storage as the main storage method. I found that if we want to save a file without overriding the default root path for GCS, the `FilesystemTenancyBootstrapper` adds an additional folder in the GCS Filesystem called `/`.

image

So I decided to pass a custom folder path.

### Case 1: using %storage_path%
If we try to use the param %storage_path% this will add the full path to our project into the GCS filesystem.
image

### Case 2: using a custom path without %storage_path%
I tried to pass a custom path, but this does not include the tenant folder. so the files from all tenants are mixed with each other (not expected)
image

So my resolution includes a minor update to the `$finalPrefix` variable. If we add a %tenant% variable, this will let us pass a more customizable path. So we can pass the `/a/random/path/for/my/tenant/%tenant%` path or any other path manually for each tenant in filesystems like GCS or S3

### DIFF
--- a/src/Bootstrappers/FilesystemTenancyBootstrapper.php
+++ b/src/Bootstrappers/FilesystemTenancyBootstrapper.php
@@ -65,7 +65,7 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
                 storage_path(),
                 $subject
             )) {
-                $root = str_replace('%tenant%', $suffix, $subject);
+                $root = str_replace('%tenant%', $suffix, $root);
                 $filesystemDisk->getAdapter()->setPathPrefix($finalPrefix = $root);
             } else {
                 $root = $this->app['config']["filesystems.disks.{$disk}.root"];
--- a/src/Bootstrappers/FilesystemTenancyBootstrapper.php
+++ b/src/Bootstrappers/FilesystemTenancyBootstrapper.php
@@ -58,12 +58,14 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
             /** @var FilesystemAdapter $filesystemDisk */
             $filesystemDisk = Storage::disk($disk);
             $this->originalPaths['disks'][$disk] = $filesystemDisk->getAdapter()->getPathPrefix();
+            $subject = $this->app['config']["tenancy.filesystem.root_override.{$disk}"] ?? '';
 
             if ($root = str_replace(
                 '%storage_path%',
                 storage_path(),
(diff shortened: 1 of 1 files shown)


stein-j:
Isn't the tenant folder created automatically ? So you simply need to do:
```php
'root_override' => [
            // ...
            'gcs' => 'tenants/',
],
```

Unless if I'm missing something

vstyler96:
> Isn't the tenant folder created automatically ? So you simply need to do:
> 
> ```
> 'root_override' => [
>             // ...
>             'gcs' => 'tenants/',
> ],
> ```
> 
> Unless if I'm missing something

I already made it and all of the files from all tenants are stored in `tenants/`
image

stancl:
Yeah I think @stein-j's suggestion should work? Not exactly sure what the problem is here. S3 doesn't use the `storage_path` *at all* and it stores files like `/tenant123/foo.txt`

vstyler96:
> Yeah I think @stein-j's suggestion should work? Not exactly sure what the problem is here. S3 doesn't use the `storage_path` *at all* and it stores files like `/tenant123/foo.txt`

Using GCS the path adds a "/" folder as root

stancl:
> Using GCS the path adds a "/" folder as root

What do you mean by that?

vstyler96:
> > Using GCS the path adds a "/" folder as root
> 
> What do you mean by that?

image
image

stancl:
I still don't understand. If you want this feature added, please explain clearly what the issue is and how this would solve it.

vstyler96:
@stancl I just updated the description :D. I tried to be clearer.

Also as mentioned in https://github.com/archtechx/tenancy/pull/802#discussion_r820821748, I'll wait for that merge and update this PR after that.

stancl:
Note that I just pushed some changes to this logic

jncalderon:
@stancl @vstyler96 some any news?

stancl:
@vstyler96 Could you resolve the conflicts?

vstyler96:
@stancl Ready to go! 🤝 
image
image

P.S: Also removed a no longer used class on `use` declaration section

jncalderon:
@stancl is ready :)

stancl:
Thanks!

danpalmieri:
i'm getting mkdir(): File exists error on [app / Jobs / CreateFrameworkDirectoriesForTenant.php : 21]
trying to create a new tenant

PR #836, opened 2022-04-01
Laravel 9 support for JobRetryRequested listener

bonroyage:
The `JobRetryRequested` event is also available on Laravel 9. The current code however, only works for Laravel 8 since it checks that the version starts with "8". This PR switches the check to PHP's native `version_compare` function to check that the Laravel version is at least 8.64 (https://github.com/laravel/framework/releases/tag/v8.64.0), since that is where this event was added.


stancl:
Great PR, thanks a lot!

stancl:
Great PR, thanks a lot!

PR #840, opened 2022-04-07
[3.x] Update PostgreSQLSchemaManager to set correct config key value

nthndnn:
In Laravel 9.0, the Postgres `schema` config key has been replaced by `search_path` (https://laravel.com/docs/9.x/upgrade#postgres-schema-configuration).

This PR updates the `PostgreSQLSchemaManager` to check which config value is in use and sets it accordingly. It might be worth setting both no matter whether the key exists or not, but I thought this would be the best solution.


nthndnn:
I've updated this to use `version_compare` as this has been done elsewhere and looks like a more readable solution 😄

stancl:
The test could probably use `version_compare()` as well

nthndnn:
@stancl cool, wasn't sure whether `app()->version()` was available in the tests

stancl:
Looks great now, thanks a lot for the PR!

PR #848, opened 2022-04-19
Fix eloquent relational query

ashraful19:
Fixing error "Call to a member function where() on array " error

### DIFF
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -55,7 +55,7 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
 
     protected function setCurrentDomain(Tenant $tenant, string $domain): void
     {
-        static::$currentDomain = $tenant->domains->where('domain', $domain)->first();
+        static::$currentDomain = $tenant->domains()->where('domain', $domain)->first();
     }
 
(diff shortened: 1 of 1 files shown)


stancl:
Could you share what needs to happen for this bug to occur?

ashraful19:
Hi @stancl 
When a `InitializeTenancyByDomain` or `InitializeTenancyBySubdomain` or `InitializeTenancyByDomainOrSubdomain` middleware is used, and the tenant is found, this error is shown as the resolver cant set the current domain, because `$tenant->domains` is returning an `array` but it should return a `Illuminate\Database\Eloquent\Relations\HasMany` object to chain further `where()` method

stancl:
I don't think this fix should be needed, there weren't any issues with this in the past. And here it'd be unnecessarily bypassing the loaded relation data.

Are you using some custom setup? I'm not sure why this would be happening to you but not others

ashraful19:
It was strange to me too that why this didnt happened to others. I didnt have any custom setup, only installed the package in my existing application. 
the problem is `$tenant->domains` isnt returning a relation/collection but a plain array, and thus the further chained `where()` gives error. 
At the moment i ended up making a custom middleware and custom resolver. 
I will check more in deep why i am getting a plain array with `$tenant->domains` and not anyone else.

Will update you.

PR #855, opened 2022-05-02
Fix composer dump autoload issue on fresh installation

suleodu:
There was a commer after the last parameter of str_replace() on line 63 this affect composer autoload to complete after instalation

Thanks.


stancl:
This has no effect on composer.

Issue #717, opened 2021-09-13
Creating a new model on a tenant with resource syncing can cause a duplicate entry error


chillbram:
I'm using resource syncing on my User model and when creating a new User from a tenant it will sometimes throw the following error `SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2' for key 'central_users.PRIMARY' (SQL: insert into 'central_users' ('name', 'email', 'password', 'global_id', 'id') values (test, test@test.com, asdasdasd, ee2bab73-708e-486b-b715-91909b0c0b5c, 2))`

Because I use a regular incrementing id and not the global_id as the primary key for the User and CentralUser databases, Tenancy also tries to sync the primary key when creating a new Central instance from a new tenant model. 

The following code causes the problem, because it tries to sync all model attributes regardless of the columns put into `User::getSyncedAttributeNames()`: https://github.com/archtechx/tenancy/blob/3.x/src/Listeners/UpdateSyncedResource.php#L61-L63.

#### Your setup
 - Laravel version: 8.59
 - stancl/tenancy version: 3.4.4

stancl:
Isn't there an issue already open for this?

chillbram:
I couldn't find an issue or PR. There have been multiple reports of this bug on the Discord though, but none with a fix/solution

stancl:
Maybe #658 is similar? Just want to keep things tidy for when I get to working on the v4

chillbram:
Ah yes, different error but the same issue. The current code forces the SyncMaster to have all columns that the Syncable model has as well, which causes a problem when using tenant-specific columns as well as when using a non-uuid primary key. Feel free to close this issue if you want to keep everything in #658

stancl:
Cool, I'll see that this was mentioned so I'll read both issues when fixing this. Thanks!

PR #865, opened 2022-05-27
Fixed Synced resources between tenants throws error when inserting new model on tenant

amjeed-ay:
Fixed Synced resources between tenants throws error when inserting new model on tenant.
**before:**
when creating new model on tenant there was an error thrown 
`SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2' for key 'central_users.PRIMARY' (SQL: insert into 'central_users' ('name', 'email', 'password', 'global_id', 'id') values (test, test@test.com, asdasdasd, ee2bab73-708e-486b-b715-91909b0c0b5c, 2))`  #717 

 `updateResourceInCentralDatabaseAndGetTenants` is try to create new model with the same attributes including `id` which may be exist in the central database since it is incrementing id.

**now:**
  the `id` is excluded so it can create the model with all attributes except the id which can be ignored.

### DIFF
--- a/src/Listeners/UpdateSyncedResource.php
+++ b/src/Listeners/UpdateSyncedResource.php
@@ -60,7 +60,7 @@ class UpdateSyncedResource extends QueueableListener
             } else {
                 // If the resource doesn't exist at all in the central DB,we create
                 // the record with all attributes, not just the synced ones.
-                $centralModel = $event->model->getCentralModelName()::create($event->model->getAttributes());
+                $centralModel = $event->model->getCentralModelName()::create(collect($event->model->getAttributes())->except(['id'])->toArray());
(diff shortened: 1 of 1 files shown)


amjeed-ay:
@stancl I think  found a better solution for this issue.

Instead of altering `updateResourceInCentralDatabaseAndGetTenants` which later I realize that there is issue again where authenticated user can switch to another tenant without logged in if user exist with same `id`

In below method I increment `id` manually when creating new user, by finding the last user id from `CentralUser` Model and increment it, therefore no users will exist with same id from all the tenant.
```
User::create([
    'id' => CentralUser::all()->last()->id + 1;   //find the last id and increment
    'name' => 'John Doe',
    'email' => 'johndoe@email.com',
    'password' => Hash::make('password'),
  ]);

```

PR #869, opened 2022-06-01
[4.x] Add pending tenants (modified #782)

stancl:
Resolved conflicts in #782 

### Original description

As discussed here #779, this PR adds a new feature to this package: **pending tenants**.

This adds:
- Four new config settings: `pending.include_in_scope`, `pending.count`, `pending.older_than_days` and `pending.older_than_hours`
- Two new commands (`tenants:pending` and `tenants:pending-clear`)
- A `hasPending` trait that contains a pendingScope which was inspired by the `SoftDelete` trait.

All of this can be combined with the scheduler to automatically have x amount of tenants ready to be used and link to a new client/domain and skip the time of creating and migrating (and seeding) the client database.


stein-j:
Hello @stancl, I see you're moving things around. If you need me to do something, I'm here 👍🏼

src/Commands/CreatePendingTenants.php
```diff
@@ -0,0 +1,64 @@
...
+use Illuminate\Console\Command;
+
+class CreatePendingTenants extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'tenants:pending {--count= : The number of tenant to be in a pending state}';
```
lukinovec:
```suggestion
    protected $signature = 'tenants:pending {--count= : The number of tenants to be in a pending state}';
```

src/Commands/CreatePendingTenants.php
```diff
@@ -0,0 +1,64 @@
...
+        $pendingObjectifCount = (int) ($this->option('count') ?? config('tenancy.pending.count'));
+
+        $pendingCurrentCount = $this->getPendingTenantCount();
+
+        $deployedCount = 0;
+        while ($pendingCurrentCount < $pendingObjectifCount) {
+            tenancy()->model()::createPending();
+            // We update the number of pending tenants every time with a query to get a live count.
+            // this prevents to deploy too many tenants if pending tenants are being created simultaneous somewhere else
+            // during the runtime of this command.
```
lukinovec:
```suggestion
            // Update the number of pending tenants every time with a query to get a live count
            // To prevent deploying too many tenants if pending tenants are being created simultaneously somewhere else
            // While running this command
```

style + minor wording changes

src/Commands/CreatePendingTenants.php
```diff
@@ -0,0 +1,64 @@
...
+            $pendingCurrentCount = $this->getPendingTenantCount();
+            $deployedCount++;
+        }
+
+        $this->info("$deployedCount tenants deployed, $pendingObjectifCount tenant(s) are ready to be used.");
```
lukinovec:
```suggestion
        $this->info($deployedCount . ' ' . str('tenant')->plural($deployedCount) . ' deployed.');
        $this->info($pendingObjecifCount . ' ' . str('tenant')->plural($pendingObjectifCount) . ' ready to be used.');
```

Make the text plural when the count is more than one instead of writing "tenant(s)".

edit: added space after the first `$deployedCount` in the first line

src/Commands/CreatePendingTenants.php
```diff
@@ -0,0 +1,64 @@
...
+
+        return 1;
+    }
+
+    /**
+     * Calculates the number of pending tenants currently deployed
+     * @return int
```
lukinovec:
```suggestion
     * Calculate the number of currently deployed pending tenants.
     *
     * @return int
```

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public static function createPending($attributes = []): void
+    {
+        $tenant = static::create($attributes);
+
+        event(new CreatingPendingTenant($tenant));
+
+        // We add the pending value only after the model has then been created.
+        // this ensures the model is not marked as pending until the migrations, seeders, etc. are done
```
lukinovec:
```suggestion
        // Add the pending value only after creating the model
        // To ensure it's not marked as pending until finishing running the migrations, seeders, etc.
```

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    protected function addWithoutPending(Builder $builder)
+    {
+        $builder->macro('withoutPending', function (Builder $builder) {
+
+            // Only use whereNull('data->pending_since') when Laravel 6 support is dropped
+            // Issue fixed in Laravel 7 https://github.com/laravel/framework/pull/32417
+            $builder->withoutGlobalScope($this)
+                ->where('data->pending_since', 'like', 'null')
+                ->orWhereNull('data');
```
lukinovec: Would `orWhereNull` work in Laravel 6?

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,101 @@
...
+     * Apply the scope to a given Eloquent query builder.
+     *
+     * @param  Builder  $builder
+     * @param  Model  $model
+     * @return void
+     */
+    public function apply(Builder $builder, Model $model)
+    {
+        $builder->when(!config('tenancy.pending.include_in_queries'), function (Builder $builder){
+            $builder->whereNull('data->pending_since');
```
lukinovec: `whereNull` isn't used in the other methods because of Laravel 6 support. Shouldn't that be the case here too?

src/Concerns/HasATenantsOption.php
```diff
@@ -13,6 +13,7 @@ protected function getOptions()
     {
         return array_merge([
             ['tenants', null, InputOption::VALUE_IS_ARRAY | InputOption::VALUE_OPTIONAL, '', null],
+            ['with-pending', null, InputOption::VALUE_NONE, 'include pending tenants in query', null],
```
lukinovec: The trait's "HasA**Tenants**Option", so ideally, "with-pending" would be a separate trait

src/Tenancy.php
```diff
@@ -135,19 +136,24 @@ public function central(callable $callback)
...
 
         // Convert incrementing int ids to strings
         $tenants = is_int($tenants) ? (string) $tenants : $tenants;
 
         // Wrap string in array
         $tenants = is_string($tenants) ? [$tenants] : $tenants;
 
-        // Use all tenants if $tenants is falsey
-        $tenants = $tenants ?: $this->model()->cursor();
+        // Use all tenants if $tenants is false
```
lukinovec:
```suggestion
        // Use all tenants if $tenants is falsy
```

"falsey" was correct

src/Commands/CreatePendingTenants.php
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $this->info('Deploying pending tenants.');
+
+        $pendingObjectifCount = (int) ($this->option('count') ?? config('tenancy.pending.count'));
+
+        $pendingCurrentCount = $this->getPendingTenantCount();
```
lukinovec: `$pendingObjectifCount` could have a better name. Maybe `$maxPendingTenantCount`? Also maybe `$pendingCurrentCount` to `$pendingTenantCount`.

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,90 @@
...
+    public static function pullPendingTenant(bool $firstOrCreate = false): ?Tenant
+    {
+        if (!static::onlyPending()->exists()) {
+            if (!$firstOrCreate) {
+                return null;
+            }
+            static::createPending();
+        }
+
+        // At this point we can guarantee a pending tenant is free and can be called.
```
lukinovec:
```suggestion
        // At this point, we can guarantee a pending tenant is free and can be called
```

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,90 @@
...
+            'pending_since' => now()->timestamp
+        ]);
+
+        event(new PendingTenantCreated($tenant));
+    }
+
+    public static function pullPendingTenant(bool $firstOrCreate = false): ?Tenant
+    {
+        if (!static::onlyPending()->exists()) {
+            if (!$firstOrCreate) {
```
lukinovec:
```suggestion
            if (! $firstOrCreate) {
```

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Apply the scope to a given Eloquent query builder.
+     *
+     * @param  Builder  $builder
+     * @param  Model  $model
+     * @return void
+     */
+    public function apply(Builder $builder, Model $model)
+    {
+        $builder->when(!config('tenancy.pending.include_in_queries'), function (Builder $builder){
```
lukinovec:
```suggestion
        $builder->when(! config('tenancy.pending.include_in_queries'), function (Builder $builder){
```

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,90 @@
...
+
+
+    /**
+     * Determine if the model instance is in a pending state.
+     *
+     * @return bool
+     */
+    public function pending()
+    {
+        return !is_null($this->pending_since);
```
lukinovec:
```suggestion
        return ! is_null($this->pending_since);
```

src/Commands/ClearPendingTenants.php
```diff
@@ -0,0 +1,67 @@
...
+
+
+        $deletedPendingCount = tenancy()
+            ->query()
+            ->onlyPending()
+            ->when($savedExpiredDate->notEqualTo($expireDate), function (Builder $query) use ($expireDate) {
+                $query->where('data->pending_since', '<', $expireDate->timestamp);
+            })
+            ->get()
+            ->each // This makes sure the events or triggered on the model
```
lukinovec:
```suggestion
            ->each // Make sure the model events are triggered by deleting the tenants one by one
```

src/Commands/ClearPendingTenants.php
```diff
@@ -0,0 +1,67 @@
...
+            ->delete()
+            ->count();
+
+        $this->info("$deletedPendingCount pending tenant(s) deleted.");
```
lukinovec:
```suggestion
        $this->info($deletedPendingCount . ' pending ' . str('tenant')->plural($deletedPendingCount) . ' deleted.');
```

lukinovec (commented):
Most of the suggestions are code/comment style-related

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,90 @@
...
+        $tenant->update([
+            'pending_since' => now()->timestamp
+        ]);
+
+        event(new PendingTenantCreated($tenant));
+    }
+
+    public static function pullPendingTenant(bool $firstOrCreate = false): ?Tenant
+    {
+        if (!static::onlyPending()->exists()) {
```
lukinovec:
```suggestion
        if (! static::onlyPending()->exists()) {
```

src/Database/Concerns/PendingScope.php
stancl: FYI this is targeting `master` which will not support L6

src/Database/Concerns/PendingScope.php
stancl: Same as what I answered above. If there's any Laravel 6-specific code, it should be removed.

src/Tenancy.php
stancl: `falsy`, not `falsey`

lukinovec: okay, changing the suggestion to `falsy`

src/Commands/CreatePendingTenants.php
stein-j: Didn't know it could be written like that, awesome 👍🏼

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,101 @@
...
+     * @return void
+     */
+    protected function addWithoutPending(Builder $builder)
+    {
+        $builder->macro('withoutPending', function (Builder $builder) {
+
+            // Only use whereNull('data->pending_since') when Laravel 6 support is dropped
+            // Issue fixed in Laravel 7 https://github.com/laravel/framework/pull/32417
+            $builder->withoutGlobalScope($this)
+                ->where('data->pending_since', 'like', 'null')
```
lukinovec:
```suggestion
            $builder->withoutGlobalScope($this)
                ->whereNull('data->pending_since')
```

This doesn't have to support L6 (https://github.com/archtechx/tenancy/pull/869/files#r917901875)

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,101 @@
...
+     * @param  \Illuminate\Database\Eloquent\Builder  $builder
+     * @return void
+     */
+    protected function addOnlyPending(Builder $builder)
+    {
+        $builder->macro('onlyPending', function (Builder $builder) {
+
+            // Use whereNotNull when Laravel 6 is dropped
+            // Issue fixed in Laravel 7 https://github.com/laravel/framework/pull/32417
+            $builder->withoutGlobalScope($this)->where('data->pending_since', 'not like', 'null');
```
lukinovec:
```suggestion
            $builder->withoutGlobalScope($this)->whereNotNull('data->pending_since');
```

https://github.com/archtechx/tenancy/pull/869/files#r917901875

src/Database/Concerns/PendingScope.php
lukinovec: Added suggestions for removing the support:

https://github.com/archtechx/tenancy/pull/869#discussion_r918580920
https://github.com/archtechx/tenancy/pull/869#discussion_r918582076

src/Concerns/HasATenantsOption.php
lukinovec: Renamed the trait to "HasTenantOptions"

assets/config.php
```diff
@@ -35,6 +35,37 @@
...
+
+    /**
+     * Pending tenancy config.
+     * This is useful if you're looking for a way to always have a tenant ready to be used.
+     */
+    'pending' => [
+        /**
+         * If disabled, pending tenants will be excluded from all tenant queries. Unless if
+         * told otherwise with ::withPending() or ::onlyPending().
+         * Note: when disabled, this will also ignore tenants when runnings any tenants commands (migration, seed, etc.)
```
stancl: - The `told otherwise` line is weird, but keep the examples (`::withPending()` and `::onlyPending()`)
- `runnings` in second line, general grammar

assets/config.php
```diff
@@ -35,6 +35,37 @@
...
+         */
+        'include_in_queries' => true,
+        /**
+         * Defines how many tenants you want to be in a pending state.
+         * This value should be changed depending on how often a new tenant is created
+         * and how often you run the `tenancy:pending` command via the scheduler.
```
stancl: Not sure about this sentence. If it's meant to advise the user about setting the value, it should just say "This depends on the volume of tenants you're creating" or something like that. The "prepared pool" (perhaps we want to use some term like that, I like `pool`) depends on how often you're creating tenants.

stancl:
Check the grammar and clarity of all docblocks. The code itself seems fine. I'll do a complete review when you're done with those changes and mark this ready for review (use the request review feature).

assets/config.php
lukinovec: Right, maybe even the key could be something like 'pending.pool_size' instead of 'pending.count'

lukinovec: Also the CreatePendingTenants command's signature is `tenants:pending`, I don't think that captures what the command does. The signature could be better (e.g. `tenants:create-pending`)

assets/config.php
lukinovec: I updated the comments regarding the pending tenants in the config file

https://github.com/archtechx/tenancy/compare/4d7831d6f6a172760dd54f64e527ea9895768e02..a834efa692c0f05a431beab1155251f90cfaf12e

lukinovec: Something's up with the CI tests – after pushing the comment updates, there's this error:
image

assets/config.php
lukinovec: "Pending tenancy" or just "Pending" is used a lot here (like "Pending tenancy config", "Pending events"), and I'm not sure about keeping that. "Pending tenant" ("Pending tenant config" etc.) seems more appropriate because the feature allows "pending tenants". What do you think @stancl?

lukinovec: Updated the comments, and the naming of some variables, you can check the changes out [here](https://github.com/archtechx/tenancy/compare/4d7831d6f6a172760dd54f64e527ea9895768e02..stein-j-readied-tenant)

stancl: > pending.pool_size

Note that I'm not saying we should just stick the word `pool` here, we should use consistent terminology everywhere. I'm just saying that if we need a term to refer to this "pool of prepared tenants", `pool` *could* work well.

But the used terminology is up to you — at this point. I'll do a review when you fix all of the writing and then we can change the terminology if needed. Right now it should just be descriptive and consistent.

> The signature could be better

Agree.

> Pending tenancy

I don't like this term, it isn't very self-explanatory. "Pending tenants" is fine, pending tenancy is likely not what we want.

lukinovec: I think the writing is fine now. But I don't like [this config section](https://github.com/archtechx/tenancy/pull/869/files#diff-579787e0054e183a7e66b899216a02aef9cec3bdf11c4c298f052a14c0699b03R56-R64). Since the time constraints are used only in the ClearPendingTenants command, I think it would be better to just schedule the command with the time constraints as options instead of having them in the config.

stancl: > I think it would be better to just schedule the command with the time constraints as options instead of having them in the config

Hmm that could be a solution, yeah. I do like having all of the config in one place, but if it's making the config file significantly noisier and is only being used by the command, then we might move it there.

Two questions:
- Does the user have to schedule the command manually now?
- Can you provide an example for how the command would be scheduled with these options?

lukinovec: > Does the user have to schedule the command manually now?

Yeah. There was a comment saying "For this to work automatically, make sure to call the `tenancy:pending-clear` command in the scheduler." (changed it later so it's more concise, but that doesn't matter)

> Can you provide an example for how the command would be scheduled with these options?

`$schedule->command('tenants:pending-clear --older-days=5')`

or, if you'd set `older_than_days` to 5 in the config (the config value gets used if the option is null):

`$schedule->command('tenants:pending-clear)`

stancl: In that case yeah, we can make it an argument. I'd consider having it in the config if the command were scheduled automatically (which comes with a separate deal of complexity), but if the user is scheduling the command **anyway**, we can just specify those values as arguments.

src/Commands/ClearPendingTenants.php
```diff
@@ -39,11 +39,11 @@ public function handle()
 
         // Skip the time constraints if the 'all' option is given
         if (! $this->option('all')) {
-            if ($olderThanDays = $this->option('older-days') ?? config('tenancy.pending.older_than_days')) {
+            if ($olderThanDays = $this->option('older-than-days')) {
                 $expirationDate->subDays($olderThanDays);
             }
 
-            if ($olderThanHours = $this->option('older-hours') ?? config('tenancy.pending.older_than_hours')) {
+            if ($olderThanHours = $this->option('older-than-hours')) {
```
stancl: I think only one or the other should be used, not both.

lukinovec: Yeah, makes sense, I'll update it

stancl:
The failing tests seem to be related to some mock that is being used? The entire `PendingTenantsTest` is failing in the last run.

stancl:
Ah, it seems like the tests aren't converted to pest yet. We're already using pest in `master`. @abrardev99 Lukas has a pretty busy schedule now, could you try to do the test conversion here?

And regarding the one failing test (`run command with array of tenants works`), I'll look into that after the conversion is done.

abrardev99:
>  could you try to do the test conversion here?

Sure.

stancl:
Awesome, seems like only the `run command with array of tenants works` test is now failing. I'll look into that one myself. Thanks!

stancl:
Seems like the failing test got fixed now by merging `master` into this branch.

stancl:
@lukinovec Could you resolve the conflicts here? I think it should be simple, I assume this conflicts mostly with the changes I made when I was improving the v4 codebase, so for things like type changes you can probably just accept the updated version in `master`. But of course always double check every change/conflict.

lukinovec:
> Seems like the failing test got fixed now by merging master into this branch.

@stancl, the test is failing again after merging master. Maybe it fails randomly. On one commit, it ran fine, and right after that commit, there's the merge commit which doesn't change anything (the diff is empty), but the test failed there.
image

stancl:
Then try to add dummy commits to see if the teat fails/passes **randomly** or if it's because of the environment.

lukinovec:
> Then try to add dummy commits to see if the teat fails/passes **randomly** or if it's because of the environment.

I kept re-running the tests on [this commit](https://github.com/archtechx/tenancy/pull/869/commits/cd375283992f3961857812742ba7a8bb757d72df) and the test eventually passed

stancl:
@lukinovec So the issue is that the tests are failing randomly? Rather than some specific issue with the environment.

lukinovec:
> @lukinovec So the issue is that the tests are failing randomly? Rather than some specific issue with the environment.

It seems that way, yeah

stancl:
Then the issue is somewhere in the test implementation.

lukinovec:
> Then the issue is somewhere in the test implementation.

@stancl, I figured it out, the `Run` command uses the `getTenants()` method which fetches all tenants passed in the `tenants` option, and the fetched tenant collection is not always in the same order as the passed options.

Could you review the new commit? https://github.com/archtechx/tenancy/pull/869/commits/0ee54da29c619f2617e4a47ec40626c05115e083

stancl:
> and the fetched tenant collection is not always in the same order as the passed options

Do you mean that if you passed e.g. `--tenants=1 --tenants=2` it'd be `[2, 1]`? And what issues was this causing?

lukinovec:
> Do you mean that if you passed e.g. `--tenants=1 --tenants=2` it'd be `[2, 1]`? And what issues was this causing?

Yeah. The issue is that I'd expect the command to be called for the tenants in the order I pass them. The test was originally checking the order too, so I think running the commands for the tenants in a random order isn't right (even though it probably won't cause any big issues).

stancl:
Why does the order matter though? What issues was it causing?

lukinovec:
> Why does the order matter though? What issues was it causing?

Only that the test was failing

src/Concerns/HasTenantOptions.php
```diff
@@ -23,6 +27,9 @@ protected function getTenants(): LazyCollection
             ->when($this->option('tenants'), function ($query) {
                 $query->whereIn(tenancy()->model()->getTenantKeyName(), $this->option('tenants'));
             })
+            ->when(! is_null($this->option('with-pending')) && tenancy()->model()::hasGlobalScope(PendingScope::class), function ($query) {
```
stancl: Instead of checking for null, could we change the option above in `getOptions()` to be a boolean? Default to `false` and if specified, it will be true.

lukinovec: Right, that would make more sense

lukinovec: But maybe not – `with-pending` is a thing you'd want to specify because of the `pending.include_in_queries` config option. When that is `true`, you don't have to pass `--with-pending` to the command to include the pending tenants in it. In that case, defaulting to false would be wrong

stancl: Can we grab the default from the `pending.include_in_queries ` config?

lukinovec: Yeah, that seems like the cleanest way to do this – in `getOptions()`, default to `config('tenancy.pending.include_in_queries')` instead of `null`

lukinovec: I see that the option is `InputOption::VALUE_NONE`, which means it was meant to be passed only as `--with-pending` (without `=true` or `=false`). Making it accept `true`/`false` and default to the config value seems better though.

lukinovec: `InputOption` doesn't have anything like `VALUE_BOOLEAN`, which could cause some issues if the user passes anything else than `true`/`false` (`$query->withPending($this->option('tenants'))` could throw an exception because `withPending` accepts `bool`)

Also I can't find any tests for running commands with the `with-pending` option. I'll write some

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,89 @@
...
+        event(new CreatingPendingTenant($tenant));
+
+        // Update the pending_since value only after the tenant is created so it's
+        // Not marked as pending until finishing running the migrations, seeders, etc.
+        $tenant->update([
+            'pending_since' => now()->timestamp,
+        ]);
+
+        event(new PendingTenantCreated($tenant));
+    }
```
lukinovec: Because `Tenant::create()` returns the created tenant, I'd expect `Tenant::createPending()` to return the created tenant too

```suggestion
    public static function createPending($attributes = []): Tenant
    {
        $tenant = static::create($attributes);

        event(new CreatingPendingTenant($tenant));

        // Update the pending_since value only after the tenant is created so it's
        // Not marked as pending until finishing running the migrations, seeders, etc.
        $tenant->update([
            'pending_since' => now()->timestamp,
        ]);

        event(new PendingTenantCreated($tenant));
        
        return $tenant;
    }
```

src/Concerns/HasTenantOptions.php
lukinovec: The option works weirdly – not passing `--with-pending` defaults to the config value (as it should), but just `--with-pending` is recognized as `null` when it should be `true`. Passing `true`/`false` (or anything else) is recognized as a string, so we'll have to do `(bool) $this->option('with-pending')`

lukinovec:
@stancl, could you please look at [this commit](https://github.com/archtechx/tenancy/pull/869/commits/e3e85a0e88e83ccd249df2cc3d57fc569b555a0e) and let me know your opinion about the `$optionNotPassedValue` property?

I've added the `$optionNotPassedValue` property because before, I was checking if the user passed the option by checking if the option is a string (meaning the option's value is user's input, e.g. the user did `tenants:run --with-pending=true`) and when the option wasn't passed, it would be `false`. That's not right because the command can also get called using `Artisan::call()` – the user can pass the boolean directly, so I think the only correct way to conclude that the option was not passed would be defaulting to a very distinct value like `'not passed'`.

stancl:
Seems a bit too hacky to be in the package. I'd think it's better to have worse syntax (for users) and cleaner code than to have to use solutions like this.

If we can't support both `--with-pending` and `--with-pending=1`/`--with-pending=true`, that's fine. We just want some way of passing `--with-pending`.

Also, another simplification can be that **when the user enables including pending tenants in commands (via the config), we don't really need to provide a way for them to do `--without-pending` in the command.** It would be ideal to have this option, but if we can't have it then it's fine. Simply having 1) `--with-pending` that's used when the config key is false, and 2) the config key to auto-include pending tenants is fine.

lukinovec:
> Simply having 1) --with-pending that's used when the config key is false, and 2) the config key to auto-include pending tenants is fine.

Updated the code to work this way. I also resolved the merge conflicts

stancl:
So could you summarize the usage now? Or how the property works & how the config key works. Also, is this PR finished now?

lukinovec:
@stancl 

> how the config key works

If `config('tenancy.pending.include_in_queries')` is `false`, the pending tenants are not included in tenant queries (e.g. `Tenant::where(...)`) by default – you have to include them by adding `->withPending()` to the query, e.g. `$query->withPending()`. 

If `config('tenancy.pending.include_in_queries')` is `true`, the pending tenants will be included in all tenant queries by default. If you want to exclude them, you can do that by adding `->withoutPending()` to the query, e.g. `$query->withoutPending()`. 

There's also `$query->onlyPending()`.

> how the property works

If you mean the `$optionNotPassedValue` property I mentioned [here](https://github.com/archtechx/tenancy/pull/869#issuecomment-1268142448), I removed it. I think you might mean the `--with-pending` option though.

You'd use the option like `tenants:migrate --with-pending` to run the command for pending tenants too. This is the only thing the option does – you can't do e.g. `--with-pending=false` because the option is `VALUE_NONE`, so `--with-pending` is the only way to pass it.

So the option is useful only if `config('tenancy.pending.include_in_queries')` is `false` because if the config value is `true`, you don't need to pass `--with-pending`.

lukinovec:
> is this PR finished now?

If you agree with [the usage summary](https://github.com/archtechx/tenancy/pull/869#issuecomment-1269716863), I think we only have the naming improvements left to do.

stancl:
> I think you might mean the --with-pending option though.

Right, I meant to say option, not property.

And yeah, the simpler solution seems better now.

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,88 @@
...
+use Illuminate\Database\Eloquent\Scope;
+
+class PendingScope implements Scope
+{
+    /**
+     * All of the extensions to be added to the builder.
+     *
+     * @var string[]
+     */
+    protected $extensions = ['WithPending', 'WithoutPending', 'OnlyPending'];
```
stancl, line 18: @stein-j Is this a common pattern used by Laravel or just something you're using here?

lukinovec: I was curious about this before. I searched for the property, and Laravel uses the same thing in [SoftDeletingScope](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/SoftDeletingScope.php#L12) (looks like Laravel provides only this scope). So I think the `$extensions` property is fine here

src/Database/Concerns/HasPending.php
```diff
@@ -0,0 +1,91 @@
...
+namespace Stancl\Tenancy\Database\Concerns;
+
+use Stancl\Tenancy\Contracts\Tenant;
+use Stancl\Tenancy\Events\CreatingPendingTenant;
+use Stancl\Tenancy\Events\PendingTenantCreated;
+use Stancl\Tenancy\Events\PendingTenantPulled;
+use Stancl\Tenancy\Events\PullingPendingTenant;
+
+/**
+ * @property $pending_since
```
stancl: Add a type here

src/Database/Concerns/PendingScope.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Apply the scope to a given Eloquent query builder.
+     *
+     * @return void
+     */
+    public function apply(Builder $builder, Model $model)
+    {
+        $builder->when(! config('tenancy.pending.include_in_queries'), function (Builder $builder) {
+            $builder->whereNull('data->pending_since');
```
stancl: The `data->` is *potentially* problematic because someone may want to:
1. rename the data column
2. create a custom column

stancl: We should probably check for the data column name, and check if it's in the custom columns.

But we probably don't want to hardcode that here. So it'd be good to have an abstraction, perhaps in the package that handles the virtual columns.

Maybe a method like `getColumnForQuery()`/`generateColumnName()`/something along those lines.

src/Jobs/ClearPendingTenants.php
```diff
@@ -0,0 +1,27 @@
...
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        Artisan::call(\Stancl\Tenancy\Commands\ClearPendingTenants::class);
```
stancl: import

src/Jobs/CreatePendingTenants.php
```diff
@@ -0,0 +1,27 @@
...
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        Artisan::call(\Stancl\Tenancy\Commands\CreatePendingTenants::class);
```
stancl: import

tests/Etc/tmp/queuetest.json
```diff
@@ -0,0 +1 @@
+{"tenant_id":"The current tenant id is: acme"}
```
stancl: Can you:
1. delete this file, and commit that change (don't push)
2. add the file to `.gitignore`, commit
3. push
4. make sure the file is not shown on GH

stancl: It's also possible that I already added it to `.gitignore` and this file was created before I did that. So removing it might not have an effect. So you might need to do this before the steps I outlined above:
- delete the file from `.gitignore`, commit
- then the steps that I wrote above

stancl:
@lukinovec 
- In this commit, I fixed test names. You can review my changes https://github.com/archtechx/tenancy/pull/869/commits/e93a5181c363ea432dda6afb501612083f9ba569
- In [this commit](https://github.com/archtechx/tenancy/pull/869/commits/090fc475ebfbcd370b6d589635234cfdbca83a83), I added a method that is **guaranteed** to return a Tenant. Having a `?Tenant` typehint is not ideal, so I added this one to have a method that people can depend on to return a `Tenant` without having to do null checks (mostly for IDE/phpstan support). **This likely breaks the code**, and tests might be failing. Even if they're passing, you should still go through the code and find all references to `pullPendingTenant(` where something is passed as an argument.
    - Now, `pullPendingTenant()` has no arguments
    - `pullPendingTenantFromPool()` is the one that has the argument
- Question: do we have a test that checks that pulling a tenant removes him from the pending pool?

tests/CommandsTest.php
```diff
@@ -244,6 +239,67 @@
     $this->assertDirectoryDoesNotExist(public_path("public-$tenantKey"));
 });
 
+test('commands do not run for pending tenants if tenancy.pending.include_in_queries is false and the with pending option does not get passed', function() {
```
stancl: I think it'd be best to move all of these new tests to the `PendingTenantsTest`

stancl:
@stein-j A few questions about usage:
- The commands for creating and clearing pending tenants are essentially for managing the pending tenants manually, correct?
- The `count` in the config would be used by the package to **automatically** manage the pool? Or would the user be manually scheduling a job? How is this handled?
- If the `count` is used, is there some logic for pruning old tenants?
- The usage of this feature is basically to replace `Tenant::create()` e.g. in the registration controller with `Tenant::pullPendingTenant()`, right?
    - @lukinovec on that note, maybe it'd be cleaner to rename this to `pullPending()` to avoid the double "tenant" in the call

stancl:
I think this concludes the review on my end. So after these small changes are done, we can merge this.

Regarding the naming @lukinovec, we'll come up with a name later and just update the terminology in the code. But we want to merge this soon to be able to work on the other PRs.

The only thing I may need to review again is https://github.com/archtechx/tenancy/pull/869#discussion_r998077333. You'll probably be making a simple PR to the virtual column repo (it's in the `composer.json` dependencies).

stein-j:
@stancl everything you've said is on point.

The commands can be used manually otherwise you simply set the _job_ or the _command_ in the scheduler and it's at this moment the `count` config will be used automatically. Most packages do that (e.g [spatie documentation](https://spatie.be/docs/laravel-backup/v8/installation-and-setup#a-idcontent-scheduling-1-hrefcontent-scheduling-1-classanchor-link-aria-hiddentrue-titlepermalinkaa-idcontent-scheduling-hrefcontent-scheduling-classanchor-link-aria-hiddentrue-titlepermalinkascheduling)). 

I think you can define the scheduler in the service provider but it's usually not done that way because users want to keep control over those.

stancl:
So the answer to

> The count in the config would be used by the package to automatically manage the pool? Or would the user be manually scheduling a job? How is this handled?

is that users schedule the commands themselves, right?

And the answer to

> If the count is used, is there some logic for pruning old tenants?

is no? Meaning, if users just schedule the command, it will create n=`count` tenants, and then if the developer wants to add pruning logic, he'll schedule the clear job as well?

Could you give a few examples of how the jobs would be scheduled?

stein-j:
Yes, the users will have to schedule the commands themselves.

Yes there is some logic to pruning old tenants.

Here is an example of how I will probably be using it:

_/config/tenancy.php_
```php
    'pending' => [
        // The number of tenants I want to already be deployed in advance, here 5
        'count' => env('TENANCY_PENDING_COUNT', 5),
    ],
```
_/app/Console/Kernel.php_
```php
        // This command is optional, in our case we seed a lot of data (statuses, types, etc.)
        // for the user so deleting unused tenants for more than 7 days will
        // make sur the  new seeders will be used.
        $schedule->command('tenants:pending-clear', [
                'older-than-days' => 7
        ])->daily()->at('01:00');

        // We don't have that many people deploying new tenants to running the command daily is totally fine,
        // but for someone with many deployment could run the command hourly, or even weekly
        // This will ensure there's always at most 5 tenants are pre-deployed 
        $schedule->command('tenants:pending-create')->daily()->at('02:00');
```

I could even call `php artisan tenants:pending-clear --all` whenever we deploy a new version to reset everything.

tests/Etc/tmp/queuetest.json
lukinovec: Did this and the file is not there anymore 👍 resolving this

tests/PendingTenantsTest.php
```diff
@@ -0,0 +1,204 @@
...
+
+test('tenancy can create if none are pending', function () {
+    expect(Tenant::all()->count())->toBe(0);
+
+    Tenant::pullPendingTenant(true);
+
+    expect(Tenant::all()->count())->toBe(1);
+});
+
+test('the include in queries global scope can include pending tenants in all queries', function () {
```
lukinovec:
```suggestion
test('pending tenants are included in all queries depending on the value of the include in queries config', function () {
```

@stancl I think I'd change the test name to this since we're testing if changing the config value makes queries include the pending tenants. Also, "include in queries global scope" doesn't seem accurate since the scope is named "PendingScope"

lukinovec: Other than this, the test names seem fine

tests/PendingTenantsTest.php
```diff
@@ -0,0 +1,204 @@
...
+
+test('tenancy can pull a pending tenant', function () {
+    expect(Tenant::pullPendingTenant())->toBeNull();
+
+    Tenant::createPending();
+
+    expect(Tenant::pullPendingTenant(true))->toBeInstanceOf(Tenant::class);
+});
+
+test('tenancy can create if none are pending', function () {
```
lukinovec: I'm a bit confused by this test name @stancl. Seems like the test checks that when we attempt to pull a pending tenant, and there isn't any, a tenant gets automatically created. Does that seem right?

lukinovec:
> do we have a test that checks that pulling a tenant removes him from the pending pool?

Seems like we don't

EDIT: [added it](https://github.com/archtechx/tenancy/pull/869/commits/0769e187acae6ca9b0144b925ee8f52aa611a4e5) along with the change I suggested [here](https://github.com/archtechx/tenancy/pull/869#discussion_r1001536066)

tests/PendingTenantsTest.php
```diff
@@ -0,0 +1,204 @@
...
+    Tenant::query()->delete();
+
+    expect(Tenant::onlyPending()->exists())->toBeFalse();
+
+    Tenant::createPending();
+
+    expect(Tenant::onlyPending()->exists())->toBeTrue();
+});
+
+test('tenancy can pull a pending tenant', function () {
```
lukinovec, line 95: This tests something else than the test name suggests. Since the pullPending method is different now, I think the test could be as simple as `expect(Tenant::pullPending())->toBeInstanceOf(Tenant::class)`

tests/PendingTenantsTest.php
```diff
@@ -0,0 +1,204 @@
...
+    config(['tenancy.pending.older_than_days' => 4]);
+
+    Artisan::call(ClearPendingTenants::class, [
+        '--all' => true
+    ]);
+
+    expect(Tenant::onlyPending()->count())->toBe(0);
+});
+
+test('tenancy can check for pending tenants', function () {
```
lukinovec:
```suggestion
test('tenancy can accurately check if there are any pending tenants', function () {
```

Maybe this name would be a bit better? Also I'm not sure if this test is needed. I think there could just be a test that checks if `Tenant::onlyPending()` works correctly. If that method works correctly, we can assume that `->exists()` works too (as well as other query methods).

src/Database/Concerns/PendingScope.php
lukinovec: I opened a PR that adds the method to the package https://github.com/archtechx/virtualcolumn/pull/9

tests/PendingTenantsTest.php
stancl: Yeah, the test body seems to assert that the method for pulling a tenant creates a tenant if the pending tenant pool is empty.

tests/PendingTenantsTest.php
stancl: The test does seem to be a bit confused, yeah. The first expectation basically checks that the pool is empty, then we add a tenant to the pool, and then we pull from the pool (though the `true` makes it seem like we're forcing creation?)

It should be changed to simply test that "tenancy can pull a pending tenant" as the name suggests. So the test should be simplified — the first expectation removed, and the `true` removed (but with new code, we want to use the method I added).

> I think the test could be as simple as

This isn't exactly accurate since that doesn't test 1) creating a pending tenant 2) pulling from the pending pool. It just forcibly creates a tenant behind the scenes.

tests/PendingTenantsTest.php
stancl: re: name — agree on the change but get rid of the `accurately`, there's no way you could get inaccurate or semi-accurate results here. It either can check or it can't.

> I think there could just be a test that checks if

Do you mean adding a new test?

tests/PendingTenantsTest.php
stancl: Agree that `include in queries` is not exactly the global scope, but the original sentence seems to have a better structure.

Also in your suggested change: `are included ... depending on` doesn't make sense, it should be `can be included ... depending on` (or `based on`). Or `are included ... when`

Also I wouldn't go too verbose here, `the value of the` can be omitted.

tests/PendingTenantsTest.php
lukinovec: > there's no way you could get inaccurate or semi-accurate results here. It either can check or it can't.

Yeah, you're right.

```suggestion
test('tenancy can check if there are any pending tenants', function () {
```

> Do you mean adding a new test?

 I meant we could remove this test along with some other tests and add a test that checks if `Tenant::onlyPending()` works correctly because that should be enough (I saw this test essentially as `check if eloquent's exist() method works`). But this test is the only one that we'd get rid of, so now, I don't think the idea was that great.

tests/PendingTenantsTest.php
lukinovec: > though the true makes it seem like we're forcing creation?

Exactly

> This isn't exactly accurate since that doesn't test 1) creating a pending tenant 2) pulling from the pending pool. It just forcibly creates a tenant behind the scenes.

I see, I'll push the updated test in a second

tests/PendingTenantsTest.php
stancl: We *could* remove this test since the method is being used in other tests, but if it's a super short test, we can just keep it.

tests/PendingTenantsTest.php
lukinovec: @stancl https://github.com/archtechx/tenancy/pull/869/commits/a13384dedeeefa8f240435d4608fd7cf5dcf6ecf the test should be fine now, what do you think?

src/Database/Concerns/PendingScope.php
lukinovec: @stancl I also updated the PendingScope to use the methods. I'll look for other uses of `data->` added in this PR and I'll also update that

EDIT: found only one other instance of `data->` [here](https://github.com/archtechx/tenancy/commit/69e37ffa300fb1ce00d50910e7a68b262b9f3292). If my changes are OK, this can be resolved

EDIT2: the change from the previous edit broke a test, so this isn't resolved yet

EDIT3: found out that was due to a dumb mistake, I pushed the fix now

tests/PendingTenantsTest.php
lukinovec: I see

```suggestion
test('pending tenants are included in all queries based on the include_in_queries config', function () {
```

I also omitted the `all` from `all queries` (I think `queries` already implies `all queries`), and I added underscores so that it's clear we're talking about a config key (not sure about having special characters in the test names, but the tests run fine, so I think it's OK to keep the underscores). What do you think about the new suggestion?

tests/PendingTenantsTest.php
lukinovec: I updated the name

https://github.com/archtechx/tenancy/pull/869/commits/18d12f3a43f2ed9609bafc13f658fe41fc28e8a7

tests/PendingTenantsTest.php
lukinovec: > but if it's a super short test, we can just keep it

Right, I'd keep it since it's short

tests/PendingTenantsTest.php
stancl: > I think queries already implies all queries

The `all` is the key emphasis here, so I'd keep it. That's the point of the option — instead of including the tenants only in *some* queries, they are included in *all* queries.

And yeah, I omitted underscores initially but I think we can use any characters in test names now that we're using pest.

tests/PendingTenantsTest.php
stancl: Looks good. You can mark this as resolved

tests/PendingTenantsTest.php
stancl: Remove the `Tenant::query()->delete();` call, update the name as suggested above, and you can mark this as resolved 👍🏻

tests/PendingTenantsTest.php
lukinovec: > That's the point of the option — instead of including the tenants only in some queries, they are included in all queries

Right. I'll edit the suggestion and commit it

src/Database/Concerns/PendingScope.php
stancl: Okay, we can mark this as resolved then

tests/PendingTenantsTest.php
```diff
@@ -0,0 +1,210 @@
...
+    Tenant::createPending();
+
+    config(['tenancy.pending.include_in_queries' => false]);
+
+    expect(Tenant::all()->count())->toBe(0);
+
+    config(['tenancy.pending.include_in_queries' => true]);
+
+    expect(Tenant::all()->count())->toBe(1);
+    Tenant::all();
```
stancl: Seems like this line is left over from some development?

stancl:
All reviews are resolved now except the one I just added. If you confirm that the call can be removed and isn't there to trigger some logic @lukinovec, we can merge this PR.

I'm still not 100% set on the terminology, but we want the changes to be in `master` so that we can work on other PRs. The only downside is that if we want to rename `pending` later, it will be another commit with a big refactor (instead of that being part of the commit that adds the feature). But that's alright and worth having the changes in `master`.

I'm ~80% sure we'll go with `pending` because I can't think of anything better, though intuitively it doesn't feel like the ideal term.

stein-j:
The only other term I could think of is `waitList`.
You can enable the `tenant wait list`. You then add tenants to the wait list and can pull a `wait listed tenant`.

tests/PendingTenantsTest.php
lukinovec: Yeah, it seems that way
```suggestion
```

lukinovec:
To me, `pre-created` or `premade` seems better than `pending` (I'd pick `premade` – there's no '-') because it makes the feature more intuitive. With both `premade` and `pending`, we'd have to explain what the term means, but I think I'd understand the feature much faster if `premade` was used instead of `pending`. "Premade tenant" also doesn't seem less accurate than "pending tenant".

lukinovec:
Also @stancl, I resolved the review (removed the line, it was extra)

stancl:
Yeah, something like `premade` seems the most accurate to me, though `premade` isn't really perfect either. The nice thing about "pending" is that it implies that the tenants are awaiting something (usage). And with "pending" it makes more sense conceptually that we can just prune all pending tenants and re-create them. "Premade" sounds a bit more "late stage" than "pending".

stancl:
Thanks to both for working on this!

stein-j:
Yay ! Very happy to see this merged.

We will be using this for our solution, working the original PR was fun for me. I hope this will be useful for many others.

bdsumon4u:
> Yay ! Very happy to see this merged.
> 
> We will be using this for our solution, working the original PR was fun for me. I hope this useful for many others.

I'm also happy. Because I need this feature.
Now badly waiting for v4.
Thank You All.

Issue #547, opened 2020-12-05
Run tenant aware query from central/tenants context 


bliveinhack:
#### Describe the bug
- When we use $tenant->run() to perform query on another tenants from central/tenants context it should work as expected.
for example : 
`$tenant->run(function () {
    User::create(...);
});`

Above code works fine when we have session driver set to `SESSION_DRIVER=files`.
But if we set session driver to `SESSION_DRIVER=database` . it fails and give `Call to a member function prepare() on null.`

#### Steps to reproduce
- Make env file to use `SESSION_DRIVER=database`
- Run tenants specific query.

#### Expected behavior
- It should run the tenant specific query and return to central or whatever tenants we calling from.

#### Your setup
 - Laravel version: 6.2.x
 - stancl/tenancy version: 3.4.0
 - Storage driver: DB
- Session Driver : DB

jandominik:
Agree, have a same problem. I'm running command which is going through all Tenants
`tenancy()->runForMultiple(null, function (Tenant $tenant) { ... });`
Problem happens when calling anything from database but only on second tenant and more. First tenant always pass.

Ad 1: Problem is happening even if I try to run Command using database via `php artisan tenants:run my:command`.
Ad 2: Found that this problem is related to way of using database, because problem is happening only when using criterias of package https://packagist.org/packages/prettus/l5-repository. When I skip all criterias (which should not be anything else than a list of conditions), problem has gone. If it will be useful, I can write more detailed scenario how to reproduce.

Laravel: 7.28.4
stancl/tenancy: 3.2.1

stancl:
If you could share an exception via Flare and showed the _exact_ code that reproduces this (without Jetstream, events firing from the tenant call, ...) I can take a look at this

ngugijames:
Here is sample code to replicate this error:
```
   /**
     * @return \Illuminate\Http\Response
     */
    public function getToolbox()
    { 
        $tenants = Tenant::get();
        
        $businesses = [];
        $locations = [];
        
        tenancy()->runForMultiple($tenants, function ($tenant) use (&$businesses, &$locations) {
            $businesses[$tenant->id] = $tenant->business_name;
            
            //for each tenant, get the business locations
            $locations[$tenant->id] = BusinessLocation::pluck('name', 'id')
                                                      ->toArray();
        });
        
        return view('core::toolkit.index', compact('businesses', 'locations'));
    }
```

**Your setup**
Laravel version: 7.x
[...]

PR #874, opened 2022-06-08
[4.x] Add batch tenancy queue bootstrapper

Riley19280:
This PR adds a bootstrapper class for working with Laravel job batches. This fixes issues related to #547, where a `Call to a member function prepare() on null` exception gets thrown when the queue worker attempts to get a batch from the database for a different tenant than the previous one the queue worker processed. This is fixed by temporarily replacing the BatchRepository connection with the correct tenant connection, and then reverting it back when exiting the tenant context.


stancl:
I don't think #547 is related, it's simply the same error message

Riley19280:
Yeah, that is entirely possible. I still believe there is value in supporting the batch system though.

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,51 @@
...
+             * Access the resolved batch repository instance and update its connection to use the tenant connection
+             */
+            $batchRepositoryReflection = new ReflectionClass($batchRepository);
+            $connectionProperty        = $batchRepositoryReflection->getProperty('connection');
+            $connectionProperty->setAccessible(true);
+            $connection = $connectionProperty->getValue($batchRepository);
+
+            $this->previousConnection = $connection;
+
+            $connectionProperty->setValue($batchRepository, DB::connection('tenant'));
```
stancl: Aren't there any getters/setters whatsoever?

If not, and there's a valid use case for modifying the connection (like what we do here), then we could PR the getters & setters to laravel/framework. They typically merge simple PRs like that when they come with good explanations of why the changes are useful.

tonning:
Thank you @Riley19280, this was a lifesaver for me trying to figure out why batch jobs would fail with the `Call to a member function prepare() on null` exception on subsequent dispatches when the first and second dispatch are from different tenants.

For what it's worth @stancl, I added this as a bootstrapper on `v3` and it's been work flawless so far.

tests/BatchTest.php
```diff
@@ -0,0 +1,71 @@
...
+use Stancl\Tenancy\Listeners\RevertToCentralContext;
+use Stancl\Tenancy\Tests\Etc\Tenant;
+
+class BatchTest extends TestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->app->singleton(BatchTenancyBootstrapper::class);
```
stancl: I don't think tests should be doing this, the package code should take care of it

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,51 @@
...
+use Illuminate\Bus\BatchRepository;
+use Illuminate\Bus\DatabaseBatchRepository;
+use Illuminate\Support\Facades\DB;
+use ReflectionClass;
+use Stancl\Tenancy\Contracts\TenancyBootstrapper;
+use Stancl\Tenancy\Contracts\Tenant;
+
+class BatchTenancyBootstrapper implements TenancyBootstrapper
+{
+    private $previousConnection = null;
```
stancl: Add docblock, type, and change this to `protected`

src/Bootstrappers/BatchTenancyBootstrapper.php
stancl: @abrardev99 could you look into this? Seems like OP can't finish the PR now.

abrardev99: Sure.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: Added.

tests/BatchTest.php
abrardev99: The test failed when I removed this line. 
image

 But, strangely, only the last assertion failed.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: > Aren't there any getters/setters whatsoever?

I see there are no getters and setters for connection. I'll send the PR to Laravel, and after merging, I'll implement it in this review. Then we also need to use the Latest Laravel release. 

To ensure PR doesn't reject, I'll make the changes in `DatabaseBatchRepository` class. Correct? 
And changes will be. (Just check last two methods)
```
<?php

namespace Illuminate\Bus;

use Carbon\CarbonImmutable;
use Closure;
use DateTimeInterface;
use Illuminate\Database\Connection;
use Illuminate\Database\PostgresConnection;
use Illuminate\Database\Query\Expression;
use Illuminate\Support\Str;

class DatabaseBatchRepository implements PrunableBatchRepository
{
    /**
     * The batch factory instance.
     *
     * @var \Illuminate\Bus\BatchFactory
     */
    protected $factory;

    /**
     * The database connection instance.
     *
     * @var \Illuminate\Database\Connection
     */
    protected $connection;

    /**
     * The database table to use to store batch information.
     *
     * @var string
     */
    protected $table;

    /**removed existing methods to showcase current changes**/
    
    /**
     * Set the desired connection for the batch.
     *
     * @return \Illuminate\Database\Connection
     */
    public function getConnection()
    {
        return $this->connection;
    }

    /**
     * get the batch current connection
     *
     * @param  \Illuminate\Database\Connection  $connection
     */
    public function setConnection(Connection $connection)
    {
        $this->connection = $connection;
    }
}

```

Then we can make use of these methods in this PR and get rid of reflection methods.

stancl: The docblocks are switched up and not formatted well but yeah, you can try to PR those methods.

tests/BatchTest.php
stancl: See if other bootstrappers are ever registered as singletons (in the service provider). As I said, it *may* be correct to register it as a singleton, but it's likely not correct to do it in a test.

stancl: Bootstrappers are registered as singletons: https://github.com/archtechx/tenancy/blob/3.x/src/TenancyServiceProvider.php#L53

Therefore registering the singleton in the test again is unnecessary.

So the issue is likely that the bootstrapper wasn't added to the config early enough in the lifecycle?

assets/config.php
```diff
@@ -33,6 +33,7 @@
         Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\QueueTenancyBootstrapper::class,
         // Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper::class, // Note: phpredis is needed
+        // Stancl\Tenancy\Bootstrappers\BatchTenancyBootstrapper::class,
```
stancl: @Riley19280 Could this be enabled by default? Would that lead to any issues?

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: Ah, How dumb 😬. Thanks for pointing out.

abrardev99: I've sent the [PR](https://github.com/laravel/framework/pull/43869). Once merged, I'll update this PR.

tests/BatchTest.php
abrardev99: > So the issue is likely that the bootstrapper wasn't added to the config early enough in the lifecycle?

I think the issue is something else. 

> Bootstrappers are registered as singletons: [3.x/src/TenancyServiceProvider.php#L53](https://github.com/archtechx/tenancy/blob/3.x/src/TenancyServiceProvider.php?rgh-link-date=2022-08-25T17%3A45%3A32Z#L53)

Yes, Bootstrappers are registered. But I noticed TenancyServiceProvider picking config from assets/config. Here I dump the Bootstrappers in the register loop. 
image

These Bootstrappers are always the same doesn't matter what we define as a config in tests. 

As RedisBootstrapper is commented, you are manually registering it in TestCase.php, so Redis tests would pass. https://github.com/archtechx/tenancy/blob/79e3d53b06f33aa6e30ee4454177e1d798918704/tests/TestCase.php#L108

So either we should enable it by default ([I see you asked](https://github.com/archtechx/tenancy/pull/874#discussion_r955257750)) or register it in TestCase.php class.

stancl: > enable it by default

We can also just enable it in the test.

stancl: > These Bootstrappers are always the same doesn't matter what we define as a config in tests

You can set config values and it will work, it just has to be done at the right point in the lifecycle. If setting `config(['foo' => 'bar'])` doesn't work somewhere, a different hook needs to be used.

But I'm 99% sure that using `config(['foo' => 'bar'])` at the start of the test or in `setUp()` will work.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: PR got merged. https://github.com/laravel/framework/pull/43869#event-7268635821

I'll update this PR once Laravel releases the latest tag this Tuesday.

tests/BatchTest.php
abrardev99: > But I'm 99% sure that using config(['foo' => 'bar']) at the start of the test or in setUp() will work.

I tried both. I set the config in setUp method and also at the start of the tests. Nothing works. Maybe I'm missing something but when I dump in the singleton registration, and in tests that how the flow looks like. 

image

You can see, all the singleton above. The setUp dump call is in `setUp` method. It's being executed after the singletons got registered.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: Updated 👍 and now resolving the review.

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,42 @@
...
+use Illuminate\Bus\BatchRepository;
+use Illuminate\Bus\DatabaseBatchRepository;
+use Illuminate\Support\Facades\DB;
+use Stancl\Tenancy\Contracts\TenancyBootstrapper;
+use Stancl\Tenancy\Contracts\Tenant;
+
+class BatchTenancyBootstrapper implements TenancyBootstrapper
+{
+    /**
+     * The database previous connection instance.
```
stancl:
```suggestion
     * The previous database connection instance.
```

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @var \Illuminate\Database\Connection
+     */
+    protected $previousConnection = null;
+
+    public function bootstrap(Tenant $tenant)
+    {
+        $batchRepository = app(BatchRepository::class);
```
stancl: Could this be a property on the class instead of something we resolve real-time?

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,42 @@
...
+
+        if ($batchRepository instanceof DatabaseBatchRepository) {
```
stancl: What other types could this be? @abrardev99

tests/BatchTest.php
abrardev99: As discussed internally, resolving this by registering the singleton in TestCase with todo to find a nice way afterward.

assets/config.php
abrardev99: > Could this be enabled by default? Would that lead to any issues?

I think we can enable it by default, and there will be no issue. I enabled it by default, and all tests passed since it's switching batch connection only when tenancy is initialized. 

My only concern is whether we need to enable it by default if the app is not using BatchJobs. It's similar to `RedisTenancyBootstrapper`. We don't need it until we are using Redis.

stancl: > We don't need it until we are using Redis

No — it's not a matter of *need*. Redis is disabled because *it depends on a specific environment (phpredis)*.

If the Batch tenancy bootstrapper won't cause any issues by being enabled by default, then we can keep it enabled.

stancl:
@abrardev99 Seems like GH was bugged again and some of my reviews didn't get submitted, they were "Pending". I submitted them now.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: Yes, we can. I've converted it into the property, and it's working fine.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: Looks like `DatabaseBatchRepository` is the only type because the [interface binds to `DatabaseBatchRepository` concrete implementation in BusServiceProvider](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Bus/BusServiceProvider.php#L44). Also, Laravel batch achieved using database table only. 
Screenshot 2022-09-13 at 11 49 55 AM

Tests are passing after removing the `if` check. 
So the only reason to keep the check is maybe Laravel will add more options for Batching. Also, I found that `BatchRepositoryFake` also implements the same interface.

stancl: > Also, I found that BatchRepositoryFake also implements the same interface.

Could you check if the reason for the `if` check was that? Meaning: if you try to fake the batch repository, and the if check is removed, then the code will run into some errors.

assets/config.php
stancl: Seems like my reply to this didn't get submitted either. I wrote that Redis is unique because **there's an environment dependency (phpredis)**, that's why it's disabled by default.

We don't really care about which parts of Laravel the developer is using. We just disable Redis by default because the package will run into exceptions if the developer isn't using phpredis.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: > Could you check if the reason for the if check was that?

After looking into the source code, I see`BatchRepository ` is always resolved to `DatabaseBatchRepository` no matter if we use the Fake. I've tried `Bus::fake` in tests. I think that it makes sense because it looks like `BatchRepositoryFake` [is being used internally](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Testing/Fakes/BusFake.php#L684-L689). 

Even after this, I'm still no 100% that we should remove the `if` check.

assets/config.php
abrardev99: I see. 

But my point is still valid. 
> My only concern is whether we need to enable it by default if the app is not using BatchJobs.

Not every app will need batching. If the developer uses batch, he will enable this bootstrapper otherwise, keeping it disabled is okay.

src/Bootstrappers/BatchTenancyBootstrapper.php
stancl: Then try diving into that a bit deeper I guess, so that we're completely sure what code we should use here.

assets/config.php
stancl: Sure, but not every app will need queues either. Yet we keep the queue bootstrapper enabled by default. I think the bootstrapper **doesn't do anything unless it's used**. Therefore it can simply be enabled by default. Again, the only reason why we'd even disable any bootstrapper in the first place — what we did with Redis — is environment requirements that would make the app fail.

abrardev99: >  is environment requirements that would make the app fail.

Make sense. 

So going to enable the bootstrapper by default.

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: So I dug it up again today. In class, the interface always resolves to `DatabaseBatchRepository` no matter whether I do `Bus::fake` or not. When faking Bus, use the `BatchRepositoryFake` [internally](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Testing/Fakes/BusFake.php#L684-L689). 

Since, I'm not able to be 100 clear that if interface would resolve anything else, I talked to Lukas. 

image

So we might want to keep the check because of the connection getter and setter. Also, I liked his suggestion:
> So either I'd keep the if, or make the $batchRepository property DatabaseBatchRepository

stancl: @lukinovec 

> In that case, I'd definitely keep the if.

If it's only present on class X, **and the variable can also only ever be type X**, then the `if` is unnecessary. But I guess it can help with IDE support/static analysis and narrowing the type within that scope.

It's fine if there are more implementations of the interface, but I'm wondering if there are any cases when this would resolve as those other types.

Regarding the idea to hardcode the `DatabaseBatchRepository` type, that *could* work but there are more things to keep in mind:
1. It's possible that **the correct way to resolve the batch repository is to resolve the interface**. For example, the interface may be bound as a singleton, or have `->extend` bindings etc. With `DatabaseBatchRepository` only being the class which implements the interface, but **not the class that has this "important position" in the service container**.
2. I generally don't like typehinting implementations in case someone used a different one (e.g. wrote a `RedisBatchRepository`), but the `if` check has the same problem, so typehinting isn't any worse. Also, people can just extend or replace the `BatchTenancyBootstrapper`

Regarding point 1), I would:
1. Try changing the typehint and seeing if the code still works
2. Dive into Laravel source code to see if the interface has any special status in the service container, compared to the DB implementation class

abrardev99: I think we should not typehint `DatabaseBatchRepository ` mostly because of Samuel's "2. I generally don't like..." point. 
The main issue here is whether keep the `if` check or not. It depends on the interface resolution. As I stated above, according to my discovery, I found the interface bound singleton to only the `DatabaseBatchRepository` class. The 
The `BatchRepositoryFake` class also implements the interface, but it never resolved this class. @lukinovec can you please try what other classes the interface can resolve and make sure if we can remove the `if` check? You can read my replies above for more information.

stancl: Seems like the interface has no special logic (e.g. `->extend` calls), it simply resolves to the DB repository:

```php
protected function registerBatchServices()
{
    $this->app->singleton(BatchRepository::class, DatabaseBatchRepository::class);

    $this->app->singleton(DatabaseBatchRepository::class, function ($app) {
        return new DatabaseBatchRepository(
            $app->make(BatchFactory::class),
            $app->make('db')->connection($app->config->get('queue.batching.database')),
            $app->config->get('queue.batching.table', 'job_batches')
        );
    });
}
```

So my view is that:
1. In theory, typehinting the interface would be more correct, and then the `if ()` check adds more IDE support and help for static analysis
2. In practice, typehinting the interface and pairing it with an `if ()` check is a bit messy. When it comes to the behavior, typehinting the DB implementation will probably work because they're *both* singletons

So ultimately: typehinting the DB implementation is a bit less correct, but it's more clear. Therefore: **assuming everything works well with that approach**, we'll do that.

Try changing the typehint to use `DatabaseBatchRepository`, and if everything works well, commit it here.

abrardev99: > Try changing the typehint to use DatabaseBatchRepository, and if everything works well, commit it here.

Just committed the changes. Everything is working, and the tests are passing.

stancl:
- Keep the property name `$batchRepository`
- Since we're using dependency injection, we shouldn't be using the `DB::` facade in the code. Use DI for that as well. It should be something like Laravel's `DatabaseManager` I think

abrardev99:
>Keep the property name $batchRepository
Since we're using dependency injection, we shouldn't be using the DB:: facade in the code. Use DI for that as well. It should be something like Laravel's DatabaseManager I think

Done.

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,42 @@
...
+use Stancl\Tenancy\Contracts\Tenant;
+
+class BatchTenancyBootstrapper implements TenancyBootstrapper
+{
+    /**
+     * The previous database connection instance.
+     *
+     * @var \Illuminate\Database\Connection
+     */
+    protected $previousConnection = null;
```
stancl: You can add the typehint here and remove the annotation

assets/config.php
```diff
@@ -33,6 +33,7 @@
         Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\QueueTenancyBootstrapper::class,
         // Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper::class, // Note: phpredis is needed
+        Stancl\Tenancy\Bootstrappers\BatchTenancyBootstrapper::class,
```
stancl, line 35: Move this above Redis so that the commented out line is last

src/Bootstrappers/BatchTenancyBootstrapper.php
```diff
@@ -0,0 +1,41 @@
...
+    public function __construct(
+        protected DatabaseBatchRepository $batchRepository,
+        protected DatabaseManager $databaseManager
+    ) {
+    }
+
+    public function bootstrap(Tenant $tenant)
+    {
+        // Update batch repository connection to use the tenant connection
+        $this->previousConnection = $this->batchRepository->getConnection();
```
stancl, line 29: I wasn't sure if this wouldn't get overridden when tenants are initialized in a sequence, without tenancy being ended first. So I updated the tests and they still pass. Seems like `revert()` is called even when we're calling `initialize()` from another tenant's context.

stancl: https://github.com/archtechx/tenancy/pull/874/commits/395d8597bb0b041ce13f637b5d8b915c42b80695

stancl:
Thanks for your work on this guys!

src/Bootstrappers/BatchTenancyBootstrapper.php
abrardev99: Thanks 👍

MujeebAnwar:
@tonning I'm trying to add this as a bootstrapper on `v3`. But it's given me an `Unresolvable dependency resolving [Parameter #2 [ <required> string $table ]] in class Illuminate\Bus\DatabaseBatchRepository` Error. Can you please help me to figure it out?

abhiburk:
On multiple queue jobs for tenant, I am facing a similar issue I tried adding this bootstrapper within `v3`

`Call to a member function prepare() on null {"exception":"[object] (Error(code: 0): Call to a member function prepare() on null`

Could someone please help me with this ?

abhiburk:
Ok I think the easiest way was to create a new file in your app `BatchTenancyBootstrapper.php` and 
add it to the `config/tenancy.php` `bootstrappers` array

PR #881, opened 2022-06-23
[4.x] Use PHP CS Fixer

### LINKED ISSUE #767: Migrate to phpcsfixer
We should move from styleci to our phpcsfixer config in [archtechx/template](https://github.com/archtechx/template)

erikgaal:
Fixes #767

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,29 @@
+name: Lint
+
+on: [push]
+
+jobs:
+  php-cs-fixer:
+    name: PHP CS Fixer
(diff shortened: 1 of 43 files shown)


stancl:
Thanks for the PR. See https://github.com/archtechx/template/blob/master/.php-cs-fixer.php, #767 mentioned that we should be using config from there.

And the workflow can also be the same, you can just add this section to our existing workflow file: https://github.com/archtechx/template/blob/master/.github/workflows/ci.yml#L42-L56

stancl:
Could you change the workflow as well?

PR #883, opened 2022-06-27
[4.x] Don't use onDeleteCascade in the migrations

### LINKED ISSUE #770: Don't use onDeleteCascade in the migrations
https://github.com/archtechx/tenancy/issues/654#issuecomment-876546291

abrardev99:
Related issue https://github.com/archtechx/tenancy/issues/770

### DIFF
--- a/assets/TenancyServiceProvider.stub.php
+++ b/assets/TenancyServiceProvider.stub.php
@@ -45,7 +45,7 @@ class TenancyServiceProvider extends ServiceProvider
                     Jobs\DeleteDomain::class,
                 ])->send(function (Events\TenantDeleted $event) {
                     return $event->tenant;
-                })->shouldBeQueued(false), // `false` by default, but you probably want to make this `true` for production.
+                })->shouldBeQueued(false),
             ],
             Events\TenantDeleted::class => [
                 JobPipeline::make([
--- a/assets/TenancyServiceProvider.stub.php
+++ b/assets/TenancyServiceProvider.stub.php
@@ -42,7 +42,7 @@ class TenancyServiceProvider extends ServiceProvider
             Events\TenantUpdated::class => [],
             Events\DeletingTenant::class => [
                 JobPipeline::make([
-                    Jobs\DeleteDomain::class,
+                    Jobs\DeleteDomains::class,
                 ])->send(function (Events\TenantDeleted $event) {
                     return $event->tenant;
                 })->shouldBeQueued(false),
--- a/assets/TenancyServiceProvider.stub.php
+++ b/assets/TenancyServiceProvider.stub.php
@@ -40,7 +40,13 @@ class TenancyServiceProvider extends ServiceProvider
             Events\TenantSaved::class => [],
             Events\UpdatingTenant::class => [],
             Events\TenantUpdated::class => [],
-            Events\DeletingTenant::class => [],
+            Events\DeletingTenant::class => [
+                JobPipeline::make([
+                    Jobs\DeleteDomain::class,
+                ])->send(function (Events\TenantDeleted $event) {
+                    return $event->tenant;
+                })->shouldBeQueued(false), // `false` by default, but you probably want to make this `true` for production.
+            ],
             Events\TenantDeleted::class => [
                 JobPipeline::make([
                     Jobs\DeleteDatabase::class,
--- a/assets/TenancyServiceProvider.stub.php
+++ b/assets/TenancyServiceProvider.stub.php
@@ -43,7 +43,7 @@ class TenancyServiceProvider extends ServiceProvider
             Events\DeletingTenant::class => [
                 JobPipeline::make([
                     Jobs\DeleteDomains::class,
-                ])->send(function (Events\TenantDeleted $event) {
+                ])->send(function (Events\DeletingTenant $event) {
                     return $event->tenant;
                 })->shouldBeQueued(false),
             ],
--- a/assets/impersonation-migrations/2020_05_15_000010_create_tenant_user_impersonation_tokens_table.php
+++ b/assets/impersonation-migrations/2020_05_15_000010_create_tenant_user_impersonation_tokens_table.php
@@ -23,7 +23,7 @@ class CreateTenantUserImpersonationTokensTable extends Migration
             $table->string('redirect_url');
             $table->timestamp('created_at');
 
-            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade');
+            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade')->onDelete('cascade');
         });
     }
 
--- a/assets/impersonation-migrations/2020_05_15_000010_create_tenant_user_impersonation_tokens_table.php
+++ b/assets/impersonation-migrations/2020_05_15_000010_create_tenant_user_impersonation_tokens_table.php
@@ -23,7 +23,7 @@ class CreateTenantUserImpersonationTokensTable extends Migration
             $table->string('redirect_url');
             $table->timestamp('created_at');
 
-            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade')->onDelete('cascade');
+            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade');
         });
     }
 
--- a/assets/migrations/2019_09_15_000020_create_domains_table.php
+++ b/assets/migrations/2019_09_15_000020_create_domains_table.php
@@ -21,7 +21,7 @@ class CreateDomainsTable extends Migration
             $table->string('tenant_id')->nullable();
 
             $table->timestamps();
-            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade')->onDelete('set null');;
+            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade');
         });
     }
 
--- a/assets/migrations/2019_09_15_000020_create_domains_table.php
+++ b/assets/migrations/2019_09_15_000020_create_domains_table.php
@@ -18,7 +18,7 @@ class CreateDomainsTable extends Migration
         Schema::create('domains', function (Blueprint $table) {
             $table->increments('id');
             $table->string('domain', 255)->unique();
-            $table->string('tenant_id');
+            $table->string('tenant_id')->nullable();
 
             $table->timestamps();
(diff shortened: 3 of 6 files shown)


assets/impersonation-migrations/2020_05_15_000010_create_tenant_user_impersonation_tokens_table.php
```diff
@@ -23,7 +23,7 @@ public function up(): void
             $table->string('redirect_url');
             $table->timestamp('created_at');
 
-            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade')->onDelete('cascade');
+            $table->foreign('tenant_id')->references('id')->on('tenants')->onUpdate('cascade');
```
stancl: Probably revert this change. We only care about logic for domains, not really about impersonation tokens. Those can disappear with no consequences.

stancl:
This should now introduce jobs/listeners (not sure, probably jobs, so that we can stay consistent with the JobPipelines used in the userland `TenancyServiceProvider`) that will delete the domains before a tenant is deleted.

So the job will be executed here: https://github.com/archtechx/tenancy/blob/db9480f54e98b33ca967bbe3f791c314e772386b/assets/TenancyServiceProvider.stub.php#L43

The job itself should be customizable somehow, but I'm thinking that we'll just let people extend it/replace it with their own job if needed. So the code inside the job will be very basic.

abrardev99:
> This should now introduce jobs/listeners

I've added the JobPipeline. It processes one job I said `DeleteDomain`, which deletes the domain.

assets/TenancyServiceProvider.stub.php
```diff
@@ -40,7 +40,13 @@ public function events()
             Events\TenantSaved::class => [],
             Events\UpdatingTenant::class => [],
             Events\TenantUpdated::class => [],
-            Events\DeletingTenant::class => [],
+            Events\DeletingTenant::class => [
+                JobPipeline::make([
+                    Jobs\DeleteDomain::class,
+                ])->send(function (Events\TenantDeleted $event) {
+                    return $event->tenant;
+                })->shouldBeQueued(false), // `false` by default, but you probably want to make this `true` for production.
```
stancl:
```suggestion
                })->shouldBeQueued(false),
```

assets/TenancyServiceProvider.stub.php
```diff
@@ -40,7 +40,13 @@ public function events()
             Events\TenantSaved::class => [],
             Events\UpdatingTenant::class => [],
             Events\TenantUpdated::class => [],
-            Events\DeletingTenant::class => [],
+            Events\DeletingTenant::class => [
+                JobPipeline::make([
+                    Jobs\DeleteDomain::class,
```
stancl: Rename this to:
```suggestion
                    Jobs\DeleteDomains::class,
```

src/Jobs/DeleteDomain.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function __construct(TenantWithDatabase $tenant)
+    {
+        $this->tenant = $tenant;
+    }
+
+    public function handle()
+    {
+        $this->tenant->domains->each(function (Domain $domain){
+            event(new DeletingDomain($domain));
```
stancl: I think these get fired by Eloquent automatically

abrardev99: I see.

stancl:
Not sure if we want the `set null` actually. You can probably keep it restricted. It will work because the domains get deleted *first* using the listeners.

abrardev99:
> because the domains get deleted first using the listeners.

Makes sense.

assets/migrations/2019_09_15_000020_create_domains_table.php
```diff
@@ -18,10 +18,10 @@ public function up(): void
         Schema::create('domains', function (Blueprint $table) {
             $table->increments('id');
             $table->string('domain', 255)->unique();
-            $table->string('tenant_id');
+            $table->string('tenant_id')->nullable();
```
stancl: Does the `nullable()` need to be here? Since you removed the `set null`

abrardev99: No, it does not. Forgot to remove it. Sorry for the inconvenience.

src/Jobs/DeleteDomains.php
```diff
@@ -0,0 +1,35 @@
...
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+use Stancl\Tenancy\Contracts\TenantWithDatabase;
+use Stancl\Tenancy\Database\Models\Domain;
+use Stancl\Tenancy\Events\DatabaseDeleted;
+use Stancl\Tenancy\Events\DeletingDatabase;
+use Stancl\Tenancy\Events\DeletingDomain;
+use Stancl\Tenancy\Events\DomainDeleted;
+
+class DeleteDomains implements ShouldQueue
```
stancl: One last change — why does this implement `ShouldQueue`? I don't think we want to queue the individual jobs in the pipeline.

abrardev99: > why does this implement ShouldQueue? I don't think we want to queue the individual jobs in the pipeline.

I mainly got motivation from the DeleteDatabase job. https://github.com/archtechx/tenancy/blob/3.x/src/Jobs/DeleteDatabase.php. DeleteDomains class is an exact copy of the DeleteDatabase class, and both classes are being used in the same way I mean in JobPipeline https://github.com/archtechx/tenancy/blob/3.x/assets/TenancyServiceProvider.stub.php#L45

stancl: I see. Database is deleted *after* the tenant is deleted. Domains are deleted *before*. So it needs to be synchronous.

I'm not sure how the JobPipeline interprets jobs having that interface though. Could you check that? Use that repo's readme and source code.

abrardev99: > I see. Database is deleted after the tenant is deleted. Domains are deleted before. So it needs to be synchronous.

Makes sense. 

> I'm not sure how the JobPipeline interprets jobs having that interface though. Could you check that? Use that repo's readme and source code.

After looking into the Pipeline source code, I think it does not matter if the Job class has `shouldQueue` interface or not. 

As JobPipeline class has boolean `shouldQueue` property, based on this property, [JobPipeline class dispatch job in a queue or synchronically. 
](https://github.com/archtechx/jobpipeline/blob/master/src/JobPipeline.php#L96)

I think having the `ShouldQueue` interface in `DeleteDomains` is fine because JobPipeline is deciding whether either job should be pushed to the queue or should execute synchronically. 

We set the `->shouldBeQueued(false)` so the processing will be synced. 

And the same is true for other Jobs. For example below code does not push job to the queue. I know because when I was learning the tenancy package, I did not run queue worker, and everything was working fine. 
```
	Events\TenantCreated::class => [
                JobPipeline::make([
                    Jobs\CreateDatabase::class,
                    Jobs\MigrateDatabase::class,
                    // Jobs\SeedDatabase::class,

                    // Your own jobs to prepare the tenant.
                    // Provision API keys, create S3 buckets, anything you want!

                ])->send(function (Events\TenantCreated $event) {
                    return $event->tenant;
                })->shouldBeQueued(false), // `false` by default, but you probably want to make this `true` for production.
            ],
```

stancl: > After looking into the Pipeline source code, I think it does not matter if the Job class has shouldQueue interface or not.

I see, you're right. Anyway, here we'll remove it because it's a `deleting` event listener 👍🏻

assets/TenancyServiceProvider.stub.php
```diff
@@ -40,7 +40,13 @@ public function events()
             Events\TenantSaved::class => [],
             Events\UpdatingTenant::class => [],
             Events\TenantUpdated::class => [],
-            Events\DeletingTenant::class => [],
+            Events\DeletingTenant::class => [
+                JobPipeline::make([
+                    Jobs\DeleteDomains::class,
+                ])->send(function (Events\TenantDeleted $event) {
+                    return $event->tenant;
```
stancl: Actually one more thing I noticed here — the event is `Events\DeletingTenant` and yet we're typehinting `Events\TenantDeleted`. Why is that?

abrardev99: > Why is that?

It's a mistake. It should be `DeletingTenant`.

src/Jobs/DeleteDomains.php
```diff
@@ -0,0 +1,35 @@
...
+    protected $tenant;
+
+    public function __construct(TenantWithDatabase $tenant)
+    {
+        $this->tenant = $tenant;
+    }
+
+    public function handle()
+    {
+        $this->tenant->domains()->delete();
```
stancl: Does this trigger the `deleting` event for those domain models?

abrardev99: > Does this trigger the deleting event for those domain models?

No. 
I think we need to get rid of bulk deletion and delete it using `domains->each->delete()` to make use of events.

stancl: Correct. The current approach misses the point of why we are doing this. Please read the linked issue/comment in depth.

src/Jobs/DeleteDomains.php
```diff
@@ -0,0 +1,35 @@
...
+    protected $tenant;
+
+    public function __construct(TenantWithDatabase $tenant)
+    {
+        $this->tenant = $tenant;
+    }
+
+    public function handle()
+    {
+        $this->tenant->domains->delete();
```
stancl: Does the collection even have `delete()`?

I'd just make this `each()` and make the code very obvious just from looking at it. `domains->delete()` is obscure. I checked the Eloquent Collection class. It has no `delete()`. If this works, it uses some Collection magic that I'm not sure if we even want here.

`each()` makes it very clear that the event will be triggered on every single model, which is what this code should communicate.

abrardev99: > Does the collection even have delete()?

No. 

> each() makes it very clear that the event will be triggered on every single model, which is what this code should communicate.

You are right. I even suggested using the `each->delete()` in the above review. I've updated it. Sorry for wasting your time.

abrardev99: Also added the test for the `DeleteDomains` job to make sure I'm not making any mistakes.

stancl:
Thanks, merged. Can you pull changes into the pest PR and update your test from this PR to use pest?

PR #884, opened 2022-06-28
[4.x] Migrate tests to Pest

stancl:
This pull request contains changes for migrating your test suite from PHPUnit to [Pest](https://pestphp.com/) automated by the [Pest Converter](https://laravelshift.com/phpunit-to-pest-converter).

**Before merging**, you need to:

- Checkout the `shift-64622` branch
- Review **all** of the comments below for additional changes
- Run `composer update` to install Pest with your dependencies
- Run `vendor/bin/pest` to verify the conversion


stancl:
:warning: Shift detected _helper methods_ in the following PHPUnit tests. Currently Pest does not support namespaces. As such, these functions may result in a naming conflict.

While Shift converted these methods to functions in the new Pest test, you should review these files to see if these functions may be inlined or extracted to a [custom helper](https://pestphp.com/docs/helpers).

- [ ] tests/AutomaticModeTest.php
- [ ] tests/BootstrapperTest.php
- [ ] tests/DatabasePreparationTest.php
- [ ] tests/EventListenerTest.php
- [ ] tests/QueueTest.php
- [ ] tests/ResourceSyncingTest.php
- [ ] tests/SingleDatabaseTenancyTest.php
- [ ] tests/TenantAssetTest.php
- [ ] tests/TenantDatabaseManagerTest.php
- [ ] tests/TenantModelTest.php
- [ ] tests/TenantUserImpersonationTest.php

stancl:
:information_source: Shift converted common PHPUnit assertions to [Pest expectations](https://pestphp.com/docs/expectations). However, there may also be opportunities to chain your expectations together.

For example:

```php
// Before
expect($comment)->toBeArray();
expect($comment)->toHaveCount(3);
expect($comment)->toContain('foo');

// After
expect($comment)->toBeArray()->toHaveCount(3)->toContain('foo');
```

stancl:
:warning: Shift detected references to the PHPUnit test runner (`vendor/bin/phpunit`) in the following files. You should review these instances to see if they may be updated to the Pest test runner (`vendor/bin/pest`).

- [ ] ./test

stancl:
:information_source: Pest is currently beta testing their [Parallel Plugin](https://pestphp.com/docs/plugins/parallel). If you would like to save time by running your tests in parallel, you may install this plugin and run your tests with the `-p` options.

```sh
composer require pestphp/pest-plugin-parallel --dev
vendor/bin/pest -p
```

stancl:
:alembic: This Shift is still being refined. **Please report any issues or suggestions** to [shift@laravelshift.com](mailto:shift@laravelshift.com). Your feedback is what helps improve the experience for everyone.

tests/AutomaticModeTest.php
```diff
@@ -2,134 +2,111 @@
...
-    public function bootstrap(\Stancl\Tenancy\Contracts\Tenant $tenant)
-    {
-        app()->instance('tenancy_initialized_for_tenant', $tenant->getTenantKey());
-    }
-
-    public function revert()
-    {
-        app()->instance('tenancy_ended', true);
-    }
+    app()->instance('tenancy_ended', true);
```
stancl: The original class should be added back in https://github.com/archtechx/tenancy/blob/3.x/tests/AutomaticModeTest.php#L124-L135

tests/BootstrapperTest.php
```diff
@@ -26,202 +22,187 @@
 use Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
 
-class BootstrapperTest extends TestCase
-{
-    public $mockConsoleOutput = false;
```
stancl, line 31: This might need to be added back in some way

stancl:
The conversion went pretty well, but some adjustments are needed.

First, we should resolve PHP issues that prevent the test suite from running at all. For example: https://github.com/archtechx/tenancy/runs/7101200032?check_suite_focus=true#step:5:8

The main issue seems to be helper methods that got converted to global functions. They should have unique names, or be refactored in some other way (e.g. if they're only used by a single test, it might be better to move the helper function to a Closure `$variable` inside the test, and run it `$likeThis()` instead of having a separate function).

Once that's fixed, the test suite should at least run. There will likely still be issues, but we'll see those listed out so we'll be able to fix them individually.

Right now I think even `./test --filter SomeTestName` wouldn't work due to the PHP autoloading issues, so those have to be addressed before we can move to anything else.

stancl:
@abrardev99 There are two open reviews here. If you resolved them, could you comment how? I believe you did add the class back, but I'm not sure what was done about the `mockConsoleOutput`.

stancl:
Also perhaps look into https://github.com/archtechx/tenancy/pull/884#issuecomment-1169234684

tests/BootstrapperTest.php
abrardev99: I totally removed the `$mockConsoleOutput`  property. I see we are not using this property in the whole class anywhere. So it was unused property affected nothing after removal.

tests/AutomaticModeTest.php
abrardev99: And in this case, I added the back `MyBootstrapper` class. https://github.com/archtechx/tenancy/blob/shift-64622/tests/AutomaticModeTest.php#L108-L119

abrardev99:
>  If you resolved them, could you comment how?

Both reviews are resolved code-wise. But I answered your questions in both reviews. [mockConsoleOutput](https://github.com/archtechx/tenancy/pull/884#discussion_r925573209) and [adding class back](https://github.com/archtechx/tenancy/pull/884#discussion_r925577064).

abrardev99:
> Also perhaps look into https://github.com/archtechx/tenancy/pull/884#issuecomment-1169234684

Yes, we can chain some expectations.

stancl:
There are some conflicts now, I think they might be related to the php-cs-fixer changes?

If so, you can just use the version in this PR for everything.

stancl:
(But do that **only** if the conflicts are due to CS, don't remove any real changes in `master`)

abrardev99:
> There are some conflicts now, I think they might be related to the php-cs-fixer changes?

Yeah, there are four conflicted files. All of the conflicts in them are related to namespace changes (possibly due to cs fixer), and one conflict is complex in the `TenantDatabaseManagerTest` class. Maybe some tests were added in that class after this PR.

stancl:
The imports are easy to resolve. Either properly accept/combine the changes or just go with what was used in **this** PR and it will be fixed by php-cs-fixer after merge. Also note that there are no "namespace changes" at all. It's just reordering imports.

For the complex file, just accept everything from this PR and ignore all changes in `master`.
CleanShot 2022-07-20 at 4 06 05@2x

abrardev99:
Opened a PR #906 from this PR and resolved conflicts in the #906.

stancl:
And include the conversion of the new tests here as well, we don't want two PRs, we want one PR.

abrardev99:
> And include the conversion of the new tests here as well, we don't want two PRs, we want one PR.

Yeah, I just did.

tests/BootstrapperTest.php
stancl: Hmm I'm thinking why the property was there. It is possible that tests are passing, but that the property needed to be set for some tests to work properly?

For example, some tests may be passing without actually making the necessary assertions now. So if the logic that they're testing broke, they'd still pass due to the property not being set.

stancl:
I'll merge this now, but could you look into https://github.com/archtechx/tenancy/pull/884#discussion_r927856784 on Monday? I'd like to make sure that we didn't lose any assurance by removing the property.

stancl:
And if there's anything else unresolved, let me know.

tests/BootstrapperTest.php
abrardev99: >  but that the property needed to be set for some tests to work properly?

I've checked the whole class. Firstly, `$mockConsoleOutput` is never being used (because we are not calling any command on those tests). 
Secondly, We just have one test [database_data_is_separated()](https://github.com/abrardev99/tenancy/blob/3.x/tests/BootstrapperTest.php#L49) where we are calling the `->artisan` command. And this test logic is pretty clear and looks good. I think everything is fine.

abrardev99:
> I'll merge this now, but could you look into https://github.com/archtechx/tenancy/pull/884#discussion_r927856784 on Monday? I'd like to make sure that we didn't lose any assurance by removing the property.

I replied on the review https://github.com/archtechx/tenancy/pull/884#discussion_r928656895

abrardev99:
> And if there's anything else unresolved, let me know.

There is not anything else. We are good here.

tests/BootstrapperTest.php
stancl: `$mockConsoleOutput` was set in `BootstrapperTest` and `QueueTest`. Artisan is used in both.

> because we are not calling any command on those tests

https://github.com/archtechx/tenancy/blob/600bb823de185ef02c3ff5aeba0c75f41409bafa/tests/BootstrapperTest.php#L58

abrardev99: I rechecked both classes. I don't see any usage or possible usage of the property.

stancl: > I don't see any usage or possible usage of the property.

You're missing the point. It affects how artisan work. I showed you that artisan is called in those tests.

It's not about the property being used by us. It's about the property changing behavior of Laravel during the tests.

abrardev99: Resolving this review in the favour #907.

PR #889, opened 2022-07-04
[4.x] Add feature that allow to determine when model is syncable (after #915)

leandrogehlen:
Sometimes is necessary to ignore resource synchronization.
In my case, i have `contacts` in tenant tables and these contacts can be or not an user. This way, i want to synchronize with central only contacts that contains `email` attribute defined.

Using the example bellow:

```php
// Contact represents a tenant user
class Contact extends Authenticatable implements Syncable
{
    use HasFactory,       
        ResourceSyncing;

    /**
     * The attributes that are mass assignable.
     *
     * @var array<int, string>
     */
    protected $fillable = [
        'name',
        'email',
        'password'
    ];

    /**
     * The attributes that should be hidden for arrays.
     *
     * @var array<int, string>
     */
    protected $hidden = [
        'password',
        'api_token'
    ];  

    /**
     * The global identifier key value
     *
     * @return string
     */
    public function getGlobalIdentifierKey()
    {
    }

    /**
     * The global identifier key name
     *
     * @return string
     */
    {
        return 'email';
    }

    /**
     * The central model class
     *
     * @return string
     */
    {
        return CentralUser::class;
    }

    /**
     * The synced attribute names
     *
     * @return array
     */
    {
        return [
            'name',
            'password',
        ];
    }

    /**
     * Synchronize only contacts that have email attribute defined
     *
     * @return bool
     */
    public function isSyncEnabled(): bool
    {
        return $this->email !== null;
    }
 ```
 
 ```php
 class CentralUser extends Authenticatable implements MustVerifyEmail, SyncMaster
{
    use HasApiTokens,
        HasFactory,
        Notifiable,
        ResourceSyncing,
        CentralConnection;

    /**
     * The attributes that are mass assignable.
     *
     * @var array<int, string>
     */
    protected $fillable = [
        'name',
        'email',
        'password',
    ];

    /**
     * The attributes that should be hidden for serialization.
     *
     * @var array<int, string>
     */
    protected $hidden = [
        'password',
        'remember_token',
    ];

    /**
     * The attributes that should be cast.
     *
     * @var array<string, string>
     */
    protected $casts = [
        'email_verified_at' => 'datetime',
    ];

    /**
     * The users relationship definition
     *
     * @return BelongsToMany
     */
    {
        return $this->belongsToMany(Tenant::class, 'tenant_users', 'email', 'tenant_id', 'email')
            ->using(TenantPivot::class);
    }


    /**
     * The global identifier key value
     *
     * @return string
     */
    public function getGlobalIdentifierKey()
    {
    }

    /**
     * The global identifier key name
     *
     * @return string
     */
    {
        return 'email';
    }

    /**
     * The central model class
     *
     * @return string
     */
    {
        return static::class;
    }

    /**
     * The tenant model class
     *
     * @return string
     */
    {
        return Contact::class;
    }

    /**
     * The synced attribute names
     *
     * @return array
     */
    {
        return [
            'name',
            'password'
        ];
    }
}
```
Using:

```php

$user = CentralUser::create([
    'name' => 'Steve',
    'email' => 'steve@test.com'
]);

$user = CentralUser::create([
    'name' => 'Bill',
    'email' => 'bill@test.com'
]);

// it will not be synced
$contact = Contact::create([
     'name' => 'Steve Jobs'
]);

//it will be synced
$contact = Contact::create([
      'name' => 'Bill Gates',
      'email' => 'bill@test.com'
]);

```
The same logic can be applyed with central user.
With this PR, is possible to determine a "kind" of `users` (in my case `contacts`) must be synced

### DIFF
--- a/src/Database/Concerns/ResourceSyncing.php
+++ b/src/Database/Concerns/ResourceSyncing.php
@@ -20,10 +20,10 @@ trait ResourceSyncing
         });
 
         static::creating(function (self $model) {
-            $key = $model->getGlobalIdentifierKeyName();
+            $keyName = $model->getGlobalIdentifierKeyName();
 
-            if (! $model->getAttribute($key) && app()->bound(UniqueIdentifierGenerator::class) && $model->isSyncEnabled()) {
-                $model->setAttribute($key, app(UniqueIdentifierGenerator::class)->generate($model));
+            if (! $model->getAttribute($keyName) && app()->bound(UniqueIdentifierGenerator::class)) {
+                $model->setAttribute($keyName, app(UniqueIdentifierGenerator::class)->generate($model));
             }
         });
     }
--- a/src/Database/Concerns/ResourceSyncing.php
+++ b/src/Database/Concerns/ResourceSyncing.php
@@ -14,15 +14,16 @@ trait ResourceSyncing
     {
         static::saved(function (Syncable $model) {
             /** @var ResourceSyncing $model */
-            $model->triggerSyncEvent();
+            if ($model->isSyncEnabled()) {
+                $model->triggerSyncEvent();
+            }
         });
 
         static::creating(function (self $model) {
-            if (! $model->getAttribute($model->getGlobalIdentifierKeyName()) && app()->bound(UniqueIdentifierGenerator::class)) {
-                $model->setAttribute(
-                    $model->getGlobalIdentifierKeyName(),
-                    app(UniqueIdentifierGenerator::class)->generate($model)
-                );
+            $key = $model->getGlobalIdentifierKeyName();
+
+            if (! $model->getAttribute($key) && app()->bound(UniqueIdentifierGenerator::class) && $model->isSyncEnabled()) {
+                $model->setAttribute($key, app(UniqueIdentifierGenerator::class)->generate($model));
             }
         });
     }
@@ -31,5 +32,10 @@ trait ResourceSyncing
     {
         /** @var Syncable $this */
         event(new SyncedResourceSaved($this, tenant()));
+    }
+
+    public function isSyncEnabled()
+    {
+        return true;
     }
 }
--- a/src/Contracts/Syncable.php
+++ b/src/Contracts/Syncable.php
@@ -15,4 +15,6 @@ interface Syncable
     public function getSyncedAttributeNames(): array;
 
     public function triggerSyncEvent();
+
+    public function isSyncEnabled();
 }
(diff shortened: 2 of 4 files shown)


leandrogehlen:
The tests are passing.
It is the only change.

src/Database/Concerns/ResourceSyncing.php
```diff
@@ -14,15 +14,16 @@ public static function bootResourceSyncing()
...
         static::creating(function (self $model) {
-            if (! $model->getAttribute($model->getGlobalIdentifierKeyName()) && app()->bound(UniqueIdentifierGenerator::class)) {
-                $model->setAttribute(
-                    $model->getGlobalIdentifierKeyName(),
-                    app(UniqueIdentifierGenerator::class)->generate($model)
-                );
+            $key = $model->getGlobalIdentifierKeyName();
+
+            if (! $model->getAttribute($key) && app()->bound(UniqueIdentifierGenerator::class) && $model->isSyncEnabled()) {
+                $model->setAttribute($key, app(UniqueIdentifierGenerator::class)->generate($model));
```
stancl: I can understand the `isSyncEnabled()` change, but what exactly are we doing here? What's the point of the change?

leandrogehlen: I'm sorry, i did not pay attention.
Does not make sense, sync is enabled to set key attribute.
This must be removed.

stancl: So if we were to re-work this PR, we'd just add the `isSyncEnabled()` method?

leandrogehlen: I have removed ```$model->isSyncEnabeld()``` check.

This:
```php
 if (! $model->getAttribute($key) && app()->bound(UniqueIdentifierGenerator::class) && $model->isSyncEnabled()) {
```
should be: 
```php
 if (! $model->getAttribute($key) && app()->bound(UniqueIdentifierGenerator::class)) {
```

stancl: The code is still changed, so I'm wondering what exactly is the point of the change in this section?

leandrogehlen: It is just a refactoring to turn code more clean. Instead of invoke ` $model->getGlobalIdentifierKeyName()` twice i stored the value in $keyName

stancl: I see, so there are no logical changes, it's just a refactor?

leandrogehlen: Exacly

stancl:
Implemented in #993

Issue #892, opened 2022-07-09
Universal routes doesn't work on RequestData based initialization.


l4nos:
### Bug description

I was trying to enable passport for admin logins (central connection), this requires universal routes, i've enabled as per the guide.

However as we are using SPA based approach, our API has only one domain (api.domain.com) and we are using InitializeTenancyByRequestData::class to identify tenants.

This isn't supported by Universal routes so when admin's log in, they receive:

`{
    "Error": "The tenant could not be identified, please check the domain and try again.",
    "Exception": "Tenant could not be identified by request data with payload: "
}`

Now this can be easilly fixed by adding the middleware class into Stancl\Tenancy\Features\UniversalRoutes.

![image](https://user-images.githubusercontent.com/31861108/178120820-437d1f70-0366-4bb7-8e39-4ef3c8112ca3.png)

The above image shows my proposed fix, as you can see i've appended to the $identificationMiddlewares to add the middleware we are using to identify tenants. When this is added in, everything works as it should.

Can we inplement this fix ASAP, or would you accept a pull request for this?

### Steps to reproduce

Using InitializeTenancyByRequestData::class try to hit a universal route without specifying a tenant.

### Expected behavior

Should just allow the user to proceed on central connection.

### Laravel version

9

### stancl/tenancy version

3.1

l4nos:
Submitted a PR, if you can review at earliest convenience.

PR #893, opened 2022-07-09
A fix for the issue #892

l4nos:
https://github.com/archtechx/tenancy/issues/892

Described the isuse there in detail, please approve PR if you feel it's a suitable fix. I've tested it my side and everything works as intended.

### DIFF
--- a/src/Features/UniversalRoutes.php
+++ b/src/Features/UniversalRoutes.php
@@ -18,6 +18,7 @@ class UniversalRoutes implements Feature
     public static $identificationMiddlewares = [
         Middleware\InitializeTenancyByDomain::class,
         Middleware\InitializeTenancyBySubdomain::class,
+        Middleware\InitializeTenancyByRequestData::class
     ];
 
     public function bootstrap(Tenancy $tenancy): void


stancl:
Hey!

So as I wrote in the email, this feature isn't really _intended_ to work with request data. It may still work (and if it's needed for Passport to work, definitely use it), but it's not intended so we don't want it enabled by default.

As for the change in this PR — you can achieve the same without modifying the package code. Add this to the `register()` method of `AppServiceProvider`/`TenancyServiceProvider`:
```php
UniversalRoutes::$identificationMiddleware = [
    InitializeTenancyByDomain::class,
    InitializeTenancyBySubdomain::class,
    InitializeTenancyByRequestData::class
];
```

(And import all of those classes)

syn6UK:
This wasn't Possible.

`
	<title>Access to undeclared static property Stancl\Tenancy\Features\UniversalRoutes::$identificationMiddleware (500
		Internal Server Error)</title>
`

l4nos:
This worked, but you missed the "s" pural out on identificationMiddleware. it's actually identificationMiddlewares.

PR #894, opened 2022-07-12
Fix ArgumentCountError on the TenantAssetsController

megawubs:
When no `$path` is provided in the `tenant_asset()`, the link that is generated does not contain the `$path` part, but in the route registration {path?} is optional. This results in an `ArgumentCountError`. 

By making it default `null` it now returns an `404` as expected.

### DIFF
--- a/src/Controllers/TenantAssetsController.php
+++ b/src/Controllers/TenantAssetsController.php
@@ -15,7 +15,7 @@ class TenantAssetsController extends Controller
         $this->middleware(static::$tenancyMiddleware);
     }
 
-    public function asset($path)
+    public function asset($path = null)
     {
         try {
             return response()->file(storage_path("app/public/$path"));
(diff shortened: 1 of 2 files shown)


stancl:
> When no $path is provided in the tenant_asset()

What's the use case for this? Passing variables that may be empty to `tenant_asset()`?

megawubs:
> > When no $path is provided in the tenant_asset()
> 
> What's the use case for this? Passing variables that may be empty to `tenant_asset()`?

Yes, exactly that.

stancl:
I think the `null` should be explicitly handled in the method, we don't call the rest of the code when it's null.

Add an `abort_if($path === null, 404)` (not sure about arg order) above the `try {}` block

megawubs:
Done! ✅

stancl:
Thanks for the contribution!

PR #902, opened 2022-07-20
php-cs-fixer (Completing PR #881)

abrardev99:
Added the `php_cs_fixer` config and CI according to the template.


abrardev99:
This PR forked from #881. Merging this PR should also close the #881. Thanks, @erikgaal for your contributions.

stancl:
Did you make any changes to the workflow from the template?

abrardev99:
> Did you make any changes to the workflow from the template?

Yes, I just added the php-cs-fixer job like https://github.com/archtechx/template/blob/master/.github/workflows/ci.yml#L42-L56.

Looks like php-cs-fixer is failing because my branch is forked from another forked PR branch. Same tests are passing in my local forked repo PR https://github.com/abrardev99/tenancy/pull/3. Maybe we should merge it as is. And I believe the php-cs-fixer check will pass in #881.

stancl:
> Yes, I just added the php-cs-fixer job like

It seems different — I was asking **what changes you made**

stancl:
I mean changes compared to the template. For example, the email is different. So I'm asking why, and what other changes you made.

abrardev99:
> I mean changes compared to the template. For example, the email is different. So I'm asking why, and what other changes you made.

I just changed the email. The rest of the workflow is the same as in the template.

stancl:
Make it the same as it is in the template.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 env:
   COMPOSE_INTERACTIVE_NO_CLI: 1
+  PHP_CS_FIXER_IGNORE_ENV: 1
```
stancl, line 5: I think this can be removed now

stancl: Try it in your branch and then we can update the template and this repo

abrardev99: Yes, checks are passing after removing this. I'll update the template.

stancl: Both this repo and the template

abrardev99: Got it.

abrardev99: PRs sent to both repos. 

https://github.com/archtechx/tenancy/pull/904
https://github.com/archtechx/template/pull/4

PR #905, opened 2022-07-21
Convert newly added tests to pests 

abrardev99:
- New tests added in the [[4.x] Don't use onDeleteCascade in the migrations](https://github.com/archtechx/tenancy/pull/883#issuecomment-1190135781) PR. 

- https://github.com/archtechx/tenancy/pull/884#issuecomment-1169234684 as for this review, I couldn't find many places to change Pest expectations to chains.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master, shift-64622] # todo0 remove shift-64622 branch before merge
+    branches: [ master ]
 
 jobs:
   tests:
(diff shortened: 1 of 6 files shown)


abrardev99:
Looks like all [checks are passing](https://github.com/archtechx/tenancy/runs/7444987315?check_suite_focus=true). So, I removed the `shift-64622` branch from CI.

abrardev99:
This PR base branch should be changed to #906 if we are going to merge #906 PR instead of #884

stancl:
Can't you push into `shift-64622` directly? I don't like creating sub-PRs like this

PR #689, opened 2021-07-19
[4.x] This adds support for tenancy aware Storage::url() method

dontfreakout:
This fixes #78 and #201 and enables using `Storage::url('file.jpg')` with local driver in tenants (see [Laravel docs](https://laravel.com/docs/8.x/filesystem#file-urls)).

With this, you can use for example Spatie media library and local disk without any modifications.
You only need to create symlinks in the public directory which points to the tenants' disks root.

For creating symlinks you can use the new command `tenants:link`

PR #909, opened 2022-07-27
[4.x] Storage::url() support (modified #689)

stancl:
We'll either merge this PR or #689 (after force pushing these changes into that branch), but for wrapping up the changes we'll be using this PR.


stancl:
@lukinovec You can do a review of all of the added code (similar to how you're wrapping up the pending tenants feature). And I'll dive into the complex low level changes.

src/Jobs/CreateStorageSymlinks.php
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        event(new CreatingStorageSymlink($this->tenant));
+
+        Artisan::call('tenants:link', [
+            '--tenants' => [$this->tenant->getTenantKey()],
+        ]);
```
stancl: @lukinovec Do you think the structure should be a job calling an artisan command, or the opposite?

The opposite seems more natural to me.

Or perhaps we could create an action class to extract the logic?

Let me know which solution makes the most sense to you.

src/Jobs/RemoveStorageSymlinks.php
```diff
@@ -0,0 +1,49 @@
...
+    public function handle()
+    {
+        event(new RemovingStorageSymlink($this->tenant));
+
+        Artisan::call('tenants:link', [
+            '--remove' => true,
+            '--tenants' => [$this->tenant->getTenantKey()],
+        ]);
+
+        event(new StorageSymlinkRemoved($this->tenant));
```
stancl: Similar to the review in `CreateStorageSymlinks`. Also firing events *only* in the job and not the command seems weird. The events should be fired as part of the logic, it should be a single atomic thing.

tests/SingleDatabaseTenancyTest.php
```diff
@@ -61,7 +61,7 @@
     expect(Comment::count())->toBe(2);
 });
 
-test('secondary models a r e scoped to the current tenant when accessed directly and parent relationship traitis used', function () {
+test('secondary models ARE scoped to the current tenant when accessed directly and parent relationship traits used', function () {
```
stancl: `traitis` perhaps should be `trait is`, rather than `trait`? @abrardev99

src/Commands/Link.php
```diff
@@ -0,0 +1,126 @@
...
+        ->all();
+    }
+
+    /**
+     * Determine if the provided path is a removable symlink.
+     */
+    protected function isRemovableSymlink(string $link, bool $force): bool
+    {
+        return is_link($link) && $force;
+    }
```
stancl: It should be better described what `$force` is here.

src/Commands/Link.php
```diff
@@ -0,0 +1,126 @@
...
+        $suffixBase = config('tenancy.filesystem.suffix_base');
+
+        $tenants = $this->option('remove') && filled($this->option('tenants'))
+            ? collect($this->option('tenants'))
+            : $this->getTenants()->map(function (Tenant $tenant) {
+                return $tenant->getTenantKey();
+            });
+
+        return $tenants->map(function ($tenantKey) use ($suffixBase, $diskUrls, $disks) {
+            $symLinks = [];
```
stancl: Should be lowercase, we use `symlinks` everywhere else

assets/config.php
```diff
@@ -118,6 +118,16 @@
...
 
+        /*
+         * This makes local driver disks support tenant-aware Storage::disk()->url().
+         * Use `php artisan tenants:link` to create a symbolic link from the tenant's storage to his public directory.
+         */
+        'url_override' => [
+            // 'local disk' => 'public directory' (%tenant_id% will be replaced with actual tenant id)
+            // The local disk must exist in the tenancy.filesystem.root_override config
+            'public' => 'public-%tenant_id%',
+        ],
```
stancl: I think this section should be a bit better explained. For the other keys like this we have some examples I believe

tests/BootstrapperTest.php
```diff
@@ -184,6 +187,70 @@
...
+    $this->assertEquals(
+        'http://localhost/public-'.$tenant1->getTenantKey().'/',
+        Storage::disk('public')->url('')
+    );
+
+    tenancy()->initialize($tenant2);
+    $this->assertEquals(
+        'http://localhost/public-'.$tenant2->getTenantKey().'/',
+        Storage::disk('public')->url('')
+    );
```
stancl: I think it'd be better if the tests actually tried to save some text file and then asserted that it can be fetched on the URL. Now we're only making assertions about the root URL, but we want to test that the feature actually works for files.

So **on top** of the root assertions, there should be assertions about files.

You can make that a separate test if you want. Or feel free to reorganize them in any way.

src/Commands/Link.php
```diff
@@ -0,0 +1,126 @@
...
+     * @var string
+     */
+    protected $description = 'Create symbolic links for tenants.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
```
stancl, line 41: This method, as well as `links()` is pretty hard to read. @lukinovec do you think you could refactor it, perhaps decrease the complexity of the nesting, extract things into methods, add variables or comments? You don't have to do exactly those things of course, but something along those lines to improve readability. The code is very low level and crammed, so it's very hard to get an idea of what's happening here without actually reading the class for several minutes.

tests/SingleDatabaseTenancyTest.php
lukinovec: Right, that makes more sense

abrardev99: Yeah it is `"trait is"`.

src/Jobs/CreateStorageSymlinks.php
lukinovec: Yeah, the opposite seems more natural. But the command has a lot of output, which makes extracting the logic more difficult. How would you go about the output?

src/Commands/Link.php
lukinovec: The `tenant:link` command does very different things depending on the `--remove` option. If it's passed, instead of creating the symlinks, the existing ones are removed.
```php
public function handle()
{
    if ($this->option('remove')) {
        $this->removeLinks();

        return;
    }

    $this->createLinks();
}
```
Sounds like these two actions should have separate commands.

lukinovec:
```php
$tenants = $this->option('remove') && filled($this->option('tenants'))
            ? collect($this->option('tenants'))
            : $this->getTenants()->map(function (Tenant $tenant) {
                return $tenant->getTenantKey();
            });
```

Using the tenants from the option only when removing the links doesn't make much sense to me. If we specify `tenants` in the option, the command should work with these. Something like this should be sufficient:
```php
$tenants = $this->option('tenants') ?? $this->getTenants()->map->getTenantKey()
```

src/Commands/Link.php
lukinovec: I'd rename `force` to `overwrite`.

Looks like passing the `force` option instructs the command to overwrite the links if they exists. This method shouldn't accept `$force`. We should check if the user passed the overwrite/force option AND if the link is removable.

That would make the code that uses the method look lke this:
```php
if (! $this->option('force') && ! $this->isRemovableSymlink($link)) {
    $this->error("The [$link] link already exists.");
    continue;
}
```

src/Commands/Link.php
lukinovec: I refactored the command https://github.com/archtechx/tenancy/pull/909/commits/a45f02445cfd4e21268ec68b5a8c6b075aa8be1f (separated creating, deleting and getting the symlinks)

tests/SingleDatabaseTenancyTest.php
lukinovec: Fixed it

src/Jobs/CreateStorageSymlinks.php
stancl: Seems like the nature of those messages is either:
- success message
- error message

Is that correct?

We could handle stuff like the `--remove` in the **command**, based on that execute the right **action/job**, and show the right success message. Errors could be communicated via exceptions.

lukinovec: When **a** link gets removed, we get a success message, and after all links get removed, we get the final success message. Same goes for creating the links. So if 5 links are created, 6 success messages get displayed

> We could handle stuff like the --remove in the command, based on that execute the right action/job, and show the right success message. Errors could be communicated via exceptions.

I don't think that would work with the multiple success messages I described

tests/BootstrapperTest.php
lukinovec: I tried this, and fetching the URL gives me a 404

src/Jobs/CreateStorageSymlinks.php
lukinovec: > We could handle stuff like the --remove in the command, based on that execute the right action/job, and show the right success message. Errors could be communicated via exceptions.

https://github.com/archtechx/tenancy/pull/909/commits/2594b1960fbf663b02292c9740e11977ba313f37

I handled this the way you described with an addition to make multiple success messages work (passing a closure that outputs a message if the removal/creation was successful for each link).

@stancl, could you please review the file structure? I wasn't sure where to put the actions and the trait.

assets/config.php
lukinovec: Updated the comments for this config key https://github.com/archtechx/tenancy/pull/909/commits/f2d562cd8b81255877d14ecb92f351e9c78d8b4f

src/Jobs/CreateStorageSymlinks.php
stancl: Done

tests/BootstrapperTest.php
lukinovec: @stancl, [I changed the test](https://github.com/archtechx/tenancy/compare/b474ae219c844cde536a5ed0d052fc0b59e5cfaf..storage-url-conflict-resolution) according to our BC discussion

stancl:
After https://github.com/archtechx/tenancy/commit/f2d562cd8b81255877d14ecb92f351e9c78d8b4f#r82301420 is done, let me know if I can do a final review and merge this.

lukinovec:
> After [f2d562c#r82301420](https://github.com/archtechx/tenancy/commit/f2d562cd8b81255877d14ecb92f351e9c78d8b4f#r82301420) is done, let me know if I can do a final review and merge this.

Done👍

src/Commands/Link.php
```diff
@@ -0,0 +1,75 @@
...
+        try {
+            if ($this->option('remove')) {
+                $this->removeLinks($tenants);
+            } else {
+                $this->createLinks($tenants);
+            }
+        } catch (Exception $exception) {
+            $this->error($exception->getMessage());
+
+            return 1;
```
stancl: Can we return from this method? Are the returned values used?

There's a `@return void` annotation on this method. So probably check the docs or source dive to see what's actually correct.

src/Concerns/DealsWithTenantSymlinks.php
```diff
@@ -0,0 +1,43 @@
...
+
+            $symlinks->push([public_path($publicPath) => storage_path($storagePath)]);
+        }
+
+        return $symlinks->mapWithKeys(fn ($item) => $item);
+    }
+
+    /**
+     * Determine if the provided path is an existing symlink.
+     */
```
stancl:
```suggestion
    /** Determine if the provided path is an existing symlink. */
```

FYI we use this style for docblocks that don't have any params or return types.

stancl: Keeping this unresolved so that you see it @lukinovec. You can resolve this, I already committed the change.

stancl:
Moving this convo here since discussions on commits are messy (not sure how the reviews got applied to the commit instead of the PR) https://github.com/archtechx/tenancy/commit/2594b1960fbf663b02292c9740e11977ba313f37#r82107711

Do our tests check that the jobs work well for this? As in, the tests should set up the event environment to be the same way as what the developer would do in the `TenancyServiceProvider`.

src/Commands/Link.php
lukinovec: > Can we return from this method? Are the returned values used?

I see the return value isn't used anywhere. I added this return value so we can test if the command failed using `->assertFailed()`. Now, the link command calls aren't tested. There are tests only for the jobs Create/RemoveSymlinks jobs. I think I should add these tests back, but I'd remove the `return 1` here and instead assert something else (e.g.  outputting an error message).

Also, the underlying actions for the jobs aren't separately tested because the jobs just do `XStorageSymlinksAction::handle($tenant)`, and the tests would look almost the same. It would probably be best to have them tested either way.

@stancl, what do you think?

lukinovec:
> Moving this convo here since discussions on commits are messy (not sure how the reviews got applied to the commit instead of the PR) [2594b19#r82107711](https://github.com/archtechx/tenancy/commit/2594b1960fbf663b02292c9740e11977ba313f37#r82107711)
> 
> Do our tests check that the jobs work well for this? As in, the tests should set up the event environment to be the same way as what the developer would do in the `TenancyServiceProvider`.

[At the beginning of the test](https://github.com/archtechx/tenancy/pull/909/files#diff-b22931f4c76926f01513d299f2f87fb1528e0d3a92e539ef66cb4675b4342f4bR274-R296), we're adding the listeners and setting the config values, so I think it is tested properly.

stancl:
Ah I see, that's fine then.

src/Commands/Link.php
stancl: > Now, the link command calls aren't tested

Why is that?

> I think I should add these tests back

Were they removed as part of some change?

lukinovec: I removed them during a merge: https://github.com/archtechx/tenancy/pull/909/commits/8aff44215d920bed5adff810f6fe8836fbef9ad9#diff-2d46e8828aea0422777576ca8e903262bf80a262978ec02ad3d0783c38fe1b32L210-L215

I'm not sure but I think I removed them because I thought the job tests would be enough

stancl: I think testing everything is best, if there's repetition I can show you how that can be fixed with [datasets](https://pestphp.com/docs/datasets) after I see the tests.

src/Concerns/DealsWithTenantSymlinks.php
```diff
@@ -27,7 +27,9 @@ protected static function possibleTenantSymlinks(Tenant $tenant): Collection
             $storagePath = str_replace('%storage_path%', $suffixBase . $tenantKey, $disks[$disk]);
             $publicPath = str_replace('%tenant_id%', $tenantKey, $publicPath);
 
-            $symlinks->push([public_path($publicPath) => storage_path($storagePath)]);
+            tenancy()->central(function() use ($symlinks, $publicPath, $storagePath) {
+                $symlinks->push([public_path($publicPath) => storage_path($storagePath)]);
+            });
```
lukinovec: Reason: The paths are prefixed/suffixed manually (L27-28) because this action is intended to be used in the central context.  But when running the action as a tenant (which is possible at the moment), the path gets prefixed twice.

@stancl, should we keep it like this, or should we enforce not using the symlink actions in the tenant context? I think that since the actions accept collections of tenants, using the actions as a tenant doesn't make much sense – the tenant would be technically able to manage the symlinks of other tenants. But how to enforce the need to be in the central context to run these actions? Do we just check if tenancy is initialized, and if it is, throw an exception?

src/Commands/Link.php
lukinovec: I wrote the tests (https://github.com/archtechx/tenancy/compare/044e6441ea89a607a977cfc5d1835447662a2f61..storage-url-conflict-resolution#diff-c3d921c826566fca3c70dd75bb435239958d1236902d400cabf6aaaa0e034712R1). There's [a thing left to solve](https://github.com/archtechx/tenancy/pull/909#discussion_r959227902) though EDIT: the thing is resolved

src/Concerns/DealsWithTenantSymlinks.php
lukinovec: Resolving this in response to Samuel's reply on BC

tests/BootstrapperTest.php
```diff
@@ -184,6 +191,125 @@
...
+    config([
+        'tenancy.bootstrappers' => [
+            FilesystemTenancyBootstrapper::class,
+        ],
+        'tenancy.filesystem.suffix_base' => 'tenant-',
+        'tenancy.filesystem.root_override.public' => '%storage_path%/app/public/',
+        'tenancy.filesystem.url_override.public' => 'public-%tenant_id%'
+    ]);
+
+    /** @var \Stancl\Tenancy\Database\Models\Tenant $tenant */
```
stancl: If `Tenant::create()` works, do we need to specify the full namespace in the `@var`?

lukinovec: I see, no, there's no need for that. I updated this now

stancl: I think there were multiple instances of this, did you change all of them?

lukinovec: You're right, I forgot about ActionTest

stancl:
@lukinovec Seems like there are conflicts now

stancl:
See the failing tests. Seems like the conflicts didn't get resolved well.

lukinovec:
> See the failing tests. Seems like the conflicts didn't get resolved well.

It should be OK now

src/Actions/RemoveStorageSymlinksAction.php
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+    }
+
+    protected static function removeLink(string $publicPath, Tenant $tenant): void
+    {
+        if (static::symlinkExists($publicPath)) {
+            event(new RemovingStorageSymlink($tenant));
+
+            app()->make('files')->delete($publicPath); // todo change this to use DI
```
stancl: Do we plan on changing the class to be non-static?

stancl:
I think we discussed this before, but I want to bring it up once again. Do we need the jobs which just redirect the call to the action? For instance `RemoveStorageSymlinks`. Where exactly is it used, the JobPipeline? Couldn't we simply pass the action to the JobPipeline? It also uses `handle()`.

Or do we need the job class because it's an object that includes the tenant and can get serialized for the queue?

src/Concerns/DealsWithTenantSymlinks.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Stancl\Tenancy\Concerns;
+
+use Illuminate\Support\Collection;
+use Stancl\Tenancy\Database\Models\Tenant;
+
+trait DealsWithTenantSymlinks
+{
+    /**
+     * Get all possible tenant symlinks, existing or not (array of ['public path' => 'storage path']).
+     * This is used for creating all possible tenant symlinks and removing all existing tenant symlinks.
```
stancl: I think this should explain a bit better why there are "possible symlinks" i.e. why there are multiple possible ones. I think it relates to the config being an array that can have multiple values, and therefore multiple symlinks?

Still, there should be one more sentence here explaining that.

src/Concerns/DealsWithTenantSymlinks.php
```diff
@@ -0,0 +1,43 @@
...
+     *
+     * @return array
```
stancl: This returns a `Collection`, and the typehint exists, so you can remove the `@return` annotation.

src/Actions/RemoveStorageSymlinksAction.php
stancl: I think we don't. So even if `app()->make('files')` isn't ideal, we probably won't ever change it? Therefore the `todo` could be removed.

lukinovec:
> Where exactly is it used, the JobPipeline? Couldn't we simply pass the action to the JobPipeline? It also uses `handle()`.

Yeah, the job is used only in the JobPipeline. I made the jobs because the actions alone miss the functionality of jobs (because it's a **Job**Pipeline, I'd expect only jobs to be passed). If you think that is not needed, we could pass the action to the JobPipeline as you're saying, but we'd have to make the actions have constructors so that the pipeline can pass the tenant to the action (so it's just that we'll have to instantiate the actions, which I don't think is a big problem)

EDIT: I made the changes https://github.com/archtechx/tenancy/pull/909/commits/547440c887dd86d75c7a5543fec576e233487eff

src/Concerns/DealsWithTenantSymlinks.php
lukinovec: If there are two disks in the url_override config array, the tenant has two possible symlinks. So the first line of the comment should say something like:

"Get tenant symlinks (existing or not) for all disks registered in the `tenancy.filesystem.url_override` config"

Does that make it a bit more clear?

stancl:
If you converted the actions to be stateful, then that reverts what we discussed here: https://github.com/archtechx/tenancy/pull/909#discussion_r978234944

src/Concerns/DealsWithTenantSymlinks.php
stancl: I would *add another paragraph*, not replace the "Get all possible tenant symlinks" since that's correct and descriptive.

It should simply be clear why there can be multiple potential symlinks.

src/Actions/RemoveStorageSymlinksAction.php
lukinovec: I unresolved this, and am now trying to use DI here. I tried adding `protected Filesystem $filesystem` to the constructor so the action could have a `$filesystem` property, but I either get an error because the JobPipeline passed the action a `Tenant` instance – that's when I make `Filesystem` the first parameter. When I make it the second parameter, I get an error saying there are not enough parameters (`Filesystem` doesn't get resolved, but I'm not getting the 'unresolvable dependency' error that I was getting in the past). 

Making `handle()` accept `Filesystem` works, but only when the action is used in a JobPipeline, e.g. if we'd want to use the action, we'd have to do `(new RemoveStorageSymlinksAction($tenants))->handle(app()->make('files'))`. @stancl, could you please hint at a better way to go about this? I mean ideally, when we don't pass anything to `handle()`, the `Filesystem` would get resolved automatically, right?

stancl: The easiest solution would be to revert the changes and make it static again. I simply *asked* if the point of the job classes is to be able to use the action in the pipeline given that the action is static, and if we could try to make the action work. If we can't, it's best to create a separate class and keep the action code clean and simple.

lukinovec: Okay, I see. I'll revert the changes.

src/Concerns/DealsWithTenantSymlinks.php
lukinovec: Added a line to the comment https://github.com/archtechx/tenancy/pull/909/commits/e9cb5aae314da384147bae01291f802a68236037

stancl: Looks great now

PR #912, opened 2022-08-03
Improve `tenants:run` command

### LINKED ISSUE #686: Out expand command definition for tenant:run
#### Description
Just add the arguments and options of the target sub command as allowed options so a command can be copied from existing docs.
Example:
Current:
```
php artisan tenants:run --tenants=xxxxx route:list --option="compact"
```
New:
```
php artisan tenants:run --tenants=xxxxx route:list --compact
```

#### Why this should be added
This allows to simply copy any existing command mentioned in docs and elsewhere.

#### 
I tested this with the following code and am willing to provide a PR.

```php
<?php

declare(strict_types=1);

namespace Stancl\Tenancy\Commands;

use Illuminate\Console\OutputStyle;
use Symfony\Component\Console\Input\ArgvInput;
use Symfony\Component\Console\Input\InputDefinition;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

class Run extends Command
{
    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Run a command for tenant(s)';

    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = "tenants:run {commandname : The command's name.}
                            {--tenants=* : The tenant(s) to run the command for. Default: all}
                            {--argument=* : The arguments to pass to the command. Default: none}
                            {--option=* : The options to pass to the command. Default: none}";

    /**
     * The sub arguments for this run
     *
     * @var array
     */
    private $subArguments = [];

    /**
     * Run the console command.
     *
     * @param  \Symfony\Component\Console\Input\InputInterface  $input
     * @param  \Symfony\Component\Console\Output\OutputInterface  $output
     * @return int
     */
    public function run(InputInterface $input, OutputInterface $output)
    {

        $this->subArguments = $this->parseArgv();
        if (empty($this->subArguments)) {
            return parent::run($input, $output);
        }

        $subCommand = $this->getApplication()->find($this->subArguments[0]);
        $this->getDefinition()->addArguments($subCommand->getDefinition()->getArguments());
        $this->getDefinition()->addOptions($subCommand->getDefinition()->getOptions());

        return parent::run($input, $output);
    }

    private function parseArgv()
    {
        $argv = $_SERVER['argv'] ?? [];
        array_shift($argv);
        $subCommandFound = false;
        $thisCommandFound = false;
        $arguments = [];
        foreach ($argv as $arg) {
            if ($arg === 'tenants:run') {
                $thisCommandFound = true;
                continue;
            }
            if ($thisCommandFound && $subCommandFound === false && strpos($arg, '-') === 0) {
                continue; // still options of tenant command
            }
            if ($thisCommandFound && $subCommandFound === false) {
                $arguments[] = $arg;
                $subCommandFound = true;
                continue;
            }
            if ($thisCommandFound && $subCommandFound !== false) {
                $arguments[] = $arg;
            }
        }

        return $arguments;
    }

    /**
     *
     * @return mixed
     */
    public function handle()
    {
        $subArguments = $this->subArguments;
        tenancy()->runForMultiple($this->option('tenants'), function ($tenant) use ($subArguments) {
            $this->line("Tenant: {$tenant['id']}");
            tenancy()->initialize($tenant);

            $subCommand = $this->getApplication()->find($this->subArguments[0]);
            $input = new ArgvInput($this->subArguments, $subCommand->getDefinition());
            $subCommand->run($input, $this->output);
        });
    }
}
```

abrardev99:
PR improves the `tenants:run` command. Let's say we want to run the `route:list` command for tenant(s). 

**Current**
`php artisan tenants:run --tenants=xxxxx route:list --option="except-vendor" --option="sort=name" `

**PR proposed**
`php artisan tenants:run --tenants=xxxxx "route:list --except-vendor --sort=name"`

This allows to copy any existing command mentioned in docs and elsewhere simply. PR originally addressed the issue https://github.com/archtechx/tenancy/issues/686.

**Output**
I'm using `Artisan::call` instead of `$this->call` to avoid complex arguments/options extraction and prepare them ecactly as needed for `$this->call` function. 
So using `Artisan::call` is simple. We print the command output, but we lost the command's original colors, and I think that's fine. 
image

### DIFF
--- a/src/Commands/Run.php
+++ b/src/Commands/Run.php
@@ -21,7 +21,7 @@ class Run extends Command
      *
      * @var string
      */
-    protected $signature = "tenants:run {commandname : The command's name.}
+    protected $signature = "tenants:run {commandname : The artisan command.}
                             {--tenants=* : The tenant(s) to run the command for. Default: all}";
 
     /**
--- a/src/Commands/Run.php
+++ b/src/Commands/Run.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Stancl\Tenancy\Commands;
 
 use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Artisan;
 
 class Run extends Command
 {
@@ -21,9 +22,7 @@ class Run extends Command
      * @var string
      */
     protected $signature = "tenants:run {commandname : The command's name.}
-                            {--tenants=* : The tenant(s) to run the command for. Default: all}
-                            {--argument=* : The arguments to pass to the command. Default: none}
-                            {--option=* : The options to pass to the command. Default: none}";
+                            {--tenants=* : The tenant(s) to run the command for. Default: all}";
 
     /**
      * Execute the console command.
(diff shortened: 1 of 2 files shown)


src/Commands/Run.php
```diff
@@ -21,9 +22,7 @@ class Run extends Command
      * @var string
      */
     protected $signature = "tenants:run {commandname : The command's name.}
```
stancl: Perhaps change this to:
```suggestion
    protected $signature = "tenants:run {command : The artisan command.}
```
So that it's clear that it's not just the *name* but also the options and arguments.

abrardev99: Thanks for the review. I've updated the `commandname`'s description. But looks like we can't use the `command` as a name because it's being used internally.
image

stancl: I see, that's fine. Thanks for checking that, I'll merge this with `commandname` then

stancl:
Also - seems like phpcsfixer fails often on PRs. It might be that it fails on all PRs that are coming from a different repository.

Could you look into that? Maybe it's configured differently than in other repos. Or it may be broken in all repos.

abrardev99:
> Could you look into that? Maybe it's configured differently than in other repos. Or it may be broken in all repos.

Sure. 

Also, you can close the linked issue too.

PR #915, opened 2022-08-08
[4.x] Configure attributes for synced resources when creating models

### LINKED ISSUE #658: Synced resources between tenants throws error when inserting new model on tenant

abrardev99:
## Problem
When a model is being created somewhere, it may need a bit more extra data than what's in the synced attributes. The synced attributes are used for updates, but we want to use different attributes for **creating**. There are multiple scenarios.

1. Central model is simple, but the tenant model is a little complex (has attributes does not exist in the central model)
2. tenant model is simple, but the central model is complex, then attaching the User to a tenant cause issues. 

Must read the connected issue in this PR for more information. 

## Solution 

This PR enables the ability to specify the attributes in both the central and tenant models. A new method `getSyncedCreationAttributes` was added to the `Syncable` interface which returns `null` by default. 

1. Returning `null` will copy the resources 1:1. When you want to return `null`, you don't need to specify.
2. Returning attribute names will pick the specified attributes from the model. 
```php
public function getSyncedCreationAttributes(): array
{
    return [
        'name' 
        'email',
        'role',
    ];
}
```

3. Returning key-value array will be considered the **default values** and used as it. 
```php
public function getSyncedCreationAttributes(): array
{
    return [
        'name' => 'Default User',
        'email' => 'default@localhost',
        'role' => 'admin',
    ];
}
```
4. Returning a mixture of attribute names (array values) and default values (key value) will pick the specified attributes from the model and the key-value array will be considered the **default values**. 
```php
public function getSyncedCreationAttributes(): array
{
    return [
         'name' 
         'email',
         'role' => 'admin',
    ];
}
```

## Usage
Let's see an example where we want to sync `users` resources where the tenant user will be created by using default values provided by central and central will be created using the mixture provided by the tenant model. This minimal setup. You should see the docs on how to set up this feature properly. 

### Setup 
Add relationship in App\Models\Tenant.php
```php
class Tenant
{
    public function users()
    {
        return $this->belongsToMany(CentralUser::class, 'tenant_users', 'tenant_id', 'global_user_id', 'id', 'global_id')
            ->using(TenantPivot::class);
    }
}
```

Prepare central User resources. 

Migrations are
```php
Schema::create('users', function (Blueprint $table) {
    $table->increments('id');
    $table->string('global_id')->unique();
    $table->string('name');
    $table->string('email');
    $table->string('password');
    $table->string('role');
    $table->string('foo');
});
```

Model 
```php
class CentralUser extends Model implements SyncMaster
{
    use ResourceSyncing, CentralConnection;

    {
        return $this->belongsToMany(ResourceTenant::class, 'tenant_users', 'global_user_id', 'tenant_id', 'global_id')
            ->using(TenantPivot::class);
    }

    {
        return ResourceUser::class;
    }

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return static::class;
    }

    {
        return [
            'global_id',
            'name',
            'password',
            'email',
        ];
    }

    public function getSyncedCreationAttributes(): array
    {
        // The tenant model will be created using the default values
        return [
            'name' => 'Default User',
            'email' => 'default@localhost',
            'password' => 'password',
            'role' => 'admin',
        ];
    }
}
```

Create a tenant User model

Migrations are: Note we don't have 'foo' column here. 
```php
Schema::create('users', function (Blueprint $table) {
    $table->increments('id');
    $table->string('global_id')->unique();
    $table->string('name');
    $table->string('email');
    $table->string('password');
    $table->string('role');
});
```

Model
```php
class TenantUser extends Model implements Syncable
{
    use ResourceSyncing;

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return CentralUser::class;
    }

    {
        return [
            'global_id',
            'name',
            'password',
            'email',
        ];
    }

    public function getSyncedCreationAttributes(): array
    {
        // The central model will be created using a mixture eg (will use the array values to pick this model attributes and will use key values as default values.
        return [
           'name',
           'email',
           'role',
           // default values
           'foo' => 'bar'
           'password' => 'password',
        ];
    }
}
```

### Let's see how we can use it. 

Creating the central user will create the tenant user by using the default values provided by `CentralUser`. 
```php
CentralUser::create()

$centralUser->tenants()->attach('t1');
```

Creating the tenant user will create the central user using the attribute and default values provided by `TenantUser`. 
```php
TenantUser::create()
```


stancl:
What are the `ci.yml` changes here?

stancl:
Wrote a comment regrading the above on Basecamp 👍🏻

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +126,34 @@
     ], ResourceUser::first()->getAttributes());
 });
 
+test('creating the resource in tenant database creates it in central database and used custom attributes', function () {
+    // Assert no user in central DB
```
stancl:
```suggestion
    // Assert no user exists in central DB
```

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +126,34 @@
...
+
+    $tenant = ResourceTenant::create();
+    pest()->artisan('tenants:migrate', [
+        '--path' => __DIR__ . '/Etc/synced_resource_migrations/custom',
+        '--realpath' => true,
+    ])->assertExitCode(0);
+
+    tenancy()->initialize($tenant);
+
+    // Create the same user in tenant DB
```
stancl: What is "the same" user?

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +126,34 @@
     ], ResourceUser::first()->getAttributes());
 });
 
+test('creating the resource in tenant database creates it in central database and used custom attributes', function () {
```
stancl: The test name seems a bit messy

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +126,34 @@
...
+        'email' => 'john@localhost',
+        'password' => 'secret',
+        'role' => 'commenter', // unsynced
+        'code' => 'bar' // extra column which does not exist in central users table
+    ]);
+
+    tenancy()->end();
+
+    // assert user was created
+    expect(CentralUser::first()->global_id)->toBe('acme');
```
stancl: I'd make a specific assertion about the `code`:
```suggestion
        'code' => 'bar' // extra column which does not exist in central users table
    ]);

    tenancy()->end();

    // Assert central user was created without `code` property
    expect(CentralUser::first()->global_id)->toBe('acme');
    expect(CentralUser::first()->code)->toBeNull();
```
Also, does the `role` exist in the central DB?

tests/ResourceSyncingTest.php
```diff
@@ -598,6 +626,18 @@ public function getSyncedAttributeNames(): array
             'email',
         ];
     }
+
+    public function getCreateAttributeNames(): array
+    {
+        // attributes should be used when syncing resources from central to tenant DB
```
stancl: Better writing:
```suggestion
        // Attributes used when syncing resources from central to tenant DB
```

Also, is this *really* what that method is? I don't think it necessarily has anything to do with central/tenant DB.

tests/ResourceSyncingTest.php
```diff
@@ -633,4 +673,16 @@ public function getSyncedAttributeNames(): array
             'email',
         ];
     }
+
+    public function getCreateAttributeNames(): array
+    {
+        // attributes should be used when syncing resources from tenant to central DB
```
stancl:
```suggestion
        // Attributes used when syncing resources from tenant to central DB
```

stancl: The description also seems opposite of what it is in the above model?

tests/ResourceSyncingTest.php
abrardev99: > does the role exist in the central DB

Yes.

abrardev99:
@stancl from your review on comments https://github.com/archtechx/tenancy/pull/915#discussion_r941930758 and https://github.com/archtechx/tenancy/pull/915#discussion_r941930690, I realized that we need this method only when creating Central model and we need sync in tenant model (only when creating). In fact, the original issue was also related to creating a model in central DB when creating in tenant DB with the extra column. 

In other words, we don't need the `getCreateAttributeNames` method for the central model. As it implements `Syncable`, adding method in here enforces to add in tenant model too because `SyncMaster` is extending from `Syncable`. 

What do you suggest that the developer implement this method only in the tenant model? I can think of a separate interface.

tests/ResourceSyncingTest.php
abrardev99: Maybe better 

```suggestion
        // Attributes used when creating resources from tenant to central DB
```

As these attributes are used only when **creating** (not syncing after) models from tenant to central.

tests/ResourceSyncingTest.php
abrardev99: I don't know, most probably leftover comments when writing tests. Actually, it should be only `// Create the user in tenant DB`.

tests/ResourceSyncingTest.php
abrardev99: `creating the resource in tenant database creates it in central database with provided attribute`

Is this better?

stancl:
Create a message on BC, we'll discuss it there in depth 👍🏻

abrardev99:
@stancl I've marked reviews as resolved because new changes change all of the code.

stancl:
Can you update the PR description?

abrardev99:
> Can you update the PR description?

Done.

stancl:
Code samples for points 2. and 3. in the README seem to be swapped. Also, please fix formatting and use proper syntax highlighting.

src/Listeners/UpdateSyncedResource.php
```diff
@@ -112,11 +113,24 @@ protected function updateResourceInTenantDatabases($tenants, $event, $syncedAttr
...
-                    $localModel = $localModelClass::create($eventModel->getAttributes());
+                    $localModel = $localModelClass::create($this->getResourceCreationAttributes($eventModel));
                 }
 
                 event(new SyncedResourceChangedInForeignDatabase($localModel, $tenant));
             });
         });
     }
+
+    protected function getResourceCreationAttributes(Syncable $model): array
```
stancl: I'd name this something else than the method on the model. The code is now a bit confusing, seeing `$this->getResourceCreationAttributes($eventModel)` isn't immediately clear that it's something on this class rather than on the model.

Perhaps rename this to `getAttributesForCreation($model)`.

FWIW I don't fully like the names of either of these two methods right now but I can't think of anything better, so I'll rename them later.

src/Listeners/UpdateSyncedResource.php
```diff
@@ -112,11 +113,24 @@ protected function updateResourceInTenantDatabases($tenants, $event, $syncedAttr
...
+    {
+        $attributes = $model->getAttributes();
+
+        if ($model->getResourceCreationAttributes()) {
+            // If function returned array is key-value, We assume default values are provided
+            // if array is plain values, fetch attributes from model
```
stancl: Please use perfect grammar in comments that are part of the package's codebase. This can also be rewritten to be more clear.

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +127,244 @@
     ], ResourceUser::first()->getAttributes());
 });
 
+test('creating the resource in tenant database creates it in central database as 1:1 copy when creation attributes are not specified', function () {
```
stancl: Also add an assertion that the method for creation attributes returns `null`

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +127,244 @@
...
+    tenancy()->end();
+
+    // Assert central user and Resource user has exact same attributes and values
+    expect(CentralUser::first()->toArray())->toEqual(ResourceUser::first()->toArray());
+});
+
+test('creating the resource in tenant database creates it in central database with default attributes values', function () {
+    // override method in ResourceUser class to return attribute default values
+    class ResourceUserWithDefaultValues extends ResourceUser {
+        public function getResourceCreationAttributes(): array
```
stancl: We don't define classes inside of tests. They should be at the end of the test file.

src/Listeners/UpdateSyncedResource.php
```diff
@@ -112,11 +113,24 @@ protected function updateResourceInTenantDatabases($tenants, $event, $syncedAttr
...
+            $attributes = Arr::isAssoc($model->getResourceCreationAttributes()) ? $model->getResourceCreationAttributes() : $model->only($model->getResourceCreationAttributes());
```
stancl: If it's an associative array, you should *still* be able to copy values from the model. You don't want to use *just* the default values to create the model.

So in the implementation, I think they should be merged into the model's synced attributes?

stancl: And before making any logical changes, use TDD. Update the tests to assert what we want to see, make them fail, and then modify the code to make them pass.

It's a semi complex feature and we don't want the tests to pass accidentally, so we want to see them fail first and then pass when the implementation is fixed.

src/Listeners/UpdateSyncedResource.php
abrardev99: For now, renaming the method to `getAttributesForCreation`.

src/Listeners/UpdateSyncedResource.php
abrardev99: > You don't want to use just the default values to create the model.

If I understand it correctly, You are saying even if the developer specified the default value of a column, and the same column exists in the sync attributes, I should copy the value from the model instead of using the default value.

If I'm correct, What if the developer doesn't want to use the column value and wants to provide the default value always for some columns even if they exist in the sync attribute?

stancl: > You are saying even if the developer specified the default value of a column, and the same column exists in the sync attributes

The point of defaults is that they're used when we don't have any actual value to use.

However this is not what I meant in my comment. I was talking about the **other** attributes. The code here looks like it will use ONLY the defaults. But we want to use *other columns* as well. The point of the defaults here is to fill the columns that we don't have values for. But we still want to use the columns that we do have values for.

So I'm talking about different columns, not overriding values for the same column.

abrardev99: > So I'm talking about different columns, not overriding values for the same column.

I see, merging them makes sense.

src/Listeners/UpdateSyncedResource.php
abrardev99: Resolving this review because I've rewritten the entire method.

src/Listeners/UpdateSyncedResource.php
abrardev99: I've made the changes. As you said, I did TDD. Also, the functions are almost rewritten, and the code looks clean now.

src/Contracts/Syncable.php
```diff
@@ -15,4 +15,6 @@ public function getCentralModelName(): string;
     public function getSyncedAttributeNames(): array;
 
     public function triggerSyncEvent(): void;
+
+    public function getResourceCreationAttributes(): array|null; // todo come up with a better name
```
leandrogehlen: Maybe `getCreationAttributeNames` is better and compatible with `getSyncedAttributeNames`

stancl: What do you mean? The methods work differently, they return values in different format.

leandrogehlen: Sorry, i haven't seen the  possibility to return attribute values instead of attribute names.
But i think the name of method could be `getCreationAttributes`

stancl: Ah, yeah, if we're not using the "resource" terminology anywhere in the contract then that'd probably work. @abrardev99 can you make the change?

My only concern with too generic names like that is that they could conflict with Laravel code eventually or *feel* like they're something Laravel-related. When it should be clear that they're related to Tenancy resource syncing.

abrardev99: > @abrardev99 can you make the change?

Sure I can.

leandrogehlen: > My only concern with too generic names like that is that they could conflict with Laravel code eventually or feel like they're something Laravel-related. When it should be clear that they're related to Tenancy resource syncing.

After @stancl comment, 
unfortunately, i think that method name could be `getSyncedCreationAttributes`
is compatible with `getSyncedAttributeNames` and avoid  conflict with Laravel.

abrardev99: I like that one `getSyncedCreationAttributes`, BTW.

abrardev99: @stancl Renamed the method and added todo to come up a better name

stancl:
@abrardev99 ready for final review now? And are all of my comments above addressed? Some are now collapsed in the "x hidden comments"

abrardev99:
> ready for final review now? And are all of my comments above addressed? Some are now collapsed in the "x hidden comments."

Yes, all the comments are addressed and PR is ready for final review.

src/Listeners/UpdateSyncedResource.php
abrardev99: Resolving this review since implementation got changed and requested changes are implemented.

stancl:
One thing I'm not 100% sure about is that we're using the creation attributes for both the central model and the tenant model, right?

CleanShot 2022-09-23 at 4 24 41@2x

Are the creation attributes defined separately on the central model and the tenant model? Or are we using the same ones for creating the record in both cases?

abrardev99:
> we're using the creation attributes for both the central model and the tenant model, right?

Yes correct. 

> Are the creation attributes defined separately on the central model and the tenant model? 

Yes. If you see a Syncable interface, we have the [getSyncedCreationAttributes](https://github.com/archtechx/tenancy/pull/915/files#diff-d7930fd9aa38daf04b1e43797aa014f9128858afad48b658b7d614d92a5ee5bbR19) method. As the Syncable interface is implemented by both models, so creation attributes are defined separately on the central model and the tenant model.

src/Listeners/UpdateSyncedResource.php
```diff
@@ -112,12 +113,28 @@ protected function updateResourceInTenantDatabases($tenants, $event, $syncedAttr
                 if ($localModel) {
                     $localModel->update($syncedAttributes);
                 } else {
-                    // When creating, we use all columns, not just the synced ones.
-                    $localModel = $localModelClass::create($eventModel->getAttributes());
+                    $localModel = $localModelClass::create($this->getAttributesForCreation($eventModel));
```
stancl, line 132: Could `$eventModel` come from the central database here? If so, then I think according to what we discussed here https://github.com/archtechx/tenancy/pull/915#issuecomment-1255807197, it would take the `getSyncedCreationAttributes()` from the **incoming** model rather than the local one that we're trying to create?

Do we have tests to ensure that the `getSyncedCreationAttributes()` are taken from the right model?

abrardev99: I'm a little confused here.

> it would take the `getSyncedCreationAttributes()` from the incoming model rather than the local one that we're trying to create?

If I understand correctly, You are saying that `getSyncedCreationAttributes()` are coming from the incoming model instead local model. 

Yes, it is. Why should it come from the "local one we're trying to create"? I mean, we are copying one model to another, So we are asking the resource model (which is being copied) to provide the attributes for the target model (the model will be created). 

> Do we have tests to ensure that the getSyncedCreationAttributes() are taken from the right model?

We don't have a specific test for this. Don't you think [current tests](https://github.com/archtechx/tenancy/pull/915/files#diff-0c32d140e95604d5ab8852beb5eba2255180184351adfaa0ff7b2895012d4d7dR130-R309) cover that `getSyncedCreationAttributes()` are taken from the right model?

stancl: > Why should it come from the "local one we're trying to create"?

1. We're creating the local model
2. The synced attributes include a list of attributes that are used, including defaults
3. We want to use defaults for the local model, not defaults for the incoming model

I'm not 100% sure about this, but it seems wrong to use `getSyncedCreationAttributes()` from a different model than the one we're creating.

Read our previous conversation: https://github.com/archtechx/tenancy/pull/915#issuecomment-1255807197

I specifically asked if there can be different `getSyncedCreationAttributes()` on the central and tenant model.

stancl: Or do you think that the `getSyncedCreationAttributes()` **should always be called on the opposite model** than the one that we're creating?

abrardev99: > Or do you think that the getSyncedCreationAttributes() should always be called on the opposite model than the one that we're creating?

Yes, that is what I see and also code doing right now. Do you want me to write a detailed message about the implementation of BC?

stancl: Yeah 👍🏻

abrardev99: Based on the BC discussion, a comment has been added so resolve the review.

tests/ResourceSyncingTest.php
```diff
@@ -877,4 +957,20 @@ public function getSyncedCreationAttributes(): array
...
+    public function getSyncedCreationAttributes(): array
+    {
+        // Sync name, email and password but provide default value for role
+        return
+            [
+                'global_id',
+                'name',
+                'password',
+                'email',
+                'role' => 'admin',
```
stancl: It seems that this array is the same for the central model and the tenant model? According to what we discussed on Basecamp, we want to have tests which define this method differently in different (central/tenant) models. And assert that the created model has the right data.

stancl:
I think that the tests could be a bit simplified.
1. For example, `creating the resource in tenant database creates it in central database with attributes names` seems to test the behavior that existed before this PR? It tests that the synced attribute names are used, but I don't see you making any assertions about **unsynced attributes that are part of *creation* attributes** being included on the created model
2. It seems that each test has two variants? 1) central -> tenant, 2) tenant ->central. Is there any way we could avoid the duplication, perhaps with something like [datasets](https://pestphp.com/docs/datasets). Only do this if it can improve the tests. If they were to become more complex with datasets, it's not worth making this change. But try to find other ways to simplify this. Perhaps we could assert that the logic works correctly in both directions in a single test. So there'd be a test for a list of attribute names used for creation, and it'd assert that this works in both directions. Then there'd be a test for a mix of attribute names and default values and it'd assert that it works in both directions. Etc. Try to find something like that to avoid the duplication. We ideally only want one test for each case (attribute names, default values, mix of attribute names and default values, etc).
3. I'd improve the terminology used in the test names, e.g. the `attribute names` in test names are a bit confusing, but if you can't do this I can make the changes after you clean up the tests.

stancl:
Just a note for future reference: the recent commits add behavior for excluding the model key from the attributes used for creating the model in other contexts — https://3.basecamp.com/5170965/buckets/23650968/chats/4064741444@5394258122

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +127,229 @@
     ], ResourceUser::first()->getAttributes());
 });
 
+// This tests attribute list on the central side, and default values on the tenant side
+// Those two don't depend on each other, we're just testing having each option on each side
+// using tests that combine the two, to avoid having an excessively long and complex test suite
+test('sync resource creation works when central model provides attributes and resource model provides default values', function () {
+    // when central model provides attributes => resoucre model will be created from the attribute values
```
stancl: Is this trying to say that when the central model provides **a list of attributes**, the resource model will be created from those attributes' values?

tests/ResourceSyncingTest.php
```diff
@@ -126,6 +127,229 @@
...
+    [$tenant1, $tenant2] = creareTenantsAndRunMigrations();
```
stancl: `creare`

tests/ResourceSyncingTest.php
```diff
@@ -628,9 +880,148 @@ public function getCentralModelName(): string
...
+            ->using(TenantPivot::class);
+    }
+
+    public function getTenantModelName(): string
+    {
+        return ResourceUserForDifferentSchema::class;
+    }
+}
+
+class ResourceUserForDifferentSchema extends ResourceUser {
```
stancl: You use the `different schema` terminology a lot here. Can you explain what you mean by it? The naming isn't very clear.

If it's about e.g. a missing column, that should be what's expressed in the name, rather than "different schema".

tests/ResourceSyncingTest.php
stancl: If so, then the comment is useful, but it should be more clear by:
1. Using the same terminology everywhere. See the first lines of the comments I added in the last commit. `list of attributes`, `default values`, `mixed attribute list/defaults`, `no specified attributes`. If you can think of more accurate terms, we can use those, but these are the ones I used in the comments and think they're the most descriptive. The rest of the code (tests, implementation, comments, ...) should follow consistent terminology
2. Making it clear when the resource model "will be created". This is right above a a line that creates two models, so it sounds like it's being created here. Maybe it'd be more appropriate above the `attach()` call, to say that the resource user will be created based on attributes from `$centralModel`?

tests/ResourceSyncingTest.php
```diff
@@ -537,10 +761,37 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
     expect(ResourceUser::first()->role)->toBe('commenter');
 }
 
-function migrateTenantsResource()
+function creareTenantsAndRunMigrations(): array
```
stancl: Add `/**` comments above all of the helpers you created to explain the point of them. In general, test helpers are pretty confusing unless they're super simple, so they should always be explained well.

tests/ResourceSyncingTest.php
abrardev99: Naming is confusing indeed. I'll update soon.

abrardev99: > You use different schema terminology a lot here. Can you explain what you mean by it

I prepend `DifferentSchema` to the classes making use of different table names and columns(in some cases). See all the classes below:
image

More natural name for `ResourceTenantForDifferentSchema` can be just `ResourceTenant` but same class exists already. So this name feels appropriate.

tests/ResourceSyncingTest.php
abrardev99: Added the comments on each helper method.

tests/ResourceSyncingTest.php
abrardev99: >Making it clear when the resource model "will be created". This is right above a a line that creates two models, so it sounds like it's being created here.

FYI, I followed the rule to add comments when that specific part was going to start. 

> Maybe it'd be more appropriate above the attach() call, 

It makes sense right above the attach call. 
What about this second comment? L218. 
https://github.com/archtechx/tenancy/blob/93efa56f5932e93c6141af520cf4a78f67ff025e/tests/ResourceSyncingTest.php#L218-L222
Is it at the appropriate position? I'm also considering moving it to the right above the `ResourceUserProvidingAttributeNames::create` line.

stancl: > FYI, I followed the rule to add comments when that specific part was going to start.

I don't understand what you mean by this.

> What about this second comment

1. I wouldn't use `=>`, I'd just make it a normal sentence. This may seem like a nitpick but it makes the comment more understandable at first look
2. I'd change `attributes` to `list of attributes` like I mentioned above — pick some terminology to follow everywhere. `list of attributes` is probably better than `attributes` because the latter could be understood as attribute names & values. Whereas `list of attributes` is clearly just a list of the attributes used.

> I'm also considering moving it to the right above the

Yes, move it there. I don't think `initialize()` is the appropriate place because I don't think `initialize()`  creates the models here.

tests/ResourceSyncingTest.php
stancl: > I prepend DifferentSchema

You mean suffix, right?

> making use of different table names and columns

So essentially a case where the tenant DB has less/more columns than the central DB?

Some thoughts:
1. It can indeed be difficult to think of a better term than "different schema", so the current term paired with a clear comment could work
2. However: **in what way** is it different? Does it have less or more columns than the other model?
3. And what's the other model here — the central schema? Since you're using "different schema": different to what? Is the schema of `users` completely identical for the central DB and the tenant DBs in the `master` version of these tests? So your point here is that *only in these tests* we introduce a different schema for the `users` table in the tenant DB, compared to the central DB.

abrardev99: > You mean suffix, right?

yes, I meant suffix. 🤦‍♂️

> 2. However: in what way is it different? Does it have fewer or more columns than the other model?

1) both tables have different names. 2) Columns are actually the same in both models but different from the master version or first setup for sync resources (ResourceUser, CentralUser). 

> 2. Is the schema of users completely identical for the central DB and the tenant DBs in the master version of these tests?

No. They have the same table names, but the columns are somewhat different. I mean, that was the point of an original issue when one model has an extra column, it causes a problem. We are trying to solve it by providing list of attributes. 
At some point, I thought having a different setup like this was repetitive. We are kind of testing the same thing in the master version, except the table names are the same.

tests/ResourceSyncingTest.php
abrardev99: Updated the comments and moved them to appropriate positions.

tests/ResourceSyncingTest.php
stancl: > No. They have the same table names, but the columns are somewhat different

Even **in the `master` version of these tests?** Meaning, the tests that we already have in the `master` branch in this file.

src/Listeners/UpdateSyncedResource.php
```diff
@@ -150,10 +150,11 @@ protected function getAttributesForCreation(Model&Syncable $model): array
...
         if (Arr::isAssoc($model->getSyncedCreationAttributes())) {
             // Developer provided the default values (key => value) or mix of default values and attribute names (values only)
-            // We will merge the default values with sync attributes
-            [$attributes, $defaultValues] = $this->getAttributeNamesAndDefaultValues($model);
+            // We will merge the default values with provided attributes and sync attributes
+            [$attributeNames, $defaultValues] = $this->getAttributeNamesAndDefaultValues($model);
+            $attributes = $model->only(array_merge($model->getSyncedAttributeNames(), $attributeNames));
 
-            return array_merge($model->only(array_merge($model->getSyncedAttributeNames(), $attributes)), $defaultValues);
+            return array_merge($attributes, $defaultValues);
```
stancl: What's the point of the change in this commit?

abrardev99: Code clarity and readability.

stancl: I see, the change makes sense on second look 👍🏻

tests/ResourceSyncingTest.php
stancl: Closing this since we got rid of these unnecessary tests.

stancl:
A note about code style — try to write code that follows our code style on your own, don't depend on php-cs-fixer fixing your code. It should just catch edge cases. We don't use it in tests, but tests should still follow our code style. Basic things like spaces in `function () {`, `class Foo` having `{` on the next line, etc.

stancl:
@abrardev99 There's still this:

> Note: Implementation has been changed. See PR reviews. I'll update the description when resolved all the reviews.

in the PR description. Make the PR description up to date since we're about to merge it.

abrardev99:
> @abrardev99 There's still this:
> 
> > Note: Implementation has been changed. See PR reviews. I'll update the description when resolved all the reviews.
> 
> in the PR description. Make the PR description up to date since we're about to merge it.

Forgot to remove it. Will update it.

abrardev99:
> A note about code style — try to write code that follows our code style on your own, don't depend on php-cs-fixer fixing your code. It should just catch edge cases. We don't use it in tests, but tests should still follow our code style. Basic things like spaces in `function () {`, `class Foo` having `{` on the next line, etc.

Noted. I do try to follow standards but sometimes miss especially when there is a lot of code.

tests/ResourceSyncingTest.php
```diff
@@ -628,9 +876,106 @@ public function getCentralModelName(): string
...
+                'email',
+                'role',
+                'foo' => 'bar'
+            ];
+    }
+
+}
+
+// override method in CentralUser class to return attribute default values
+class CentralUserProvidingDefaultValues extends CentralUser
```
stancl, line 925: @abrardev99 I will merge this PR now so that the related tasks can be worked on tomorrow, but there are still some minor things you should fix tomorrow.

Re-read what I wrote on BC about the `ResourceUser` terminology: https://3.basecamp.com/5170965/buckets/23651018/messages/5438582828#__recording_5498105640

If e.g. `ResourceUserProvidingDefaultValues` is meant to be the counterpart/opposite to `CentralUserProvidingDefaultValues`, then wouldn't it make more sense to use `TenantUserProvidingDefaultValues`?

tests/ResourceSyncingTest.php
```diff
@@ -628,9 +876,106 @@ public function getCentralModelName(): string
...
+        return [
+            'name',
+            'email',
+            'role' => 'admin',
+            'password' => 'secret',
+        ];
+    }
+}
+
+class ResourceUserProvidingMixture extends ResourceUser
```
stancl, line 970: Same as https://github.com/archtechx/tenancy/pull/915/files#r1013157780

tests/ResourceSyncingTest.php
stancl: `ResourceUser` exists from previous versions of this file, but I think back then there were only two main classes:
- Just a "User" class (named `ResourceUser` to avoid conflicts, as I explained)
- `CentralUser` class that specifically says that it's central

But now that we have many more models, and they're always in pairs (`XUserProviding...`), I'm thinking that it'd make more sense to name them `TenantUser*`.

That said, **I'd only do this for the suffixed classes**, since we don't want to use a generic name like `TenantUser` in this test. In that **one** case, the `Resource` prefix does make sense.

So we want to:
- Use `Tenant` instead of `Resource` in all suffixed (e.g. `X Providing ...`) models
- Either keep `ResourceUser` as it is, or rename it to `ResourceTenantUser`. Or maybe just add a comment saying `// Tenant users`

stancl:
Thanks for working on this PR! Happy to have it finally merged 🚀

After making the small tweaks I described above, you can work on these two tasks that are related to resource syncing:
1. https://3.basecamp.com/5170965/buckets/23651018/todos/5496514869
2. https://3.basecamp.com/5170965/buckets/23651018/todos/5426057990

The first one is very simple so it shouldn't take more than 30 minutes I think. The second one is more complex so that's what you can work on for the rest of the day 👍

tests/ResourceSyncingTest.php
abrardev99: @stancl Sent PR https://github.com/archtechx/tenancy/pull/992 for improving names.

PR #917, opened 2022-08-12
Adding a new session value 'tenancy_impersonated_by' when you are impersonating a tenant.

### LINKED ISSUE #548: Highlight when Impersonating
#### Description
When impersonating a tenant user it would be good if this could be checked and acted on

#### Why this should be added
When providing user support or debugging it is helpful to impersonate a user but when that is finished there is a need to switch back to your own login, to support others etc.

Ideally it would be possible to detect that a login was an impersonation and at least display this e.g. "Admin acting as User". Even better would be the option to return to being Admin

nathandaly:
This PR adds a new session value `tenancy_impersonated_by` in the tenant scope when you are impersonating a tenant.

This is similar to how Laravel Nova adds the `nova_impersonated_by session` value and could be used to identify when you are impersonating, who is impersonating. This may be useful for additional application logic and UI niceties such as an impersonation banner.

### DIFF
--- a/src/Features/UserImpersonation.php
+++ b/src/Features/UserImpersonation.php
@@ -52,6 +52,10 @@ class UserImpersonation implements Feature
         Auth::guard($token->auth_guard)->loginUsingId($token->user_id);
 
         $token->delete();
+        
+        session()->put(
+            'tenancy_impersonated_by', Auth::user()?->getAuthIdentifier()
+        );
 
         return redirect($token->redirect_url);
     }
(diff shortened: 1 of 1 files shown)


nathandaly:
Docker container failed to initialise.

stancl:
Just for some context — we're looking into this internally, since we were just about to work on something very similar. Using a session in the central context (or on the central domain) is potentially problematic, so we're going through the different solutions that could be used here to see what will work.

stancl:
Resolved in #1029

Issue #412, opened 2020-05-26
Allow setting array in `storage_to_config_map`


Neophen:
#### Description
Allow setting array in `storage_to_config_map` to be able to reuse one setting
![usecase](https://user-images.githubusercontent.com/6092928/82928236-a3c32580-9f8a-11ea-9e82-d8cc91c068c6.png)

changes needed in stancl/tenancy/src/Features/TenantConfig.php
change 1: add an is_array check to `__construct`
```php
  public function __construct(Repository $config)
  {
    $this->config = $config;

    foreach ($this->getStorageToConfigMap() as $index => $configKey) {
      if (is_array($configKey)) {
        foreach ($configKey as $key) {
          $this->originalConfig[$key] = $this->config[$index];
        }
      } else {
        $this->originalConfig[$configKey] = $this->config[$configKey];
      }
    }
  }
```

change 2: add an is_array to `unsetTenantConfig`
```php
  public function unsetTenantConfig(): void
  {
    foreach ($this->getStorageToConfigMap() as $key => $configKey) {
      if (is_array($configKey)) {
        foreach ($configKey as $originalConfigKey) {
          $this->config[$key] = $this->originalConfig[$originalConfigKey];
        }
      } else {
        $this->config[$configKey] = $this->originalConfig[$configKey];
      }
    }
  }
```

#### Why this should be added
An easy way to not to duplicate keys on tenant just to manage more than one setting

PR #920, opened 2022-08-19
Add support for nested tenant config override

rootgog:
This PR adds the ability to assign from nested JSON values on the Tenant model while staying backwards compatible with the existing implementation.

Essentially eloquent's model `getAttribute` is replaced with the Laravel helper function `Arr::get` when the values are extracted from the tenent model which allows for dot notated paths to be used to map nested JSON values.

For example this allows implementations such as the following
```php
\Stancl\Tenancy\Features\TenantConfig::$storageToConfigMap = [
    'whitelabel.config.theme' => 'whitelabel.theme',
];
```

### DIFF
--- a/src/Features/TenantConfig.php
+++ b/src/Features/TenantConfig.php
@@ -6,6 +6,7 @@ namespace Stancl\Tenancy\Features;
 
 use Illuminate\Contracts\Config\Repository;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Event;
 use Stancl\Tenancy\Contracts\Feature;
 use Stancl\Tenancy\Contracts\Tenant;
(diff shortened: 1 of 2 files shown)


stancl:
Hey, thanks for the PR! I was trying to find a past discussion about this because I *think* there was a convo about how this part of the code works, so I wanted to check if there'd be any problem with using `Arr::get()`. But it seems like this will work fine.

I found this issue https://github.com/archtechx/tenancy/issues/412 and that *may* have been it (I think there was some other discussion too but this looks similar enough). Essentially, "supporting arrays" is a thing on both sides of the `foo => bar` mapping here. I believe the commit which fixed #412 https://github.com/archtechx/tenancy/commit/6dad6ce49a2a24e070b768ee93e0a3792a8d8924 added support for mapping a single tenant value to multiple *config* values.

And this PR will support reading from arrays on the *tenant* side. Is that correct?

The implementation looks good and so does the test. Though I think I'll change the target to 4.x. We're focusing on v4 development now so we prefer that new features are added there, and 3.x changes are only things like urgent bug fixes.

If I change the target branch, there will likely be a conflict in the test file, since we migrated to pest on the master branch. You can try resolving that on your own, or @abrardev99 will fix it on Monday and we'll get this merged 👍🏻

stancl:
Ah hmm, seems like your branch is based on `3.x` which has some additional changes that can't be easily merged. I guess we'll merge this into `3.x` then and merge `3.x` into `master`

stancl:
Thanks a lot for the PR!

PR #923, opened 2022-08-26
[4.x] Improve `tenants:run` command to execute Input\Output commands 

### LINKED ISSUE #686: Out expand command definition for tenant:run

abrardev99:
As discussed internally, the `tenants:run` command got stuck if the sub/provided command asked questions. A common use case is creating a Passport client in the tenant context using the Passport provided command `passsport:client` which asks questions like name, email, etc. 

This PR fixes the above issue. 

image

### DIFF
--- a/src/Commands/Run.php
+++ b/src/Commands/Run.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace Stancl\Tenancy\Commands;
 
 use Illuminate\Console\Command;
-use Illuminate\Support\Facades\Artisan;
+use Illuminate\Contracts\Console\Kernel;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
 
 class Run extends Command
 {
@@ -29,12 +31,29 @@ class Run extends Command
      */
     public function handle()
     {
-        tenancy()->runForMultiple($this->option('tenants'), function ($tenant) {
+        $argvInput = $this->ArgvInput();
+        tenancy()->runForMultiple($this->option('tenants'), function ($tenant) use ($argvInput) {
             $this->line("Tenant: {$tenant->getTenantKey()}");
 
-            Artisan::call($this->argument('commandname'));
-            $this->comment('Command output:');
-            $this->info(Artisan::output());
+            $this->getLaravel()
+                ->make(Kernel::class)
+                ->handle($argvInput, new ConsoleOutput);
         });
+    }
+
+    /**
+     * Get command as ArgvInput instance
+     *
+     * @return ArgvInput
+     */
+    protected function ArgvInput(): ArgvInput
+    {
+        // Convert string command to array
+        $subCommand = explode(' ', $this->argument('commandname'));
+
+        // Add "artisan" as first parameter because ArgvInput expect "artisan" as first parameter and later remove it
+        array_unshift($subCommand , 'artisan');
+
+        return new ArgvInput($subCommand);
     }
 }
--- a/src/Commands/Run.php
+++ b/src/Commands/Run.php
@@ -49,7 +49,7 @@ class Run extends Command
         // Convert string command to array
         $subCommand = explode(' ', $this->argument('commandname'));
 
-        // Add "artisan" as first parameter because ArgvInput expect "artisan" as first parameter and later remove it
+        // Add "artisan" as first parameter because ArgvInput expects "artisan" as first parameter and later removes it
         array_unshift($subCommand, 'artisan');
 
         return new ArgvInput($subCommand);
--- a/src/Commands/Run.php
+++ b/src/Commands/Run.php
@@ -42,9 +42,7 @@ class Run extends Command
     }
 
     /**
-     * Get command as ArgvInput instance
-     *
-     * @return ArgvInput
+     * Get command as ArgvInput instance.
      */
     protected function ArgvInput(): ArgvInput
     {
@@ -52,7 +50,7 @@ class Run extends Command
         $subCommand = explode(' ', $this->argument('commandname'));
 
         // Add "artisan" as first parameter because ArgvInput expect "artisan" as first parameter and later remove it
-        array_unshift($subCommand , 'artisan');
+        array_unshift($subCommand, 'artisan');
 
         return new ArgvInput($subCommand);
     }
--- a/tests/CommandsTest.php
+++ b/tests/CommandsTest.php
@@ -179,6 +179,17 @@ test('run command with array of tenants works', function () {
         ->expectsOutput('Tenant: ' . $tenantId2);
 });
 
+test('run command works when sub command asks question and accepts argument', function () {
+    $id = Tenant::create()->getTenantKey();
+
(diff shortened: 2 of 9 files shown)


tests/CommandsTest.php
```diff
@@ -179,6 +179,17 @@
         ->expectsOutput('Tenant: ' . $tenantId2);
 });
 
+test('run command works when sub command asks question and accepts argument', function () {
+    $id = Tenant::create()->getTenantKey();
+
+    Artisan::call('tenants:migrate', ['--tenants' => [$id]]);
```
stancl: What's this here for?

src/Commands/Run.php
```diff
@@ -29,12 +31,27 @@ class Run extends Command
...
+
+    /**
+     * Get command as ArgvInput instance.
+     */
+    protected function ArgvInput(): ArgvInput
+    {
+        // Convert string command to array
+        $subCommand = explode(' ', $this->argument('commandname'));
+
+        // Add "artisan" as first parameter because ArgvInput expect "artisan" as first parameter and later remove it
```
stancl:
```suggestion
        // Add "artisan" as first parameter because ArgvInput expects "artisan" as first parameter and later removes it
```

tests/CommandsTest.php
```diff
@@ -179,6 +179,17 @@
...
+
+    pest()->artisan("tenants:run --tenants=$id 'age:ask Abrar' ")
+        ->expectsQuestion('What is your age?', 22)
+        ->expectsOutput("Tenant: $id")
+        ->expectsOutput("Abrar's age is 22.");
```
stancl: Do these tests verify that the tenant context is *actually* used for the command?

I guess some of the other tests in `CommandsTest` do?

abrardev99: > Do these tests verify that the tenant context is actually used for the command?

Yes, they do. `->expectsOutput("Tenant: $id")` verifies that we are in tenant context.

tests/CommandsTest.php
abrardev99: Ah, I was migrating tenant migrations because I was dumping tenant users in the `Run.php` command for testing purposes. So this is a leftover code. I'll remove it, and tests will pass without this line.

tests/CommandsTest.php
stancl: Ah, I see 👍🏻

stancl: Hmm I looked at it now and I'm not sure if it asserts that we're *inside* the tenant's context?

`tenants:run` simply prints the id of the tenant. But we're not asserting that the code you're using is actually *using* the tenant context.

The reason why I want to test this in detail is that you're instantiating the `Kernel` in the implementation (which we'd ideally not do), so I'm not 100% sure if the tenant context is shared.

abrardev99: I got the point. 

So now I've updated the tests and command too. TestCommand now creates a user. So in tests, I'm asserting that the user created in tenant context, not in central. I believe this ensures we are running commands in the tenant context.

tests/CommandsTest.php
```diff
@@ -179,13 +179,29 @@
         ->expectsOutput('Tenant: ' . $tenantId2);
 });
 
-test('run command works when sub command asks question and accepts argument', function () {
-    $id = Tenant::create()->getTenantKey();
+test('run command works when sub command asks question and accepts argument', closure: function () {
```
stancl: Why the `closure`?

stancl: Fixed this now

stancl:
@abrardev99 Please always import classes, we very rarely use `\This\Syntax` inside code.

https://github.com/archtechx/tenancy/pull/923/commits/c6a565295c9690ec2dd6c569f920bd4d5c1a79e7#diff-2d46e8828aea0422777576ca8e903262bf80a262978ec02ad3d0783c38fe1b32L200

stancl:
Thanks for the PR!

PR #925, opened 2022-08-29
[4.x] Set tenant as a default parameter for the URLs when using Path identification 

### LINKED ISSUE #471: Alternative for route() when using path identification

abrardev99:
This PR adds the tenant as a default parameter for the URLs when using Path identification. Once the tenant is initialized, you are no longer required to pass its value when generating URLs via the route helper. 

For example:
```php 
route('my-route-name', ['tenant' => tenant(), 'my-app-param' => 'foo']) 
```
Will be 
```php 
route('my-route-name', ['my-app-param' => 'foo']) 
```


abrardev99:
@stancl Note that we have a PR https://github.com/archtechx/tenancy/pull/916 for the `tenant_path_route ` helper function. Maybe you want to close it now.

tests/PathIdentificationTest.php
```diff
@@ -123,3 +127,17 @@
...
+        'id' => 'acme',
+    ]);
+
+    expect(tenancy()->initialized)->toBeFalse();
+
+    pest()->get(route('foo', ['tenant' => 'acme', 'a' => 1, 'b' => 2]));
+
+    expect(tenancy()->initialized)->toBeTrue();
+    expect(tenant('id'))->toBe('acme');
+    pest()->get(route('baz', ['a' => 1, 'b' => 2]))->assertOk(); // Assert route don't need tenant parameter
```
stancl, line 148: Also assert that the generated route is equal to what the value would be if we did specify the `tenant` argument.

tests/PathIdentificationTest.php
```diff
@@ -123,3 +127,17 @@
         ->withoutExceptionHandling()
         ->get('/acme/foo/abc/xyz');
 });
+
+test('tenant set as a default parameter for the URLs', function () {
```
stancl: Fix the grammar here please

abrardev99: Grammar looks good, but not sure about the test name overall. Renaming it to `tenant parameter is set for all routes as the default parameter once the tenancy initialized`

stancl: > Grammar looks good

It's not, it says `tenant set` rather than `tenant is set`. Also I'd change `the URLs` to `URLs` since we aren't referring to any specific URLs.

Anyway, I can rewrite this myself, just wanted to point it out.

stancl:
@abrardev99 here https://github.com/archtechx/tenancy/pull/925#discussion_r957534607 I meant specifically this https://github.com/archtechx/tenancy/pull/925/commits/def75e5e223bd811e2359d836fb3f42b4b6ac4b2#diff-c465e2d1b7f40524dcec8fa0411252f8038d38b95497741fface663cd9531620R145

abrardev99:
> @abrardev99 here [#925 (comment)](https://github.com/archtechx/tenancy/pull/925#discussion_r957534607) I meant specifically this [def75e5#diff-c465e2d1b7f40524dcec8fa0411252f8038d38b95497741fface663cd9531620R145](https://github.com/archtechx/tenancy/commit/def75e5e223bd811e2359d836fb3f42b4b6ac4b2#diff-c465e2d1b7f40524dcec8fa0411252f8038d38b95497741fface663cd9531620R145)

Nice.

PR #932, opened 2022-08-30
Reverse bootstrappers when reverting to central

emargareten:
Some of my bootstrappers are depending on previous bootstrappers but when reverting it needs to run in reverse order.
Submitting to v4 since this might be a breaking change (currently I am overriding this file).


stancl:
Ideally the bootstrappers wouldn't depend on each other, but I can see the value of this, thanks for the PR!

PR #935, opened 2022-09-01
[4.x] Make impersonation tokens require stateful guards

lukinovec:
This PR prevents creating user impersonation tokens with non-stateful guards (e.g., the `'sanctum'` guard) because user impersonation works only with stateful guards.

Relevant issue: https://github.com/stancl/tenancy-docs/issues/106


src/Exceptions/ImpersonationTokenCouldNotBeCreatedWithNonStatefulGuard.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Stancl\Tenancy\Exceptions;
+
+use Exception;
+
+class ImpersonationTokenCouldNotBeCreatedWithNonStatefulGuard extends Exception
+{
+    public function __construct(string $guardName)
+    {
+        parent::__construct("Cannot use a non-stateful auth guard ('$guardName') with user impersonation. Use a guard implementing the Illuminate\Contracts\Auth\StatefulGuard interface.");
```
stancl: I think `\` should be escaped like this:
```suggestion
        parent::__construct("Cannot use a non-stateful auth guard ('$guardName') with user impersonation. Use a guard implementing the Illuminate\\Contracts\\Auth\\StatefulGuard interface.");
```

src/Exceptions/ImpersonationTokenCouldNotBeCreatedWithNonStatefulGuard.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Exceptions;
+
+use Exception;
+
+class ImpersonationTokenCouldNotBeCreatedWithNonStatefulGuard extends Exception
```
stancl: I'd simplify the name, it's too long now. Also, all exceptions should be suffixed by `Exception`. That'd make the name even longer.

Perhaps the best way to simplify this is to remove the context/intended action, and only keep the actual error. Meaning: don't make the exception about creating impersonation tokens, instead make it simply say that a stateful guard is required.

lukinovec: https://github.com/archtechx/tenancy/pull/935/commits/f1e93a0ca0d1dfaf9f0c86a41da9a9462c32cf81 I updated the message too

stancl:
Great implementation, thanks for the PR!

PR #938, opened 2022-09-02
[4.x] Optionally delete storage after tenant deletion

### LINKED ISSUE #226: Calling Storage::fake() doesn't remove tenant directories as one would expect inline with Laravel Docs
In a test I have a function calling Storage::fake(). Calling this function makes a tenant directory after the test is run I still have the tenant folders, after running a lot of tests this can clutter things quite a bit and can be annoying to manually delete them.

#### Steps to reproduce
```
/** @test */
public function an_example_test()
{
    tenancy()->init('test.localhost');
    Storage::fake('test');
}
```

#### Expected behavior
As per laravel documentation relating to mocking Storage the following is stated;
> "By default, the fake method will delete all files in its temporary directory."

If this isn't possible it would be nice to at least have a setting in the config for;
`'delete_storage_after_tenant_deletion' => false, // delete the tenant's storage after deleting the tenant`

#### Your setup
 - Laravel version: [e.g. 6.2.0]
 - stancl/tenancy version: [2.1.0]
 - Storage driver: [DB]

lukinovec:
The PR adds a test in which I propose a way of deleting the tenant's storage after the tenant's deletion (https://github.com/archtechx/tenancy/issues/226). Apart from the addition of the config key, nothing new is really implemented – I only tested the solution, and I'll wait for @stancl's feedback (not sure if it's the best way or where's the best place to put the listener).

The storage directory would get deleted on the TenantDeleted event:

```php
Event::listen(TenantDeleted::class, function(TenantDeleted $event) {
    if (config('tenancy.filesystem.delete_storage_after_tenant_deletion')) {
        File::deleteDirectory($event->tenant->run(fn () => storage_path()));
    }
});
```

### DIFF
--- a/tests/BootstrapperTest.php
+++ b/tests/BootstrapperTest.php
@@ -22,7 +22,7 @@ use Stancl\Tenancy\Bootstrappers\CacheTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
-use Stancl\Tenancy\Events\TenantDeleted;
+use Stancl\Tenancy\Events\DeletingTenant;
 use Stancl\Tenancy\Listeners\DeleteTenantStorage;
 
 beforeEach(function () {
@@ -187,14 +187,14 @@ test('filesystem data is separated', function () {
     expect($new_storage_path)->toEqual($expected_storage_path);
 });
 
-test('tenant storage can get deleted after the tenant when TenantDeleted listens to DeleteTenantStorage', function () {
+test('tenant storage can get deleted after the tenant when DeletingTenant listens to DeleteTenantStorage', function () {
     config([
         'tenancy.bootstrappers' => [
             FilesystemTenancyBootstrapper::class,
         ],
     ]);
 
-    Event::listen(TenantDeleted::class, DeleteTenantStorage::class);
+    Event::listen(DeletingTenant::class, DeleteTenantStorage::class);
 
     tenancy()->initialize(Tenant::create());
     $tenantStoragePath = storage_path();
--- a/tests/BootstrapperTest.php
+++ b/tests/BootstrapperTest.php
@@ -23,6 +23,7 @@ use Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
 use Stancl\Tenancy\Events\TenantDeleted;
+use Stancl\Tenancy\Listeners\DeleteTenantStorage;
 
 beforeEach(function () {
     $this->mockConsoleOutput = false;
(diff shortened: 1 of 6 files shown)


tests/BootstrapperTest.php
```diff
@@ -184,6 +186,33 @@
...
+
+    Storage::fake('test');
+
+    expect(File::isDirectory(storage_path()))->toBeTrue();
+
+    Storage::put('test.txt', 'testing file');
+
+    tenant()->delete();
+
+    expect(File::isDirectory(storage_path()))->toBeFalse();
```
stancl: It feels strange to use `storage_path()` here because it depends on the tenant's environment, but the tenant got already deleted.

So while the code is not wrong, it points out that tenancy remains initialized for a tenant even after the tenant is deleted.

To make the code a bit more clear it'd be best to fetch the tenant's `storage_path()` after initialization, save it into a variable, and then make assertions with that variable (i.e. replace your current `storage_path()` references), rather than fetching it. Because calling `storage_path()` each time doesn't make it 100% clear that it's always the same value.

```suggestion
    tenancy()->initialize(Tenant::create());
    $tenantStoragePath = storage_path();

    Storage::fake('test');

    expect(File::isDirectory($tenantStoragePath))->toBeTrue();

    Storage::put('test.txt', 'testing file');

    tenant()->delete();

    expect(File::isDirectory($tenantStoragePath))->toBeFalse();
```

stancl:
And the event listener can probably go into the TenancyServiceProvider. Or do you think we could make this an optional feature, and move the logic there?

So the config option would have a note that you need to enable the feature.

Not sure if separating this into a feature is good, but it's a potential way to find a good place for the event listener.

tests/BootstrapperTest.php
lukinovec: Right, this makes more sense

lukinovec:
> And the event listener can probably go into the TenancyServiceProvider. Or do you think we could make this an optional feature, and move the logic there?
> 
> So the config option would have a note that you need to enable the feature.
> 
> Not sure if separating this into a feature is good, but it's a potential way to find a good place for the event listener.

This doesn't feel like an optional feature in the same way as user impersonation, tenant config, etc., so I'm not sure about making this a feature too.

Maybe we could put the logic in a job and add that job to TenantDeleted's job pipeline in the TenancyServiceProvider stub's `events()`? We could also remove the config key and just let the users enable/disable this by adding/removing the job from the pipeline

stancl:
Yeah, that sounds like a good idea — get rid of the config key, and just make it a commented out listener in the TenancyServiceProvider stub. That would solve the problem with putting the listener somewhere perfectly.

And yeah, it isn't an optional feature of the same "category" as the other ones.

So in the test, you'd just enable the listener using `Event::listen()` instead of setting config keys.

stancl:
I'm thinking if we even need the new events now? If the developers are already adding this by interacting with listeners, and the class can be easily extended, then the events may be unnecessary?

lukinovec:
> I'm thinking if we even need the new events now? If the developers are already adding this by interacting with listeners, and the class can be easily extended, then the events may be unnecessary?

I see, you're right. I removed the events.

stancl:
I think this should be changed (in all places) to use the deleting tenant event. It's similar to what we discussed — while hooking into TenantDeleted *does* work, it feels wrong to be [initializing](https://github.com/archtechx/tenancy/blob/1b31bdb56b2860282406b279e17a5df26e8bf104/src/Listeners/DeleteTenantStorage.php#L14) tenancy for a tenant that doesn't exist anymore.

stancl:
Can you create a TODO in the docs repo so that we document this in v4?

PR #939, opened 2022-09-05
Add docker-compose MSSQL service health checks

abrardev99:
This PR adds the missing health checks for the `mssql` service in `docker-compose.yml`. For the M1 image `azure-sql-edge`, they removed the `mssql-tools`, so checks are failing M1 for now. As discussed internally, I've commented out the code which stops containers from going up if the service is unhealthy.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,8 @@ services:
                 condition: service_healthy
             redis:
                 condition: service_healthy
+            # mssql:
+            #    condition: service_healthy
         volumes:
             - .:/var/www/html:delegated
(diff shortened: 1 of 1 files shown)


stancl:
> so checks are failing M1 for now

Just to confirm, with the commented out `condition` (like in this PR), the docker-compose will work just fine on M1?

abrardev99:
> > so checks are failing M1 for now
> 
> Just to confirm, with the commented out `condition` (like in this PR), the docker-compose will work just fine on M1?

Yes, it is working.

PR #945, opened 2022-09-12
Add skip-failing option to the Migrate command

lukinovec:
Related issue: https://github.com/stancl/tenancy-docs/issues/73


src/Commands/Migrate.php
```diff
@@ -51,8 +54,14 @@ public function handle()
...
+            try {
+                // Migrate
+                parent::handle();
+            } catch (Exception $th) {
+                if (! $this->option('skip-failing')) {
+                    throw $th;
+                }
+            }
 
             event(new DatabaseMigrated($tenant));
```
stancl: We probably want to move both of these `event()` calls inside the `try` block. That way the events (or specifically the second one) doesn't get fired when `parent::handle()` fails.

tests/CommandsTest.php
```diff
@@ -94,6 +94,23 @@
     expect(Schema::hasTable('users'))->toBeTrue();
 });
 
+test('migrate command does not throw exception if tenant does not have a database and skip-failing is passed', function() {
```
stancl: Since this test also asserts that the exception *is* thrown in the opposite case, the name should be something like "migrate command only throws exceptions if skip-failing is not passed"? So that it covers both cases.

src/Commands/Migrate.php
lukinovec: Yeah, that makes sense

tests/CommandsTest.php
lukinovec: I see, something like that will be more concise too

lukinovec:
After I moved the event dispatching to the try block, the CI tests started failing because no exception gets thrown, and that's weird because on my device, the test works fine. Why could this be happening?

stancl:
Not sure, might be related to the exact code you're using in the tests. Try debugging this.

lukinovec:
@stancl, the test is failing because if a tenant's database doesn't exist, it gets automatically created while running the `Migrate` command (if we're using MySQL or SQLite). The test also should be expecting `TenantDatabaseDoesNotExistException` (which was the exception the user that opened [the issue](https://github.com/stancl/tenancy-docs/issues/73) was getting) instead of `QueryException`.

`TenantDatabaseDoesNotExistException` only gets thrown when `app()->environment()` returns `local` (this happens in the `DatabaseTenancyBootstrapper`), but when testing, the env is `testing`, so no exception gets thrown.

So to test this properly, we'd have to be in the `local` env instead of `testing`. Or maybe there's some way to raise an exception during the `Migrate` command other than the DB not existing and the environment being `local`.

stancl:
I don't think the DB should be created in `Migrate` when using MySQL. The migrate job fails or doesn't run if the DB doesn't exist. Could you link to the relevant code?

And could you link to the bit that checks for the `local` environment? I think we could edit it to work with `testing` too. That's better than changing the environment away from `testing` in the tests.

lukinovec:
> I don't think the DB should be created in `Migrate` when using MySQL. The migrate job fails or doesn't run if the DB doesn't exist. Could you link to the relevant code?

The parent MigrateCommand class does that
https://github.com/illuminate/database/blob/bd88721cabecee43c9ebc4398a9fcedea7f22092/Console/Migrations/MigrateCommand.php#L136

> And could you link to the bit that checks for the `local` environment? I think we could edit it to work with `testing` too. That's better than changing the environment away from `testing` in the tests.

Right, that could work
https://github.com/archtechx/tenancy/blob/3.x/src/Bootstrappers/DatabaseTenancyBootstrapper.php#L28

stancl:
I'm pretty confused by the part where it creates the DB because that'd defeat the point of this PR. You should look into that more in depth and try other ways of reproducing a non-existent DB in the test suite.

lukinovec:
> I'm pretty confused by the part where it creates the DB because that'd defeat the point of this PR.

Yeah, I had trouble understanding the point of the requested feature too, but when running the app in the local environment, it doesn't get to the "create missing DB" part. So when app env is `local` and `tenants:migrate` gets called, the DB is not created + `TenantDatabaseDoesNotExistException` is thrown at the point of calling `tenancy()->runForMultiple()`

> I think we could edit it to work with testing too

I tried to add `|| app()->environment('testing')` as you suggested. Turns out that `tenancy()->runForMultiple()` throws the exception, so the current try/catch is not right (the whole `tenancy()->runForMultiple()` block should be in `try` and we should be catching `TenantDatabaseDoesNotExistException`.

So to correct things, I'd:

- add `|| app()->environment('testing')` to the DatabaseTenancyBootstrapper `bootstrap()` check
- put the whole `tenancy()->runForMultiple()` block in `try` and catch `TenantDatabaseDoesNotExistException`

stancl:
> but when running the app in the local environment, it doesn't get to the "create missing DB" part

Ah, does it only do that in tests?

lukinovec:
> Ah, does it only do that in tests?

It does that every time the app environment is `local`, so no, not only in tests (I mean the app doesn't get to create the missing DB because the database bootstrapper throws the exception when in `local` env earlier)

EDIT: [I made changes to make this work](https://github.com/archtechx/tenancy/compare/5c7a2c26ad005dbe374bfc0279ae99e4ae7b5872..add-skip-failing-options-to-migrate), but [that broke some other tests](https://github.com/archtechx/tenancy/actions/runs/3173969283/jobs/5170182190). That's because the DatabaseTenancyBootstrapper is used in some of the tests, but the database is not really being created anywhere in the tests (adding `Event::listen(TenantCreated::class, JobPipeline::make([CreateDatabase::class])->send(..))` to `beforeEach` of these test files should fix that)

Some of the tests don't use DatabaseTenancyBootstrapper and they started failing too which is weird. I'm looking into this now

EDIT2: So for instance, AutomaticModeTest doesn't add DatabaseTenancyBootstrapper to the bootstrappers beforeEach, but some tests are still failing. I checked, and the tests that weren't failing were setting `config('tenancy.bootstrappers')`, whereas the ones that were failing weren't setting that key, and when `dd`ing `tenancy()->getBootstrappers()` just before the exception is thrown, I see that there are all bootstrappers from the config _file_.

I thought about solving the skip-failing exception issue some other way, but I think the current solution might be the best, and we should address the issue with the missing tenant databases. @stancl what do you think? I think I would just create the databases, ideally on `TenantCreated`, but I see some special logic where that would create the DB twice, e.g. `withTenantDatabases()` in `QueueTest` adds a JobPipeline with `CreateDatabase` to `TenantCreated`. This method could be made into a helper, and we could use it where needed.

stancl:
Note from our DMs — the change in `DatabaseTenancyBootstrapper` seems good but it has some implications for tests. Before, if you were using the `DatabaseTenancyBootstrapper` and you created a tenant and initialized tenancy **before creating the tenant's database**, no exception was thrown.

Now, an exception is thrown. Therefore some tests of people using this package may start failing. However, that seems like a positive thing since the tests are now more accurate.

Not throwing the exception had some benefits — you'd be able to create a tenant, initialize tenancy, and make various assertions **as long as you weren't using the tenant's DB**. An exception would be thrown only when you tried to use the tenant's DB for something.

Now the exception is thrown upfront, so your tests should be more robust and actually _create tenants properly_ — with all of their dependencies. Or use simplified testing Tenants for some tests that don't need a DB.

lukinovec:
@stancl I resolved the conflicts and assigned the PR to you

src/Commands/Migrate.php
```diff
@@ -42,17 +46,24 @@ public function handle(): int
...
-            event(new DatabaseMigrated($tenant));
-        });
+            return 0;
+        } catch (TenantDatabaseDoesNotExistException|QueryException $th) {
+            if (! $this->option('skip-failing')) {
+                throw $th;
+            }
 
-        return 0;
+            return 1;
```
lukinovec: Looking at the TenantDump command, I see that the `Command::SUCCESS` (0) and `Command::FAILURE` (1) constants exist. Maybe we could return these instead of just 0 and 1? That would be prettier imo @stancl

stancl: `0` and `1` return codes in commands are super standard, even Laravel uses them, so I think we'll stick with that. I didn't know about the constants so that's cool that they're available though in this case we'll stick with the simple solution.

lukinovec: Okay, I see

stancl:
@lukinovec Can you resolve the conflicts here?

lukinovec:
@stancl, I resolved the conflicts, there are some failing tests, but these are the same as the ones that were failing before the merge, so that's fine.

src/Commands/Migrate.php
```diff
@@ -42,17 +46,24 @@ public function handle(): int
...
```
stancl: Wouldn't this basically terminate at the first failing tenant and then not continue? Even if the exception is suppressed.

lukinovec: I see, yeah, makes sense that would happen. I'll see if I can add a test for this and make it work correctly

stancl: The test looks good. I think the `try {}` should be inside the loop basically, not outside.

lukinovec: I tried that, but since `runForMultiple` throws the exception ([when initializing tenancy](https://github.com/archtechx/tenancy/blob/3.x/src/Tenancy.php#L159) I think), `try ()` inside the closure doesn't catch the error

lukinovec: I think the simplest way to make this work would be calling `$tenant->run()` one by one (the test passes with that approach – [pushed the change](https://github.com/archtechx/tenancy/pull/945/commits/bfd2669fda3ad944dac3669e40aa1cc0460c89fa)). Would that hinder the performance a lot?
```php
foreach ($tenants as $tenant) {
    try {
        $tenant->run(function()) {
            // Code...
        }};
    } catch {
        if (! $this->option('skip-failing')) {
            throw $th;
        }
    }
}
```
I don't think there's a simple way to go about this using `runForMultiple`, so if this isn't much slower, I'd use this

stancl: `runForMultiple()` is a tiny bit more efficient than `foreach()` + `run()`, but the difference is negligible so this doesn't matter in a console command.

PR #946, opened 2022-09-13
[4.x] Use a dedicated DB connection for creating/deleting tenant databases

### LINKED ISSUE #515: When creating a new tenant that should have a mysql db on a different host/server it fails
#### Describe the bug
I created a console command to create tenants. For our use case we want the tenant dbs hosted on different servers. So i added the ```tenancy_db_host```. But this doesn't seem to do anything. If i hardcode the host in the template, it works. 

#### Steps to reproduce
In ```config/database.php``` i have the 'template', here is a snippet. The env vars are there just as fallback.

```        
'tenant_db' => [
            'driver' => 'mysql',
            'host' => env('DB_HOST', '127.0.0.1'),
            'port' => env('DB_PORT', '3306'),
            ....
```

For the creation of the tenant i use this snippet:
```
        $dbSettings['tenancy_db_connection'] = 'tenant_db';
        $dbSettings['tenancy_db_host'] = $host;

        if ($dbuser) {
            $dbSettings['tenancy_db_username'] = $dbuser;
        }
        if ($dbpasswd) {
            $dbSettings['tenancy_db_password'] = $dbpasswd;
        }

        $tenant = Tenant::create($dbSettings);
```

#### Expected behavior
I would expect that the host variable in the template would be overwritten with the ```tenancy_db_host``` key.

#### Your setup
 - Laravel version: 7.28.4
 - stancl/tenancy version: v3.2.1
 - Storage driver: MySQL

abrardev99:
## Problem
When a tenant has its own database config (using the tenancy_db_... properties), the database manager still tries to create the DB on the central/template connection. The problem described in the related issue https://github.com/archtechx/tenancy/issues/515 is that `tenancy_db_host` is specified in the tenant config, but the package still uses the host from the template connection. 

## Solution
The solution is pretty straightforward. The main motivation is coming from this comment https://github.com/archtechx/tenancy/pull/656#issuecomment-841803111. The current `manager()` method got modified. It's now creating the new host connection. A **host connection** is a new temporary connection created by **replacing** the template config with the tenant's config.

**Replacing** means we override the template connection with the values from the tenant config. This way, We are able to use the tenant configs, e.g., `tenancy_db_*` properties, to build the final connection. 

### Usage

- Providing username and password through tenant config: 
```php
Tenant::create([
    'tenancy_db_username' => $username,
    'tenancy_db_password' => $password,
]);
```

- Providing the host 
```php
Tenant::create([
    'tenancy_db_host' => 'mysql2',
]);
```

> Note: You can't use the username/password when using the `PermissionControllerMySQLManager`. 

Will not work ❌
```php
Tenant::create([
    'tenancy_db_username' => $username,
    'tenancy_db_password' => $password,
]);
```

Will work ✅
```php
Tenant::create([
    'tenancy_db_connection' => 'custom_connection'
]);
```
`custom_connection` is a connection array in `database.connections` which is using the username/password. 

In Summary, there are two cases: 
1. Using credentials with MySQL manager 

It's simple and can be used like this 
```php
Tenant::create([
    'tenancy_db_username' => 'root',
    'tenancy_db_password' => 'password',
]);
```

2. Using credentials with permission controller MySQL manager 
We can't use the above approach. So we will be using the "connection" attribute. **But if "tenancy_db_username" is provided, the permission manager creates it for that specific database. 

Major diff between the two: `username` in 1 already exists and will be used to create the tenant database and `username` in 2 does not exist and will be created by the manager.**


abrardev99:
I saw a todo related to a name: https://github.com/archtechx/tenancy/blob/master/src/Database/TenantDatabaseManagers/TenantDatabaseManager.php#L9-L12
image

I think we can rename it to `TenantDatabaseManagerContract`.

stancl:
Two things to do here:
1. Describe the changes in detail in the PR description — what this adds, how the solution works
2. Based on our convo on BC, you can revert the latest commit

abrardev99:
> Two things to do here:
> 
> 1. Describe the changes in detail in the PR description — what this adds, how the solution works
> 2. Based on our convo on BC, you can revert the latest commit

Done.

assets/config.php
```diff
@@ -47,6 +47,11 @@
          */
         'template_tenant_connection' => null,
 
+        /**
+         * Tenant host connection name used to create a temporary connection for creating, deleting database
```
stancl:
```suggestion
         * The name of the temporary connection used for creating and deleting tenant databases.
```

src/Database/DatabaseConfig.php
```diff
@@ -100,6 +101,11 @@ public function getTemplateConnectionName(): string
             ?? config('tenancy.database.central_connection');
     }
 
+    public function getTenantHostConnectionName(): ?string
+    {
+        return config('tenancy.database.tenant_host_connection_name');
+    }
```
stancl: When would this be null? And what's the behavior if it's null?

src/Database/DatabaseConfig.php
```diff
@@ -158,4 +201,27 @@ public function manager(): Contracts\TenantDatabaseManager
 
         return $databaseManager;
     }
+
+    /** Get the TenantDatabaseManager for this tenant's connection. */
+    public function hostManager(): Contracts\TenantDatabaseManager
+    {
+        $this->purgeHostConnection();
```
stancl: Add comments explaining what is being done here, at all of the important/unclear lines.

In general, a method used for *fetching something* (here that is fetching the host manager), shouldn't have many side effects ideally.

stancl: Also add a `// todo` with a note to rename this method, I think the name isn't ideal but can't come up with a better one right now.

tests/TenantDatabaseManagerTest.php
```diff
@@ -100,20 +100,22 @@
 
     $tenant = Tenant::create();
 
-    expect(array_keys(app('db')->getConnections()))->toBe(['central']);
+    expect(array_keys(app('db')->getConnections()))->not()->toContain('tenant');
```
stancl: Can you explain this change?

stancl: And the changes below as well — using `contain` instead of strict `toBe()` assertions.

src/Database/DatabaseConfig.php
abrardev99: > When would this be null? 

This can be null if developer does provide value in config. 

> And what's the behavior if it's null?

We are not handling null behavior yet. If config returns null, we should provide a default value. 

I think I got your point. This can't be null. The config key must have value. Leaving the key null is not a good option here, IMO. So I'm removing the nullable option from here.

tests/TenantDatabaseManagerTest.php
abrardev99: > Can you explain this change?

So, We added the host manager in this PR, which internally creates the `tenant_host` connection. 

As We are creating tenant-right above line # 101. Tenant creation creates a host connection already. So now there are two connections in the array. 
image

That's why I tweaked the assertions a little bit. Now assertions are specific to the tenant. I mean test says `the tenant connection is fully removed`, so assertions are checking if the tenant connection exists, not exists, etc. 

> using contain instead of strict toBe() assertions.

We can use toBe(), but we will be checking two connections (as in the above image), so I think checking contains is good because we are ensuring the tenant connection exists.

src/Database/DatabaseConfig.php
abrardev99: Added the todo and possible comments. TBH, I don't see unclear code here, almost every line is pretty clear what it is doing. 

> In general, a method used for fetching something (here that is fetching the host manager), shouldn't have many side effects, ideally.

What do you mean by "shouldn't have many side effects"? What side effects this method has? This method actually creates and returns the host manager, not just **fetching the host manager**.

stancl: > What side effects this method has?

What's the `$this->purgeHostConnection()` call for? Is that for purging the connection used specifically by the returned manager?

At first look it seems like it's making a change to some general connection, but if it's specifically for the connection used by the returned hostManager then that's fine. I'd just add some comments to make it a bit more clear at first look.

> This method actually creates and returns the host manager, not just fetching the host manager.

It doesn't matter *how* it does it, my point is that it's a method for *getting something*. It creates the manager and returns it.

That's what I mean by fetching. I mean that it's not an action that does some side effects, instead it's a method for fetching some object. But as I mentioned above it's possible that it doesn't have any general side effects, so it might be fine.

So the only change I'd make here would be adding those comments. The code may be clear when you take a deep look at it, but we generally aim for code being immediately obvious upon seeing it. 

This method has a lot of steps for a method that just creates and returns an object, so the steps which:
1. Purge the host connection
2. Manipulate config
3. Access the driver

Should have simple comments explaining what we're doing and why we're doing it.

tests/TenantDatabaseManagerTest.php
stancl: Is the point of the `toContain()` assertion to check that it's `tenant` or `tenant_host_connection`? If so, I'd definitely change that to be more strict.

The name of `tenant_host_connection` is customizable, plus connection simply having "tenant" in it doesn't necessarily mean anything.

So I'd just make the assertions more strict.

abrardev99: > So I'd just make the assertions more strict.

Sure. But I don't see any benefit of more strict assertions here. We were testing these points in this test. 

- assert `tenant` connection does not exist.
- initialize tenancy  - asserts that a `tenant` connection is created 
- end tenancy - assert `tenant` connection got removed. 

And the current assertions are doing the same. Does the test do anything differently? If so, please explain we will change the assertions. 

> plus connection simply having "tenant" in it doesn't necessarily mean anything.

It does mean that the connections array has a `tenant` connection along with other connections.

src/Database/DatabaseConfig.php
abrardev99: > What's the $this->purgeHostConnection() call for? Is that for purging the connection used specifically by the returned manager?

Yes, it is. I thought the method name `purgeHostConnection` is pretty much clear that we are purging the host connection, not any connection. Anyway, adding the comments is a good idea.

tests/TenantDatabaseManagerTest.php
stancl: > But I don't see any benefit of more strict assertions here

It's what I wrote here:

> The name of `tenant_host_connection` is customizable, plus connection simply having "tenant" in it doesn't necessarily mean anything.

We want to assert **actual checks that the connection is a certain value**, not that it contains a substring.

> It does mean that the connections array has a tenant connection along with other connections.

How? If I create a central connection and name it `tenants` (because it stores the list of tenants), then that's a central connection that passes your `toContain()` check.

Just make direct assertions about what the values equal. Don't make indirect assertions about the values containing some substring.

abrardev99: >  Just make direct assertions about what the values equal. Don't make indirect assertions about the values containing some substring.

Sure.

src/Database/DatabaseConfig.php
abrardev99: Resolving this because comments and todos were added.

src/Database/DatabaseConfig.php
```diff
@@ -114,6 +120,43 @@ public function connection(): array
...
+        $template = $this->getTemplateConnectionName();
+        $templateConnection = config("database.connections.{$template}");
+
+        if ($this->manager() instanceof Contracts\ManagesDatabaseUsers) {
+            // We don't need username and password for database creation/deletion
+            // Username and password will be saved in Tenant's config
+            // and will be used for tenancy db connection
+            unset($config['username']);
+            unset($config['password']);
+        }
```
stancl: Is this code needed, or does it just simplify the `$config` array?

From the `We don't need username` it feels as if you were just simplifying the data, but I assume there's a better reason for why this is being unset?

src/Database/DatabaseConfig.php
```diff
@@ -158,4 +201,30 @@ public function manager(): Contracts\TenantDatabaseManager
 
         return $databaseManager;
     }
+
+    /** Get the TenantDatabaseManager for this tenant's connection. */
+    public function hostManager(): Contracts\TenantDatabaseManager
+    {
+        // DB cache the previous connection configurations and changing the connection configuration does not work
+        // So we purge the previous host connection before changing it
```
stancl:
```suggestion
        // Laravel caches the previous PDO connection, so we purge it to be able to change the connection details
```

src/Database/DatabaseConfig.php
```diff
@@ -114,6 +120,43 @@ public function connection(): array
...
+     * Purge host database connection.
+     */
+    public function purgeHostConnection(): void
+    {
+        $tenantHostConnectionName = $this->getTenantHostConnectionName();
+
+        if (config("database.connections.{$tenantHostConnectionName}")) {
+            DB::purge($tenantHostConnectionName);
+            config(["database.connections.{$tenantHostConnectionName}" => null]);
+        }
```
stancl: I think there should be a better check here. Just because the config value exists doesn't mean that Laravel has the PDO connection cached. See at the method which purges the `tenant` connection.

Also it's possible that we can completely remove the check and just always reset the config and the connection cache.

And I believe we have `DatabaseManager` injected so we shouldn't be using the `DB` facade here.

src/Database/DatabaseConfig.php
```diff
@@ -158,4 +201,30 @@ public function manager(): Contracts\TenantDatabaseManager
...
     }
+
+    /** Get the TenantDatabaseManager for this tenant's connection. */
+    public function hostManager(): Contracts\TenantDatabaseManager
+    {
+        // DB cache the previous connection configurations and changing the connection configuration does not work
+        // So we purge the previous host connection before changing it
+        $this->purgeHostConnection(); // todo come up with a better name
+
+        // Set the host connection configuration array fetched from the tenant's config and manipulated to form a connection
```
stancl: What are you trying to say here? The comment is confusing. Either rewrite it or share what the point is meant to be and I'll write it for you.

tests/TenantDatabaseManagerTest.php
```diff
@@ -99,20 +99,21 @@
     Event::listen(TenancyEnded::class, RevertToCentralContext::class);
 
     $tenant = Tenant::create();
+    $tenantHostConnectionName = config('tenancy.database.tenant_host_connection_name');
```
stancl: You don't need the variable, you can hardcode the string. Just like `central` is hardcoded.

tests/TenantDatabaseManagerTest.php
```diff
@@ -222,6 +223,159 @@
 
     $manager->setConnection('mysql2');
     expect($manager->databaseExists($name))->toBeTrue();
+
+    $manager->deleteDatabase($tenant);
+    expect($manager->databaseExists($name))->toBeFalse();
+});
+
+test('tenant database user can be created when using the permission controller MySQL manager', function () {
```
stancl: Doesn't the previous test already do this?

Also, the previous test should likely be fixed. It said `tenant database can be created on a foreign server` **but this behavior was broken so the test doesn't actually verify that**. Therefore the code of the test should likely be rewritten.

tests/TenantDatabaseManagerTest.php
```diff
@@ -222,6 +223,159 @@
...
+    $dbUser = $tenant->tenancy_db_username;
+
+    /** @var PermissionControlledMySQLDatabaseManager $manager */
+    $manager = $tenant->database()->manager();
+    $manager->setConnection('mysql');
+
+    expect($manager->userExists($dbUser))->toBeTrue();
+});
+
+test('tenant database can be created on template tenant connection', function () {
```
stancl: What is the exact point of this test?

tests/TenantDatabaseManagerTest.php
```diff
@@ -222,6 +223,159 @@
...
+    // central was default connection so we are making sure
+    // that database did not create on central
+    $manager->setConnection('central');
+    expect($manager->databaseExists($name))->toBeFalse();
+
+    $manager->setConnection('mysql2');
+    expect($manager->databaseExists($name))->toBeTrue();
+});
+
+test('tenant database can be created on a foreign server by using the host from tenant config', function () {
```
stancl: What's the point of this test as well? It's not very clear to me what we're trying to assert here.

stancl: And the next test as well.

tests/TenantDatabaseManagerTest.php
abrardev99: This test's point is to ensure that tenant connection can be established successfully when the developer provides the `tenant_template_connection`. 

I don't exactly remember why I added this, but I wanted to make sure `tenant_template_connection` works as expected after adding the hostConnection logic.

tests/TenantDatabaseManagerTest.php
abrardev99: > What's the point of this test as well?

This test ensures that tenant database connection can be established successfully when tenantConfig has a different host. For example, general host is `mysql`, but the developer wants to use a different host for different tenants. So he provides the host in tenantConfig. 
```php
Tenant::create([
        'tenancy_db_host' => 'mysql2',
    ]);
```
So the test ensures the connection is established using the `mysql2` host and other properties are the same as provided in the connection array. 

> And the next test as well.
`tenant database can be created on a foreign server by using the username and password from tenant config`

This test is same as above but it ensures that the tenant database connection can be established successfully from the tenantConfig username and password.

src/Database/DatabaseConfig.php
abrardev99: > I think there should be a better check here.

I see. So I checked the `purge` method. It is disconnecting the `PDO` and then removing the connection of The active connection instances array. `purge` method looks like this. 
```php
public function purge($name = null)
    {
        $name = $name ?: $this->getDefaultConnection();

        $this->disconnect($name);

        unset($this->connections[$name]);
    }
```

 
So we can modify our logic. 
```php
if (array_key_exists($tenantHostConnectionName, $databaseManager->getConnections())) {
        $databaseManager->purge($tenantHostConnectionName);
    }
```
which is working fine. 

> Also it's possible that we can completely remove the check and just always reset the config and the connection cache.

And I tried this. If we completely remove the connection, tests are still passing. But my concern is `purge` method check if the given connection name is null, it purges the default connection. And the host connection can be null, mostly when hitting the method the first time in the request life cycle. 

So I think the `if` check is better since we want to purge only the host connection. 


> And I believe we have DatabaseManager injected, so we shouldn't be using the DB facade here.

No, we don't have. We can inject this or keep using the facade.

src/Database/DatabaseConfig.php
abrardev99: I'm trying to explain that we get the modified configuration array and set the host connection configuration. You can write a better one, but I don't think we need comments here. The code is straightforward. We are pushing the `hostConnection` config in the `connections` array. `hostConnection` does require some good comments because that's the method doing all of the modifications in the config. And we have one review related to this method.

src/Database/DatabaseConfig.php
abrardev99: > Is this code needed, or does it just simplify the $config array?

This code is needed because if we don't remove the username and password from the config, an error occurred while establishing a connection like `SQLSTATE[HY000] [1045] Access denied for user '3qVdsacitd420N25'@'172.24.0.7' (using password: YES) (SQL: SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = 'foo1k7L3u4g')`. 
An error occurred only for the tests using the `PermissionControlledMySQLDatabaseManager`.  And that's the reason I wrote test for `PermissionControlledMySQLDatabaseManager` you asked why it's needed https://github.com/archtechx/tenancy/pull/946#discussion_r981115857. 

I'm trying but unable to come up with better comments.

tests/TenantDatabaseManagerTest.php
abrardev99: This review and https://github.com/archtechx/tenancy/pull/946#discussion_r982287654 are kinda connected. I explained there why I added this test.

src/Database/DatabaseConfig.php
stancl: > And the host connection can be null, mostly when hitting the method the first time in the request life cycle.

Can you expand on this? I think it shouldn't be null since the method we have for returning the name of the connection has a `: string` typehint

> No, we don't have. We can inject this or keep using the facade.

Ah I see, this is the `DatabaseConfig`. Makes sense, facade is fine then.

src/Database/DatabaseConfig.php
stancl: I see, so the problem is that we don't want to use the tenant's DB credentials since they don't have the permissions to create databases? And I assume these values (`username` and `password`) get filled with the "main" credentials? Where does that happen?

> I'm trying but unable to come up with better comments.

It should just say why we're doing this, not that "We don't need" the username. Not needing the username is different from *having* to remove the username for this to work.

src/Database/DatabaseConfig.php
abrardev99: > Can you expand on this? I think it shouldn't be null since the method we have for returning the name of the connection has a : string typehint

I mean getting the connection from the config `config("database.connections.{$tenantHostConnectionName}")` can be null at first time. 
For example if we perform this `$databaseManager->purge(config("database.connections.{$tenantHostConnectionName}"));` without `if` check, config will return the `null` and `purge` will purge the default connection since it received `null`.

src/Database/DatabaseConfig.php
abrardev99: > And I assume these values (username and password) get filled with the "main" credentials? Where does that happen?

It got filled in `makeCredentials()` method. https://github.com/archtechx/tenancy/blob/f97651e16ba300c5e4671031a19c101f1f0fae42/src/Database/DatabaseConfig.php#L88-L91
And this method got called in `CreateDatabase` job. https://github.com/archtechx/tenancy/blob/cd6e989c09981800aef439471d56e9654b1bbbf1/src/Jobs/CreateDatabase.php#L36

stancl: Okay, so this set the username and password that you're later unsetting? But could you expand on what I asked?

> And I assume these values (username and password) get filled with the "main" credentials? Where does that happen?

I'm asking about what happens **after** you unset these values, not before.

abrardev99: Ah Ok. 

> values (username and password) get filled with the "main" credentials? Where does that happen?

We are removing the username and password from the **tenant's config**, and the template connection still has "main" credentials. We are doing an array replace thing at the end and that's how main credentials got filled. 
https://github.com/archtechx/tenancy/blob/7f6166d49a6b07f8b4c4831f12a6ddc15c305b88/src/Database/DatabaseConfig.php#L143

stancl: > and the template connection still has "main" credentials. We are doing an array replace thing at the end and that's how main credentials got filled.

Okay, so the template connection always has credentials, and what you're doing in the `unset()` is removing credentials **from the config** that later gets merged into the template connection? That way we prevent the credentials in the config from overriding the template connection credentials?

And where do the credentials for `$templateConnection` come from?

abrardev99: > And where do the credentials for $templateConnection come from?

it is coming from the `config("database.connections.{$template}");` https://github.com/archtechx/tenancy/blob/7f6166d49a6b07f8b4c4831f12a6ddc15c305b88/src/Database/DatabaseConfig.php#L130-L131. `$template` generated from another method https://github.com/archtechx/tenancy/blob/7f6166d49a6b07f8b4c4831f12a6ddc15c305b88/src/Database/DatabaseConfig.php#L98-L103

src/Database/DatabaseConfig.php
stancl: What do you mean specifically by _"mostly when hitting the method the first time in the request life cycle"_?

And I think the host connection should **only** be used for logic related to creating databases? So I'm not sure why it'd be null or not null at different points in the request lifecycle. I don't think it would be even used.

stancl:
@abrardev99 I responded to a few reviews. In this PR, try to improve the tests based on what we discussed on BC recently. Then I'll review the PR again.

src/Database/DatabaseConfig.php
abrardev99: > What do you mean specifically by "mostly when hitting the method the first time in the request life cycle"?

As you said "host connection should only be used for logic related to creating databases" that's correct. But CreateDatabase fetches hostManager twice. One for ensuring a database can be created and one for actually creating a database. So when it fetches the host manager first time, it tries to purge the connection but the connection is not created yet. 

But I see that Laravel's underline method is doing the same. eg purging connection key only it is exists in connections array.  
```php
if (isset($this->connections[$name = $name ?: $this->getDefaultConnection()])) {
    $this->connections[$name]->disconnect();
}
```
So checking on our side doesn't make sense, and I'm removing it.

tests/TenantDatabaseManagerTest.php
abrardev99: > Doesn't the previous test already do this?

I see the previous test (test with dataset) is doing the same. So I'm removing it.

tests/TenantDatabaseManagerTest.php
abrardev99: I'm thinking of removing this test as well. Because there are two other tests that are making use of `tenant_template_connection` then there is no point in this test. 

Edit: Removed the test.

stancl:
Is the label because this is still WIP?

abrardev99:
> Is the label because this is still WIP?

Yes, I'm making some changes to the tests. I'll remove the label once I've done.

abrardev99:
@stancl I gave it a final review now. I replied to all of the reviews. I removed two tests because other tests are kind of doing the same. Please check open reviews.

tests/TenantDatabaseManagerTest.php
stancl: So this can be resolved now?

tests/TenantDatabaseManagerTest.php
stancl: Can this be resolved now?

tests/TenantDatabaseManagerTest.php
abrardev99: Yes. Going to resolve it.

tests/TenantDatabaseManagerTest.php
abrardev99: Yes.

tests/TenantDatabaseManagerTest.php
stancl: Marking this as resolved now.

src/Database/DatabaseConfig.php
stancl: So everything in this thread is resolved now?

tests/TenantDatabaseManagerTest.php
stancl: Okay, so the `tenancy_db_host` overrides details in the template host connection, right?

stancl: If yes, then I realized that there's one thing we may need to resolve. I realized this a few days ago.

The `PermissionControlled...` DB manager uses `tenancy_db_username` and `tenancy_db_password` **as details for the user _that will be created_**.

But what if we wanted to specify details for the host connection, including:
- the hostname
- the username (of the user that will be used **for** creating the tenant's DB user)
- the password (also of the main user)

How could we do that?

src/Database/DatabaseConfig.php
```diff
@@ -114,6 +121,37 @@ public function connection(): array
...
+        return array_replace($templateConnection, $config);
+    }
+
+    /**
+     * Purge host database connection.
+     */
+    public function purgeHostConnection(): void
+    {
+        $tenantHostConnectionName = $this->getTenantHostConnectionName();
+        DB::purge($tenantHostConnectionName);
```
stancl:
```suggestion
        DB::purge($this->getTenantHostConnectionName());
```

Also I'd add a short comment explaining what this is for (similar to what we discussed in the reviews). Should say something along the lines of: this is used for cleaning up the connection before creating it for the current tenant.

tests/TenantDatabaseManagerTest.php
```diff
@@ -220,6 +220,91 @@
...
     expect($manager->databaseExists($name))->toBeTrue();
+
+    $manager->deleteDatabase($tenant);
+    expect($manager->databaseExists($name))->toBeFalse();
+});
+
+test('tenant database can be created on a foreign server by using the host from tenant config', function () {
+    config([
+        'tenancy.database.managers.mysql' => MySQLDatabaseManager::class,
+        'tenancy.database.template_tenant_connection' => 'mysql',
```
stancl: Why are you configuring `mysql` instead of `mysql2` here?

stancl: Is it that the tenant's `tenancy_db_host` will override the hostname here? And therefore use the `mysql2` host instead of what's specified in `template_tenant_connection`?

tests/TenantDatabaseManagerTest.php
```diff
@@ -220,6 +220,91 @@
...
+
+    $manager->setConnection('mysql2');
+    expect($manager->databaseExists($name))->toBeTrue();
+});
+
+test('tenant database can be created on a foreign server by using the username and password from tenant config', function () {
+    config([
+        'tenancy.database.managers.mysql' => MySQLDatabaseManager::class,
+        'tenancy.database.template_tenant_connection' => 'mysql',
+        'database.connections.mysql' => [
```
stancl: Do we need to configure `database.connections.mysql`?

tests/TenantDatabaseManagerTest.php
abrardev99: > Is it that the tenant's tenancy_db_host will override the hostname here? And therefore use the mysql2 host instead of what's specified in template_tenant_connection?

Yes.

tests/TenantDatabaseManagerTest.php
abrardev99: > Do we need to configure database.connections.mysql?

I tested it. I removed the database.connections.mysql, and it's working fine. I thought we had to provide the connection. That means wherever we are using `mysql` I can remove the config, right?

src/Database/DatabaseConfig.php
abrardev99: > So everything in this thread is resolved now?

Yes.

src/Database/DatabaseConfig.php
abrardev99: Added the comment.

tests/TenantDatabaseManagerTest.php
stancl: Yes, I think these are already configured in `TestCase` or even by default by Laravel.

tests/TenantDatabaseManagerTest.php
stancl: I see. Add a comment at the end of line 231 saying `// This will be overridden by tenancy_db_host`.

And similar to the other review, check if removing the `mysql2` connection details works.

abrardev99: > Add a comment at the end of line 231

Done

> And similar to the other review, check if removing the mysql2 connection details works.

Removing the `mysql2` connection config does not work.

tests/TenantDatabaseManagerTest.php
abrardev99: > Okay, so the tenancy_db_host overrides details in the template host connection, right?

Yes right. 

> But what if we wanted to specify details for the host connection, including:
the hostname
the username (of the user that will be used for creating the tenant's DB user)
the password (also of the main user)

I'm afraid I did not understand your point. Do you mean how to specify the hostname along with username and password?

abrardev99: The review is resolved here https://github.com/archtechx/tenancy/pull/946 after discussing it internally.

tests/TenantDatabaseManagerTest.php
stancl: I see, seems like that's something we only used in `TenantDatabaseManagerTest` (in master). It's fine to configure it here then.

tests/TenantDatabaseManagerTest.php
```diff
@@ -220,6 +221,126 @@
...
+    ]);
+
+    /** @var PermissionControlledMySQLDatabaseManager $manager */
+    $manager = $tenant->database()->manager();
+
+    expect($manager->userExists($username))->toBeTrue();
+    expect($manager->databaseExists($name))->toBeTrue();
+});
+
+test('tenant database can be created by using the username and password from tenant config', function () {
```
stancl, line 333: @abrardev99 What does this test exactly? What role do the username & password have here?

tests/TenantDatabaseManagerTest.php
```diff
@@ -220,6 +221,126 @@
...
+        'tenancy_db_connection' => 'mysql2',
+        'tenancy_db_username' => $username = 'user_for_new_db' . Str::random(4),
+        'tenancy_db_password' => Str::random(8),
+    ]);
+
+    /** @var PermissionControlledMySQLDatabaseManager $manager */
+    $manager = $tenant->database()->manager();
+
+    expect($manager->userExists($username))->toBeTrue();
+    expect($manager->databaseExists($name))->toBeTrue();
```
stancl: Just to confirm: this test basically asserts that *even though we're using the credentials from lines 279-280 for the host connection*, a user will be created with the name of `$username`?

Or are you checking that another user can be used **for** creating the database (i.e. for the host connection)?

I'm a bit confused because I think we'd want to check that the `PermissionControlledMySQLDatabaseManager` creates the users correctly, but your assertion on line 319 only checks that the user that **you** created on line 298 exists?

tests/TenantDatabaseManagerTest.php
abrardev99: Ok. Resolving the review now.

tests/TenantDatabaseManagerTest.php
abrardev99: > What does this test exactly? What role does the username & password have here?

The main purpose of this test is to verify that we can use separate usernames and passwords through the tenant config key. Eg, I have the same host, but I want to use different credentials. 

But now looking at the test, I'm using the same username and password as the `mysql` config has. So the test is kind of useless.

stancl: > verify that we can use separate usernames and passwords through the tenant config key

Yeah but are we actually checking anything? I just see that you're passing these credentials, but then aren't making any assertions about them.

tests/TenantDatabaseManagerTest.php
abrardev99: > Just to confirm: this test basically asserts that even though we're using the credentials from lines 279-280 for the host connection, a user will be created with the name of $username?
Or are you checking that another user can be used for creating the database (i.e. for the host connection)?

We are asserting even though we are using different credentials for the host connection, it still creates the credentials for permission controlled manager. 

>  but your assertion on line 319 only checks that the user you created on line 298 exists?

Assertions basically check if the username given to the tenant config (on line 312) is created. Maybe you are confused with the variable `$username`, because I'm reusing the variable at line 312. 

> I'm a bit confused because I think we'd want to check that the PermissionControlledMySQLDatabaseManager creates the users correctly

We ARE asserting that PermissionControlledMySQLDatabaseManager creates the users correctly.

tests/TenantDatabaseManagerTest.php
abrardev99: >Yeah but are we actually checking anything? I just see that you're passing these credentials, but then aren't making any assertions about them.

I'm asserting that the database is created. That means we are using the given credentials.

tests/TenantDatabaseManagerTest.php
stancl: I see, you're using inline assignment and overriding the variable. That makes the test very confusing, use different variables for the different users.

tests/TenantDatabaseManagerTest.php
stancl: The name is: "can be created **by using the username and password** from tenant config"

How does the highlighted part relate to what is actually tested?

abrardev99: > How does the highlighted part relate to what is actually tested?

Highlted part is related because we are providing the credientials like this: 
```php
$tenant = Tenant::create([
        'tenancy_db_name' => $name,
        'tenancy_db_username' => 'root',
        'tenancy_db_password' => 'password',
    ]);
```

And when we do assert that the database is created, that means credentials in tenant config are being used to create the database successfully. So "db can be created **by using the username and password** from tenant config". I don't know why this is confusing.

src/Database/DatabaseConfig.php
abrardev99: I've removed the comments related to `hostConnection` method. It was creating confusion. `hostConnection` method itself has comments so I think we don't need one here.

tests/TenantDatabaseManagerTest.php
stancl: @abrardev99 Don't just rename the variable, get rid of the inline assignment as well.

abrardev99: Sure.

tests/TenantDatabaseManagerTest.php
stancl: Resolved on BC.

stancl:
Thanks for your work on this ❤️

PR #947, opened 2022-09-14
[4.x] Finish incomplete and missing tests

abrardev99:
This PR completes the missing/skipping tests. Now there is only one skipping test related to the Debuggable feature. As discussed internally, this feature is still incomplete, so I left the test as it is.

### DIFF
--- a/src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
+++ b/src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
@@ -10,6 +10,9 @@ use Throwable;
 
 class SQLiteDatabaseManager implements TenantDatabaseManager
 {
+    /**
+     * SQLite Database path without ending slash.
+     */
     public static string|null $path = null;
 
     public function createDatabase(TenantWithDatabase $tenant): bool
@@ -49,6 +52,12 @@ class SQLiteDatabaseManager implements TenantDatabaseManager
 
     public function getPath(string $name): string
     {
-        return static::$path ? static::$path . '/' . $name : database_path($name);
+        if (static::$path) {
+            // The path is set, so return the full path with the database name
+            return str(static::$path)->append(DIRECTORY_SEPARATOR)->append($name)->toString();
+        }
+
+        // The path is not set so return the default path
+        return database_path($name);
     }
 }
--- a/src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
+++ b/src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
@@ -53,11 +53,9 @@ class SQLiteDatabaseManager implements TenantDatabaseManager
     public function getPath(string $name): string
     {
         if (static::$path) {
-            // The path is set, so return the full path with the database name
-            return str(static::$path)->append(DIRECTORY_SEPARATOR)->append($name)->toString();
+            return static::$path . DIRECTORY_SEPARATOR . $name;
         }
 
-        // The path is not set so return the default path
         return database_path($name);
     }
 }
--- a/src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
+++ b/src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
@@ -10,10 +10,12 @@ use Throwable;
 
 class SQLiteDatabaseManager implements TenantDatabaseManager
 {
+    public static string|null $path = null;
+
     public function createDatabase(TenantWithDatabase $tenant): bool
     {
         try {
-            return file_put_contents(database_path($tenant->database()->getName()), '');
+            return (bool) file_put_contents($this->getPath($tenant->database()->getName()), '');
         } catch (Throwable) {
             return false;
         }
@@ -22,7 +24,7 @@ class SQLiteDatabaseManager implements TenantDatabaseManager
     public function deleteDatabase(TenantWithDatabase $tenant): bool
     {
         try {
-            return unlink(database_path($tenant->database()->getName()));
+            return unlink($this->getPath($tenant->database()->getName()));
         } catch (Throwable) {
             return false;
         }
@@ -30,7 +32,7 @@ class SQLiteDatabaseManager implements TenantDatabaseManager
 
     public function databaseExists(string $name): bool
     {
-        return file_exists(database_path($name));
+        return file_exists($this->getPath($name));
     }
 
     public function makeConnectionConfig(array $baseConfig, string $databaseName): array
@@ -43,5 +45,10 @@ class SQLiteDatabaseManager implements TenantDatabaseManager
     public function setConnection(string $connection): void
     {
         //
+    }
+
+    public function getPath(string $name): string
+    {
+        return static::$path ? static::$path . '/' . $name : database_path($name);
     }
 }
--- a/tests/CommandsTest.php
+++ b/tests/CommandsTest.php
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use Illuminate\Database\DatabaseManager;
 use Illuminate\Support\Facades\Artisan;
 use Illuminate\Support\Facades\DB;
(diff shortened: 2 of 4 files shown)


tests/TenantDatabaseManagerTest.php
```diff
@@ -225,7 +225,21 @@
 });
 
 test('path used by sqlite manager can be customized', function () {
-    pest()->markTestIncomplete();
+    Event::listen(TenantCreated::class, JobPipeline::make([CreateDatabase::class])->send(function (TenantCreated $event) {
+        return $event->tenant;
+    })->toListener());
+
+    // Set custom path for SQLite file
+    SQLiteDatabaseManager::$path = $customPath = storage_path();
```
stancl: I wouldn't use `storage_path()` for everything, it makes the code unclear. It'd be better to either hardcode a path here or use some gitignored directory.

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -10,10 +10,12 @@
 
 class SQLiteDatabaseManager implements TenantDatabaseManager
 {
+    public static string|null $path = null;
```
stancl, line 16: Is this essentially a new feature? Customizing the path used by this DB manager.

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -44,4 +46,9 @@ public function setConnection(string $connection): void
     {
         //
     }
+
+    public function getPath(string $name): string
+    {
+        return static::$path ? static::$path . '/' . $name : database_path($name);
```
stancl: Refactor the ternary into a proper if condition

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
abrardev99: > Is this essentially a new feature?

Yes, it is. So completing the test is first adding the ability to change the path.

tests/TenantDatabaseManagerTest.php
abrardev99: So I used the `database_path('custom')` now. I think it makes sense because it's more likely developers put the sqlite database inside Laravel's `database` directory. Using this path works well with testbench means we don't need a gitignore directory etc.

stancl:
The ternary I told you to convert to a proper if check — don't overcomplicate it. You're using `str()` just to join strings and then you have to call `->toString()` at the end. That just adds complexity.

Also, the comments aren't needed.

All you need to do there is just change the ternary into an `if ()` and the code will be perfect.

tests/SingleDatabaseTenancyTest.php
```diff
@@ -204,16 +204,16 @@
     Post::create(['text' => 'Foo', 'slug' => 'foo']);
     $data = ['text' => 'Foo 2', 'slug' => 'foo'];
 
-    $uniqueFails = Validator::make($data, [
+    $uniqueValidator = Validator::make($data, [
         'slug' => 'unique:posts',
-    ])->fails();
-    $existsFails = Validator::make($data, [
+    ]);
+    $existsValidator= Validator::make($data, [
```
stancl: What's the change you made here? I don't see the point of moving the `->fails()` call to a later point and creating a validator instance.

Also, there's wrong code style on this line. And we generally don't use indented `->and()` in tests, it's better to just make two separate assertions, especially if there are `// comments` that would be vertically aligned (right below each other)

tests/CommandsTest.php
```diff
@@ -116,8 +118,19 @@
...
 });
 
-// Incomplete test
-test('seed command works');
+test('seed command works', function (){
+    $tenant = Tenant::create();
+    Artisan::call('tenants:migrate');
+    expect(Schema::hasTable('users'))->toBeFalse();
+
+    Artisan::call('tenants:seed', ['--class' => TestSeeder::class]);
```
stancl: This part doesn't make sense. You're making assertions about the central table (that the `users` table doesn't exist there), but it's just confusing because **it looks as if the seeder created the table**.

Also, you're not actually asserting that the seeder does something.

You're meant to assert that there are no users in the tenant, then run the seeder, and then assert that there are users in the tenant.

tests/TenantDatabaseManagerTest.php
```diff
@@ -225,7 +225,26 @@
...
+    }
+
+    $name = Str::random(8). '.sqlite';
+    Tenant::create([
+        'tenancy_db_name' => $name,
+        'tenancy_db_connection' => 'sqlite',
+    ]);
+
+    expect(file_exists( $customPath . '/' . $name))->toBeTrue();
+    unlink($customPath . '/' . $name); // clean up
```
stancl: It's better to use a random file than to do clean-up at the end of the test. What if the test failed mid-way? The file wouldn't get cleaned up and there'd be issues with running the test the next time around.

abrardev99: Got it. Refactored it to a random dir name.

tests/SingleDatabaseTenancyTest.php
abrardev99: > What's the change you made here? I don't see the point of moving the ->fails() call to a later point and creating a validator instance.

I see. Ok, I'll revert the changes to the same as it was. 

Your main question is, "What's the change you made here?". 

So these are the assertions. 
```php
// pest()->assertFalse($uniqueFails); // todo get these two assertions to pass. for some reason, the validator is passing for both 'unique' and 'exists'
// pest()->assertTrue($existsFails); // todo get these two assertions to pass. for some reason, the validator is passing for both 'unique' and 'exists'
```

I think these are basically wrong. We created a post with slug `foo`. And the new data is like this 
`$data = ['text' => 'Foo 2', 'slug' => 'foo'];`

So the **unique** rule should fail because the slug is not unique, and the **exists** rule should pass because the slug already exists. The final assertions look like this. 
```php
expect($uniqueFails)->toBeTrue(); // Expect unique rule failed to pass because slug 'foo' already exists
expect($existsPass)->toBeTrue(); // Expect exists rule pass because slug 'foo' exists
```

I've refactored the test.

tests/CommandsTest.php
abrardev99: Thank you for the review. Yes, this test was not making sense, how dumb. I've just updated the test now. Now we assert user's table is empty before running the seeder command, and after running the seeder command, we assert that table got the user.

tests/TenantDatabaseManagerTest.php
stancl: Prefix the random string with something so that all of the databases created in this test start with the same string. We follow that practice everywhere. It's better than having *just* random characters.

tests/TenantDatabaseManagerTest.php
```diff
@@ -225,7 +225,25 @@
 });
 
 test('path used by sqlite manager can be customized', function () {
-    pest()->markTestIncomplete();
+    Event::listen(TenantCreated::class, JobPipeline::make([CreateDatabase::class])->send(function (TenantCreated $event) {
+        return $event->tenant;
+    })->toListener());
+
+    // Set custom path for SQLite file
+    SQLiteDatabaseManager::$path = $customPath = database_path(Str::random(8));
```
stancl:
```suggestion
    SQLiteDatabaseManager::$path = $customPath = database_path('custom_' . Str::random(8));
```

tests/TenantDatabaseManagerTest.php
stancl: I've made the change here: https://github.com/archtechx/tenancy/pull/947#discussion_r979139166

You can mark this as resolved when you see it 👍🏻

abrardev99: Ah, I see. That's a good pattern. I'll keep an eye out next time.

PR #956, opened 2022-09-23
[3.x] Add Vite helper for tenancy

wilsenhc:
This PR adds a custom implementation of the `Vite` helper that always uses the `global_asset()` helper.

I wouldn't know how to go about adding tests for this, or if it is really necessary. 
It didn't seem like something really needed when I opened the [PR](https://github.com/laravel/framework/pull/44037) to Laravel that makes this possible in the first place


src/TenancyServiceProvider.php
```diff
@@ -5,6 +5,7 @@
 namespace Stancl\Tenancy;
 
 use Illuminate\Cache\CacheManager;
+use Illuminate\Foundation\Vite as BaseVite;
```
stancl: Is this directly part of Laravel without any other dependencies?

src/TenancyServiceProvider.php
```diff
@@ -64,10 +65,13 @@ public function register(): void
         $this->app->singleton(Commands\Rollback::class, function ($app) {
             return new Commands\Rollback($app['migrator']);
         });
+
         $this->app->singleton(Commands\Seed::class, function ($app) {
             return new Commands\Seed($app['db']);
         });
 
+        $this->app->singleton(BaseVite::class, Vite::class);
```
stancl: It should be possible to disable this behavior. Perhaps we could make it an [optional feature](https://tenancyforlaravel.com/docs/v3/optional-features) and move the `singleton()` call there.

src/TenancyServiceProvider.php
wilsenhc: Yes, directly from Laravel without extra dependencies. Only version number should be [Laravel v9.29.0](https://github.com/laravel/framework/releases/tag/v9.29.0) and above for this to work

src/TenancyServiceProvider.php
wilsenhc: Sounds good to me, to move it to a optional feature so it can be disabled.

I'll mark this PR as draft and update it later today to convert this as an optional feature

stancl:
Have you tested this approach locally? I'm not 100% sure if the Feature gets executed early enough in the request lifecycle to override the Vite binding. It *should*, but it'd be great to have confirmation.

wilsenhc:
Been testing it with a project locally and it is working fine

tests/Features/ViteBundlerTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+
+test('replaces the vite helper instance with custom class', function () {
+    $concreteBindings = App::getBindings()[FoundationVite::class]['concrete'];
+
+    expect($concreteBindings(App::getInstance()))
+        ->toBeInstanceOf(FoundationVite::class);
+
+    expect($concreteBindings(App::getInstance()))
+        ->not->toBeInstanceOf(StanclVite::class);
```
stancl: We can just instantiate the class and later remove it from the container:
```suggestion
    $vite = app(\Illuminate\Foundation\Vite::class);

    expect($vite)->toBeInstanceOf(FoundationVite::class);
    expect($vite)->not->toBeInstanceOf(StanclVite::class);
```

tests/Features/ViteBundlerTest.php
```diff
@@ -0,0 +1,41 @@
...
+    expect($concreteBindings)
+        ->toBeCallable();
+
+    expect($concreteBindings(App::getInstance()))
+        ->toBeInstanceOf(FoundationVite::class);
+
+    expect($concreteBindings(App::getInstance()))
+        ->toBeInstanceOf(StanclVite::class);
+
+    tenancy()->end();
```
stancl:
```suggestion
    app()->forgetInstance(\Illuminate\Foundation\Vite::class);

    $vite = app(\Illuminate\Foundation\Vite::class);

    expect($vite)->toBeInstanceOf(StanclVite::class);
```

stancl:
Thanks a lot for the PR!

@lukinovec could you create a docs task for this on BC? Just a simple integration page mentioning that you can enable this feature to make Vite generate asset paths correctly (instead of using the tenant-scoped `asset()` helper).

PR #957, opened 2022-09-24
[4.x] Refactor cached tenant resolvers

erikgaal:
Still draft, but I was working on some refactoring that makes it easier to swap implementations in user land. Next to that, using the service container and less `static` properties allows us to work towards supporting Octane.

This PR tries to refactor the CachedTenantResolvers with a decorator pattern. The three implementations of TenantResolver no longer know about caching at all. Instead, in the service provider we create the TenantResolvers.

Also added a TenantRepository interface which makes testing the tenant resolvers in isolation a bit easier.

Let me know what you think, I'm happy to do some refactoring work!

To do:
- [ ] Make invalidating cache work again.
- [ ] Find a solution for `DomainTenantResolver::$currentDomain`, should not be static.
- [ ] Cleanup TenantResolver creation in service provider.


src/Repository/IlluminateTenantRepository.php
```diff
@@ -0,0 +1,48 @@
...
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Repository;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+use Stancl\Tenancy\Contracts\Tenant;
+
+class IlluminateTenantRepository implements TenantRepository
```
stancl, line 11: What is this class meant to represent?

Also:
1. What's the point of the `TenantRepository` interface?
2. The naming here implies that this is a part of Laravel (`Illuminate`)

assets/config.php
```diff
@@ -196,4 +196,28 @@
         '--class' => 'DatabaseSeeder', // root seeder class
         // '--force' => true,
     ],
+
+    /**
+     *
+     */
+    'tenant_resolvers' => [
```
stancl, line 203: What's your rationale for moving the static properties to config?

stancl:
Could you add a bit more description to the PR explaining the changes? It seems that there's a lot of OOP complexity introduced, so there should be a good reason for that.

stancl:
> makes it easier to swap implementations in user land

Could you provide some real use cases? If you want to customize how tenants are identified, the absolute easiest way to do that is to just write a simple middleware class. The only reason why we have the resolvers is to separate that logic from the middleware and make it more reusable, but if you want to do anything custom, the easiest approach is to just use your own middleware. I think that comes with the lowest complexity.

> less static properties allows us to work towards supporting Octane

I don't think this makes a difference here? The static properties are for "class config", I don't think there are use cases for changing them during runtime. Therefore the data is not request-specific and doesn't cause any issues when using Octane.

> Also added a TenantRepository interface which makes testing the tenant resolvers in isolation a bit easier

Does the reduced complexity in tests significantly offset the introduced complexity in the package code?

src/Repository/IlluminateTenantRepository.php
robertvansteen: Having a tenant (and domain) repository is very useful for testing. Because it allows you to test on one connection (`tenant`) because you can swap out the repository for the models in the central one for a fake implementation.

stancl: Can you give a specific example?

stancl:
I looked at the code again and I don't think this PR is worth pursuing. A lot of the added abstractions actually introduce limitations. For instance, the tenant repository is coupled with the concept of domains. The `IlluminateTenantRepository` is coupled with the specific implementation of domains. And I generally don't see the point of wrapping models in `Repository` abstractions.

The tests are very low level which is the opposite of what we're going for. We write tests that work similarly as our users' applications. There's a lot of container logic and mocks in the tests which doesn't actually test the logic in a holistic way, it just makes assertions that your abstractions work a certain way. E.g. what you wrote here:

> added a TenantRepository interface which makes testing the tenant resolvers in isolation a bit easier

isn't what we're aiming for. The package changes a lot of things about how Laravel works, so we want to test things in a complete way, not in isolation. And we definitely don't want to add more abstractions to be able to test something in isolation.

The bindings added in `TenancyServiceProvider` are long and repetitive. It just seems like an OOP overkill implementation with little actual gain.

Regarding the PR description and what @rovansteen wrote, I still don't see how this improves testability or the ability to do custom things easily in any significant way. If anything, I feel that this imposes a large OOP burden on anyone who wants to customize how the package works in their app.

I appreciate the effort but don't think this is worth introducing in the package.

robertvansteen:
@stancl I understand that the package is very intertwined with how Laravel works but there is an argument to be made for the repositories, not per se for unit tests but to test one part of the application on it's own. In most applications the majority of the code you write will be related to a tenant but to test that code you'll need two databases: the central one and the tenant one. Having a repository for the "central" models allow you to swap them out for fake ones so you'll only need the tenant database and not the central one. This will also unlock the usage of parallel testing which is otherwise not possible since it will assume one database per test.

src/Repository/IlluminateTenantRepository.php
robertvansteen: If you can use fake models for the central connection you'll only need one database which, among other things, allows you to do parallel testing.

stancl:
> Having a repository for the "central" models allow you to swap them out for fake ones so you'll only need the tenant database and not the central one. This will also unlock the usage of parallel testing 

I'd need to see a full example to consider this.

PR #958, opened 2022-09-25
use force in MigrateDatabase and SeedDatabase Job

itsrafsanjani:
When running with `APP_ENV=production` jobs could not execute those `Artisan::call` commands. Was throwing `Undefined constant "STDIN"` error.


stancl:
That's what the config is for: https://github.com/archtechx/tenancy/blob/53a9d24b7c3d2c7e91cc701bd63871fba768b840/assets/config.php#L187

Perhaps we should document this a bit better.

Though it's `true` by default, so I'm not sure why this isn't working for you. It could be the seeder that's failing, but we intentionally keep that `false` since running the seeder by accident in production is more dangerous than accidentally running migrations.

Issue #854, opened 2022-04-29
PHP 8.1: Deprecation notice Accessing static trait property ($tenantIdColumn)


oscar-ol:
### Bug description

I'm getting this deprecation notice (PHP 8.1):

PHP Deprecated:  Accessing static trait property Stancl\Tenancy\Database\Concerns\BelongsToTenant::$tenantIdColumn is deprecated, it should only be accessed on a class using the trait in /var/www/html/vendor/stancl/tenancy/src/Database/TenantScope.php on line 20
PHP Deprecated:  Accessing static trait property Stancl\Tenancy\Database\Concerns\BelongsToTenant::$tenantIdColumn is deprecated, it should only be accessed on a class using the trait in /var/www/html/vendor/stancl/tenancy/src/Database/Concerns/BelongsToTenant.php on line 19

### Steps to reproduce

php artisan tinker

tenancy()->initialize(Tenant::first())

User::first()->tenant

### Expected behavior

no messages

### Laravel version

8.83.5

### stancl/tenancy version

3.5.6

stancl:
Thanks, we'll change this to use something like a config key.

ahmedsayedabdelsalam:
fixed ?

hai-shift:
fixed ?

oscar-ol:
@stancl please reopen this issue

garethnic:
Oki so for everyone else wondering, I sorted it out temporarily like this:

In `config/tenancy.php` I added `'tenant_column' => 'tenant_id',`.

I copied the following files from `vendor` to locations under `app/`:
- \Stancl\Tenancy\Database\Concerns\BelongsToTenant::class
- \Stancl\Tenancy\Database\TenantScope::class

Inside of the boot method in `app/Providers/AppServiceProvider.php`:
```
$this->app->bind(BelongsToTenant::class, \Stancl\Tenancy\Database\Concerns\BelongsToTenant::class);
$this->app->bind(TenantScope::class, \Stancl\Tenancy\Database\TenantScope::class);
```

In those files that I now have locally, I override any instance of `BelongsToTenant::$tenantIdColumn` with `config('tenancy.tenant_column')`

Then in my models I just use the local trait `use App\Helpers\Traits\BelongsToTenant;`.

I also ran `php artisan config:clear`  and `composer dumpautoload -o`.

I stopped seeing `PHP Deprecated` messages in tinker. Hope it helps.

stancl:
[...]

PR #964, opened 2022-09-30
854 - Accessing static trait property $tenantIdColumn resolve

zach2825:
Issue - #854

### DIFF
--- a/src/Database/Concerns/BelongsToTenant.php
+++ b/src/Database/Concerns/BelongsToTenant.php
@@ -16,7 +16,7 @@ trait BelongsToTenant
 
     public function tenant()
     {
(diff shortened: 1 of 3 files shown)


stancl:
I don't think adding the trait to all of those classes is a good solution. I think we'll have to get rid of the property in general and move it to some config key in v4

zach2825:
I completely agree but for v3 until v4 is released etc etc

stancl:
It's not easy to fix this in v3 because it will require restructuring the code in a significant way. This PR for instance would break the code of anyone using single DB tenancy with a custom tenant id column.

And I don't think there's any priority issue here, it's just a deprecation notice. It shouldn't even show up in the browser, at least in production

PR #967, opened 2022-10-07
[4.x] Add tenancy maintenance drivers and events

stein-j:
As a follow up of the conversation from the doc repo https://github.com/stancl/tenancy-docs/issues/214 and because #761 was merge.

I figured this package could also get the privilege of maintenance drivers like Laravel has done. This PR imitates exactly how Laravel handles its maintenance mode. Here is what have been added :

- Two tenancy drivers `database` & `cache` (whereas Laravel has `file` & `cache`). Note: The `file` driver cannot be done without touching the _public/index.php_
- A new config `tenancy.maintenance.driver` which will determine which driver to use (database by default) and `tenance.mainenance.store` which will determine the store to use when the driver is _cache_.
- A new .env variable `TENANCY_MAINTENANCE_DRIVER` (Laravel didn't make an environment variable, so maybe we shouldn't add one)
- Events are now thrown when a tenant is put into or out of maintenance mode.

> ⚠️Note : The cache driver doesn't seem to be using the prefixed added by the package in the bootstrapper for the cache. So I've prefixed the cache key with the tenant key... But it's not ideal because it really should be complying to how the package handles the cache prefixes for identified tenants. I'm open to any suggestions. Everything works, though.

### DIFF
--- a/assets/TenancyServiceProvider.stub.php
+++ b/assets/TenancyServiceProvider.stub.php
@@ -58,6 +58,8 @@ class TenancyServiceProvider extends ServiceProvider
                     return $event->tenant;
                 })->shouldBeQueued(false), // `false` by default, but you probably want to make this `true` for production.
             ],
+            Events\TenantMaintenanceModeEnabled::class => [],
+            Events\TenantMaintenanceModeDisabled::class => [],
 
             // Domain events
             Events\CreatingDomain::class => [],
--- a/assets/config.php
+++ b/assets/config.php
@@ -253,7 +253,7 @@ return [
      *
      */
     'maintenance' => [
-        'driver' => env('TENANCY_DATABASE_DRIVER', 'database'),
+        'driver' => env('TENANCY_MAINTENANCE_DRIVER', 'database'),
         // 'store'  => 'redis',
     ],
 
--- a/assets/config.php
+++ b/assets/config.php
@@ -243,6 +243,21 @@ return [
     ],
 
     /**
+     * Maintenance mode driver
+     *
+     * These configuration options determine the driver used to determine and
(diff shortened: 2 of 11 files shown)


stancl:
Appreciate the PR, though I'm not sure if this is worth pursuing. Like we discussed in the other thread, our maintenance mode implementation has some limitations — *but that's not necessarily a bad thing*. This implementation seems to add a decent amount of complexity and I'm not sure if the benefit to the end user outweighs that.

Does this PR add anything that's missing besides supporting multiple drivers?

Supporting only the DB driver can have its benefits too — storing this kind of data on the `Tenant` record across the package features nicely centralizes things there.

So I appreciate your work on this but I don't think this makes sense to add to the package.

stein-j:
It's up to you, I think it's a good idea because this feature is an imitation of Laravel's implementation of the maintenance mode, so using the same code is useful. If someone tries to understand how the maintenance work, they'll only need to understand it once.

Also, one cool feature of using the Manager class is that users may add their own drivers :

```php
public function boot(MaintenanceModeManager $maintenanceModeManager)
    {
        $maintenanceModeManager->extend('file', function ($container) {
            return $container->make(FileBasedMaintenanceMode::class);
        });
}
```

Also, this PR did implement events for when a tenant goes into and out of maintenance mode. I can do another PR just for that if you wan to.

stein-j:
@stancl just in case you missed it :

> Also, this PR did implement events for when a tenant goes into and out of maintenance mode. I can do another PR just for that if you wan to.

Should I make another PR that adds the maintenance events (like Laravel has) ?

stancl:
You can, yeah. Just fire them from `MaintenanceMode` 👍🏻

PR #968, opened 2022-10-09
[4.x] Update commands CLI outputs

stein-j:
Now that 4.x will have Laravel 9 as a minimum requirement, the cli output can be updated to the new Laravel standard:

Here's every change:
- All commands now uses the new Laravel CLI output (with `$this->components->info()`)
- The `tenanc:install` command will now offer to open the Github repo to add a star (Inspired by [Pest](https://github.com/pestphp/pest/blob/3d16183a9377c6c24c6385fb9a76e8c1fbb87860/src/Console/Thanks.php#L42))
- All commands now always return a success or failure confirmation

Preview of the CLI outpts:

### Listing tenants:
The id of the tenant are displayed on the left whereas the domains are on the right.
Screen Shot 2022-10-08 at 23 57 09

### Commands running on multiple tenants :
Screen Shot 2022-10-09 at 00 14 00

### Install tenancy:
Screen Shot 2022-10-08 at 23 57 38

### DIFF
--- a/src/Commands/Down.php
+++ b/src/Commands/Down.php
@@ -22,24 +22,28 @@ class Down extends DownCommand
 
     public function handle(): int
     {
-        // The base down command is heavily used. Instead of saving the data inside a file,
-        // the data is stored the tenant database, which means some Laravel features
-        // are not available with tenants.
-
+        // First we retrieve the compiled payload with all the
+        // parameters given, then for each tenant we will
+        // put down for maintenance and pass the payload.
         $payload = $this->getDownDatabasePayload();
 
-        // This runs for all tenants if no --tenants are specified
         tenancy()->runForMultiple($this->getTenants(), function ($tenant) use ($payload) {
-            $this->line("Tenant: {$tenant['id']}");
+            $this->components->info("Tenant: {$tenant->getTenantKey()}");
             $tenant->putDownForMaintenance($payload);
         });
 
-        $this->comment('Tenants are now in maintenance mode.');
+        $this->components->info('Tenants are now in maintenance mode.');
 
         return 0;
     }
 
-    /** Get the payload to be placed in the "down" file. */
+    /**
+     * Get the payload to be placed in the "down" file. This
+     * payload is the same as the original function
+     * but without the 'template' option
+     *
+     * @return array
+     */
     protected function getDownDatabasePayload(): array
     {
         return [
--- a/src/Commands/Down.php
+++ b/src/Commands/Down.php
@@ -37,9 +37,7 @@ class Down extends DownCommand
     /**
      * Get the payload to be placed in the "down" file. This
      * payload is the same as the original function
-     * but without the 'template' option
-     *
-     * @return array
+     * but without the 'template' option.
      */
     protected function getDownDatabasePayload(): array
     {
--- a/src/Commands/Down.php
+++ b/src/Commands/Down.php
@@ -22,9 +22,6 @@ class Down extends DownCommand
 
     public function handle(): int
     {
-        // First we retrieve the compiled payload with all the
-        // parameters given, then for each tenant we will
-        // put down for maintenance and pass the payload.
         $payload = $this->getDownDatabasePayload();
 
         tenancy()->runForMultiple($this->getTenants(), function ($tenant) use ($payload) {
--- a/src/Commands/Install.php
+++ b/src/Commands/Install.php
@@ -10,13 +10,13 @@ class Install extends Command
 {
     protected $signature = 'tenancy:install';
 
-    protected $description = 'Install stancl/tenancy.';
+    protected $description = 'Install Tenancy for Laravel.';
 
     public function handle(): int
     {
         $this->newLine();
 
-        $this->components->task('Publishing config file', function () {
+        $this->components->task('Publishing config file [config/tenancy.php]', function () {
             $this->callSilent('vendor:publish', [
                 '--provider' => 'Stancl\Tenancy\TenancyServiceProvider',
                 '--tag' => 'config',
@@ -25,8 +25,8 @@ class Install extends Command
 
         $this->newLine();
 
-        if (!file_exists(base_path('routes/tenant.php'))) {
-            $this->components->task('Publishing routes', function () {
+        if (! file_exists(base_path('routes/tenant.php'))) {
+            $this->components->task('Publishing routes [routes/tenant.php]', function () {
                 $this->callSilent('vendor:publish', [
                     '--provider' => 'Stancl\Tenancy\TenancyServiceProvider',
                     '--tag' => 'routes',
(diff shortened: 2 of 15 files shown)


stancl:
> Commands running on multiple tenants

Is there an empty line between the `Tenant: ...` lines? Not sure if that's your terminal or if there are extra spaces here. Same for `install`.

src/Commands/Install.php
```diff
@@ -12,42 +12,97 @@ class Install extends Command
...
+
+    /**
+     * Pause the console execution for 1 second, help the user to have time and read the output
+     *
+     * @return void
+     */
+    public function patienceIsKeyToLife(): void
+    {
+        sleep(1);
+    }
```
stancl: Get rid of this method, we can just use `sleep(1)` inline. Also, the output isn't long, so I don't think there's a reason to add multiple pauses to the installation command?

Currently the command feels instant and the output is short and readable. So making it feel like there's a lot happening in the background by adding `sleep()`s is probably not worth it.

src/Commands/Install.php
```diff
@@ -12,42 +12,97 @@ class Install extends Command
...
+
+    /**
+     * If the user accepts, opens the GitHub project in the browser
+     *
+     * @return void
+     */
+    public function askForSupport(): void
+    {
+        if ($this->components->confirm("Would you like to show your support by starring the project on github ?", true)) {
```
stancl:
```suggestion
        if ($this->components->confirm("Would you like to show your support by starring the project on GitHub?", true)) {
```

src/Commands/TenantDump.php
```diff
@@ -39,9 +31,15 @@ public function tenant(): Tenant
             $tenant = tenancy()->find($tenant);
         }
 
-        throw_if(! $tenant, 'Could not identify the tenant to use for dumping the schema.');
+        if (is_null($tenant)){
```
stancl:
```suggestion
        if ($tenant === null) {
```

src/Commands/TenantList.php
```diff
@@ -14,19 +14,44 @@ class TenantList extends Command
...
+
+    /**
+     * Generate the visual cli output for the domain names
+     *
+     * @param  Model  $tenant
+     * @return string|null
+     */
+    protected function domainsCli(Model $tenant): ?string
+    {
+        if (is_null($tenant->domains)){
```
stancl:
```suggestion
        if (! $tenant->domains) {
```

tests/CommandsTest.php
```diff
@@ -167,7 +167,9 @@
         mkdir($dir, 0777, true);
     }
 
-    pest()->artisan('tenancy:install');
+    pest()->artisan('tenancy:install')
+        ->expectsConfirmation('Would you like to show your support by starring the project on github ?', 'no')
```
stancl:
```suggestion
        ->expectsConfirmation('Would you like to show your support by starring the project on GitHub?', 'no')
```

tests/CommandsTest.php
```diff
@@ -220,7 +223,8 @@
 
     pest()->artisan('tenants:link', [
         '--remove' => true,
-    ]);
+    ])
+        ->assertExitCode(0);
```
stancl:
```suggestion
    ])->assertExitCode(0);
```

src/Commands/Down.php
```diff
@@ -22,24 +22,28 @@ class Down extends DownCommand
 
     public function handle(): int
     {
-        // The base down command is heavily used. Instead of saving the data inside a file,
-        // the data is stored the tenant database, which means some Laravel features
-        // are not available with tenants.
-
+        // First we retrieve the compiled payload with all the
+        // parameters given, then for each tenant we will
+        // put down for maintenance and pass the payload.
```
stancl: You can probably get rid of the component altogether since the code is self-explanatory

src/Commands/Down.php
```diff
@@ -22,24 +22,28 @@ class Down extends DownCommand
...
     }
 
-    /** Get the payload to be placed in the "down" file. */
+    /**
+     * Get the payload to be placed in the "down" file. This
+     * payload is the same as the original function
+     * but without the 'template' option
+     *
+     * @return array
+     */
```
stancl: What's the `template` option for?

stein-j: When using the `php artisan down` command with the `--render` option. It generates the html and stores is the in payload as "template". See [Laravel's version](https://github.com/laravel/framework/blob/5f89ab07d2916f8b3474afa5bdccbdd49bced8ba/src/Illuminate/Foundation/Console/DownCommand.php#L85) of this function

stein-j:
> Is there an empty line between the Tenant: ... lines? Not sure if that's your terminal or if there are extra spaces here. Same for install.

@stancl good eye. Laravel always adds a line before each [INFO] / [WARN] / etc. It was my terminal because it's not doing it when trying now.

stancl:
> Laravel always adds a line before each [INFO] / [WARN] / etc. It was my terminal because it's not doing it when trying now.

What do you mean by the second sentence?

In my reviews, I meant that we *probably* don't want to add empty lines between these outputs. Is it Laravel that adds them?

stein-j:
Yes, Laravel adds the space.

I did have to add a line space in the `tenancy:install` command otherwise the visual would be weird.

stancl:
What do you mean by weird? Can you show how it looks without those?

Also, the first picture in your PR description doesn't have the added lines. Is it because that's a table?

stein-j:
Look at the screenshots from the orignal [PR](https://github.com/laravel/framework/pull/43065) that introduced the new CLI output.

All output `$this->components->info()`, `$this->components->error()` `$this->components->warn()` will add a new line after. But all others outputs `$this->components->task()`, `$this->components->confirm()`, etc. do not. 

This is the output without the `$this->newLine();`
Screen Shot 2022-10-10 at 19 34 28

stancl:
@abrardev99 FYI I don't know if for PRs, the ci config is taken from the target branch or the incoming branch. So it'd probably be best to test this in your own tenancy repo, I think you have a second GH account that you can use for the PR.

tests/CommandsTest.php
stancl: @stein-j You marked this as resolved but didn't fix it?

stein-j: Yes I did in this commit a8f416dc1a3ccb6c0050d6545acf088da10aaeec
Did I use the Github feature wrong ?

stancl:
Working on finishing this. Some of the installation steps had missing checks & warnings, which made the output look weird when only some issues were reported. Also, some steps were missing clickable links which also looked odd due to the inconsistency.

Changed it to this (installation after Tenancy has already been installed):
CleanShot 2022-10-18 at 6 26 24@2x

And this when it's installed correctly:
CleanShot 2022-10-18 at 6 26 08@2x

tests/CommandsTest.php
stancl: https://github.com/archtechx/tenancy/pull/968/files#diff-2d46e8828aea0422777576ca8e903262bf80a262978ec02ad3d0783c38fe1b32R171

Not sure what you mean by "Github feature". If you want to directly commit my changes you can just use this:

CleanShot 2022-10-18 at 6 29 17@2x

stancl: Anyway, I fixed that now.

stancl:
Thanks for the PR!

PR #970, opened 2022-10-10
Add `current()` and `currentOrFail()` tenant methods

lukinovec:


src/Database/Models/Tenant.php
```diff
@@ -45,6 +46,16 @@ public function getTenantKey(): int|string
         return $this->getAttribute($this->getTenantKeyName());
     }
 
+    public static function current(): self|null
```
stancl: Unless there's a specific reason for using `self`, always use `static`.

`self` is static with limitations that make sense in certain cases.

src/Database/Models/Tenant.php
```diff
@@ -45,6 +46,16 @@ public function getTenantKey(): int|string
...
+    {
+        return tenant();
+    }
+
+    public static function currentOrFail(): self
```
stancl:
```suggestion
    /** @throws TenancyNotInitializedException */
    public static function currentOrFail(): static
```

src/Database/Models/Tenant.php
lukinovec: Yeah, I used the model name again and cs fixer changed it to `self` after I pushed. I'll change that to `static`

lukinovec:
@stancl I think you can merge this now

stancl:
Thanks!

PR #971, opened 2022-10-10
[4.x] Drop tenant databases on `migrate:fresh`

lukinovec:
@stancl, I'm drafting this as a follow-up on [the comments on BC](https://3.basecamp.com/5170965/buckets/23651018/todos/5400823470). From the conversation:

Samuel:

> The general goal is that sometimes, it's appropriate to delete tenant DBs when running migrate:fresh. Reason: when you run migrate:fresh you generally expect your DB to become clean, so it'd make sense to also delete databases.
> 
> Some thoughts:
> We probably only want to delete DBs of the tenants that are currently in the `tenants` table and will be removed when migrate:fresh runs. We don't want to use any logic like detecting tenant DBs based on the configured prefix (e.g. tenant_) because someone could have those DBs for other purposes such as testing (e.g. tenant_tests)
> We can't modify the default Laravel command, but we likely also don't want to add a new command. One way to handle this could be checking if migrate:fresh dispatches any events. I think `migrate` does, so `migrate:fresh` could have some specific ones too. Then, we could simply do Tenants::all()->each->delete() or something like that in a listener

me:

> The `migrate:fresh` command dispatches a bunch of events. `DatabaseRefreshed` is the most specific one (before that gets dispatched, there's a bunch of `StatementPrepared` and `QueryExecuted` events, and these probably aren't useful to us). Also, the `migrate:refresh` command dispatches DatabaseRefreshed too, so the event is not specific to `migrate:fresh`.

> Maybe this is too hacky, but it's the most accurate thing I found so far:

> There's a `CommandStarting` event. The event has a `$command` property which we can access, and if we call `migrate:fresh`, the property's value would be "migrate:fresh". So we could listen for that event and if the property's value equals "migrate:fresh", delete the tenant databases.

### DIFF
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -91,9 +91,13 @@ class TenancyServiceProvider extends ServiceProvider
             Commands\Up::class,
         ]);
 
-        if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
-            $this->app->extend(FreshCommand::class, fn () => new Commands\MigrateFreshOverride);
-        }
+        $this->app->extend(FreshCommand::class, function (FreshCommand $originalCommand) {
+            if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
+                return new Commands\MigrateFreshOverride;
+            }
+
+            return $originalCommand;
+        });
 
         $this->publishes([
             __DIR__ . '/../assets/config.php' => config_path('tenancy.php'),
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -73,7 +73,7 @@ class TenancyServiceProvider extends ServiceProvider
         });
 
         if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
-            $this->app->extend('command.migrate.fresh', fn() => new Commands\MigrateFreshOverride);
+            $this->app->extend('command.migrate.fresh', fn () => new Commands\MigrateFreshOverride);
         }
     }
 
@@ -149,7 +149,7 @@ class TenancyServiceProvider extends ServiceProvider
     public function provides()
     {
         return [
-            'command.migrate.fresh'
+            'command.migrate.fresh',
         ];
     }
 }
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -5,17 +5,17 @@ declare(strict_types=1);
 namespace Stancl\Tenancy;
 
 use Illuminate\Cache\CacheManager;
+use Illuminate\Console\Events\CommandStarting;
 use Illuminate\Support\Facades\Event;
+use Illuminate\Support\Facades\Schema;
 use Illuminate\Support\ServiceProvider;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
 use Stancl\Tenancy\Contracts\Domain;
 use Stancl\Tenancy\Contracts\Tenant;
+use Stancl\Tenancy\Database\Contracts\TenantWithDatabase;
 use Stancl\Tenancy\Enums\LogMode;
 use Stancl\Tenancy\Events\Contracts\TenancyEvent;
 use Stancl\Tenancy\Resolvers\DomainTenantResolver;
-use Illuminate\Console\Events\CommandStarting;
-use Illuminate\Support\Facades\Schema;
-use Stancl\Tenancy\Database\Contracts\TenantWithDatabase;
 
 class TenancyServiceProvider extends ServiceProvider
 {
@@ -133,7 +133,7 @@ class TenancyServiceProvider extends ServiceProvider
             $tenantModel = tenancy()->model();
 
             if ($event->command === 'migrate:fresh' && Schema::hasTable($tenantModel->getTable())) {
-                $tenantModel::all()->each(function(Tenant $tenant) {
+                $tenantModel::all()->each(function (Tenant $tenant) {
                     if (method_exists($tenant, 'domains')) {
                         $tenant->domains()->delete();
                     }
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -91,12 +91,8 @@ class TenancyServiceProvider extends ServiceProvider
             Commands\Up::class,
         ]);
 
-        $this->app->extend(FreshCommand::class, function (FreshCommand $originalCommand) {
-            if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
-                return new Commands\MigrateFreshOverride;
-            }
-
-            return $originalCommand;
+        $this->app->extend(FreshCommand::class, function () {
+            return new Commands\MigrateFreshOverride;
         });
 
         $this->publishes([
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Stancl\Tenancy;
 
 use Illuminate\Cache\CacheManager;
+use Illuminate\Database\Console\Migrations\FreshCommand;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\ServiceProvider;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
@@ -71,10 +72,6 @@ class TenancyServiceProvider extends ServiceProvider
         $this->app->bind('globalCache', function ($app) {
             return new CacheManager($app);
         });
-
-        if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
-            $this->app->extend('command.migrate.fresh', fn () => new Commands\MigrateFreshOverride);
-        }
     }
 
     /* Bootstrap services. */
@@ -95,7 +92,7 @@ class TenancyServiceProvider extends ServiceProvider
         ]);
 
         if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
-            $this->commands(Commands\MigrateFreshOverride::class);
+            $this->app->extend(FreshCommand::class, fn () => new Commands\MigrateFreshOverride);
         }
 
         $this->publishes([
@@ -144,12 +141,5 @@ class TenancyServiceProvider extends ServiceProvider
 
             return $instance;
         });
-    }
-
-    public function provides()
-    {
-        return [
-            'command.migrate.fresh',
-        ];
     }
 }
--- a/src/TenancyServiceProvider.php
+++ b/src/TenancyServiceProvider.php
@@ -5,14 +5,11 @@ declare(strict_types=1);
 namespace Stancl\Tenancy;
 
 use Illuminate\Cache\CacheManager;
-use Illuminate\Console\Events\CommandStarting;
 use Illuminate\Support\Facades\Event;
-use Illuminate\Support\Facades\Schema;
 use Illuminate\Support\ServiceProvider;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
 use Stancl\Tenancy\Contracts\Domain;
 use Stancl\Tenancy\Contracts\Tenant;
-use Stancl\Tenancy\Database\Contracts\TenantWithDatabase;
 use Stancl\Tenancy\Enums\LogMode;
 use Stancl\Tenancy\Events\Contracts\TenancyEvent;
 use Stancl\Tenancy\Resolvers\DomainTenantResolver;
@@ -74,6 +71,10 @@ class TenancyServiceProvider extends ServiceProvider
         $this->app->bind('globalCache', function ($app) {
             return new CacheManager($app);
         });
+
+        if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
+            $this->app->extend('command.migrate.fresh', fn() => new Commands\MigrateFreshOverride);
+        }
     }
 
     /* Bootstrap services. */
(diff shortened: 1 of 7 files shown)


stein-j:
You can always create a new command with the same signature and Laravel will call your command instead of Laravel's.

stein-j:
I don't have access to the original content and am not sure of the current issue.

Shouldn't the package have its own command instead of adding on the original command ? Also I think tenants should be created once the command is done, as the command :fresh is suppose to do.

lukinovec:
I've updated the PR description to make the problem clearer

stancl:
@lukinovec Is the issue with the events that there's only `DatabaseRefreshed`, which runs *after* the DB is refreshed (so we can't get a list of tenants that were used)? Rather than some event that'd run before this, e.g. `RefreshingDatabase`

lukinovec:
> Is the issue with the events that there's only `DatabaseRefreshed`, which runs _after_ the DB is refreshed (so we can't get a list of tenants that were used)? Rather than some event that'd run before this, e.g. `RefreshingDatabase`

I meant the issue is that another command (`migrate:refresh`) also dispatches the event. So deleting tenants on `DatabaseRefreshed` would also make `migrate:refresh` delete the tenants. But I missed the obvious issue that you're pointing out. Even if the event were specific to `migrate:fresh`, it wouldn't be useful because the `tenants` table would be already empty.

> Shouldn't the package have its own command instead of adding on the original command ?

Yeah, I think we'll probably end up doing that. We were figuring out if `migrate:fresh` dispatches some specific event (e.g. `WipingDatabase` or something like that) because if there was an event like that, the tenant dbs could get deleted when the event gets dispatched.

stein-j:
@lukinovec, like I just mentioned. You can just create a new command with the same signature, Laravel will call your command instead of the orignal one, do what you need and then call the original parrent::handle();

```php
<?php

namespace Illuminate\Database\Console\Migrations;

class TenantMigrateFresh extends FreshCommand
{
     /**
     * The console command name.
     *
     * @var string
     */
    protected $name = 'migrate:fresh';

    //public function isEnabled()
    //{
    //    return config('tenancy.commands.fresh.drop_tenants');
    //}

    /**
     *
     * @return int
     */
    public function handle()
    {
       // Drops tenants

        return parent::handle();
    }
}
```

You could probably even use the _isEnabled_ function to disable the command (never tried it) with a logic so that the original command is used instead of this one.

stancl:
> You could probably even use the isEnabled function to disable the command (never tried it) with a logic so that the original command is used instead of this one.

We can just read the config and based on that register our own command in the SP. I'll look into the code a bit and decide the path we want to follow here.

stancl:
Yeah, a custom command that's conditionally registered (based on a new config key) seems best. @lukinovec you can change the PR to that approach so that I see the code. No need to prioritize this over the testing task btw, you can do them in any order.

lukinovec:
> create a new command with the same signature, Laravel will call your command instead of the orignal one

I tried that, and adding the command to `$this->commands()` in TenancyServiceProvider's `boot()` doesn't override Laravel's `migrate:fresh`. I haven't figured this out yet. Do you know of some other ways to override it?

> You could probably even use the isEnabled function to disable the command (never tried it)

I tried it – if it returns `false` and you try to call the command, you'll get an exception saying that the command doesn't exist. But I don't think this would work the way we want – this would affect the whole `migrate:fresh` command (I think the command would get enabled/disabled depending on the config value, but we want the config value to only enable/disable dropping tenant DBs)

stein-j:
> I tried that, and adding the command to $this->commands() in TenancyServiceProvider's boot() doesn't override Laravel's migrate:fresh. I haven't figured this out yet. Do you know of some other ways to override it?

I'm not quite sure, I will look into it. Because commands are declared differently in a Laravel Package it's probably not acting the same way.

> I tried it – if it returns false and you try to call the command, you'll get an exception saying that the command doesn't exist. But I don't think this would work the way we want – this would affect the whole migrate:fresh command (I think the command would get enabled/disabled depending on the config value, but we want the config value to only enable/disable dropping tenant DBs)

You probably have to do like Samuel mentioned. And register the command in the provider conditionally.

stancl:
Conditional registration sounds better than the `isEnabled()` method, since I guess then we'd be disabling the command completely after overriding it.

@lukinovec try googling "laravel override migrate command" (or any other common command, to get good search results). The first few results seem like they'd work.

`$this->commands()` sounds like it could be problematic anyway compared to the `$this->app->extend` approach (suggested in the results I found) since it would depend on the service provider priority.

lukinovec:
> Conditional registration sounds better than the isEnabled() method, since I guess then we'd be disabling the command completely after overriding it.

Yeah

> the $this->app->extend approach (suggested in the results I found)

I found [someone who wants to override Laravel's commands using a package too](https://laracasts.com/discuss/channels/laravel/override-artisan-commands?page=1&replyId=557479). Later in the discussion, [someone claims their package overrides the `make:model` command](https://laracasts.com/discuss/channels/laravel/override-artisan-commands?page=1&replyId=623069). The user linked a repo with the service provider, and he used both `$this->commands()` and `$this->app->extend()`, so I tried it too, but without success, it still doesn't override the command.

stancl:
Show me the command class you're using. You can just commit the changes so that I can look at it — the current code will definitely not be used so even if the new code is incomplete, it's fine.

lukinovec:
> Show me the command class you're using. You can just commit the changes so that I can look at it — the current code will definitely not be used so even if the new code is incomplete, it's fine.

I pushed the changes @stancl. [Here's the command class](https://github.com/archtechx/tenancy/pull/971/files#diff-7759da7da38866bb765d2e5db44ce427ce1a62d5f111595de02030ee08b83b7a)

src/TenancyServiceProvider.php
```diff
@@ -156,4 +145,11 @@ public function boot(): void
             return $instance;
         });
     }
+
+    public function provides()
+    {
+        return [
+            'command.migrate.fresh'
+        ];
+    }
```
stancl: What's the point of this?

src/TenancyServiceProvider.php
```diff
@@ -74,6 +71,10 @@ public function register(): void
         $this->app->bind('globalCache', function ($app) {
             return new CacheManager($app);
         });
+
+        if ($this->app['config']['tenancy.database.drop_tenant_databases_on_migrate_fresh']) {
+            $this->app->extend('command.migrate.fresh', fn() => new Commands\MigrateFreshOverride);
```
stancl: I see nothing with `command.migrate.fresh` binding in the Laravel source code.

Try using `\Illuminate\Database\FreshCommand::class` instead of that for the binding.

Also, FWIW for testing get rid of the `if ()`, or at least add a `dd()` inside it to verify that it works. Sometimes config can't be read this early in the request lifecycle, sometimes you need to do it at a later point.

lukinovec: I see. Thanks, I'll try the binding with `FreshCommand`

src/TenancyServiceProvider.php
lukinovec: I saw that the guy with the `make:model` override package had it in the package's provider. I tried adding it to see if that would help with overriding the command, but it didn't help. I couldn't find documentation of the method, so honestly, I don't know what the method does

src/TenancyServiceProvider.php
lukinovec: Great, you were right, using `FreshCommand::class` instead of `command.migrate.fresh` works to override the command (`$this->app->extend(FreshCommand::class, fn () => new Commands\MigrateFreshOverride);` in `boot()`)

lukinovec:
> Also I think tenants should be created once the command is done, as the command :fresh is suppose to do.

I think when `migrate:fresh` clears `tenants` (or any other table), the records shouldn't be brought back. @stein-j, could you please expand on what you mean here?

src/TenancyServiceProvider.php
lukinovec: Just found the documentation and learned about deferred providers (the `provides` method is only needed in deferred providers) https://laravel.com/docs/9.x/providers#deferred-providers

src/TenancyServiceProvider.php
lukinovec: @stancl, testing that the config value enables/disables the `migrate:fresh` override is tricky because TenancyServiceProvider has to boot again for the config change to have an effect. Booting the service provider mid-test doesn't override the command. Do you have any ideas on how to test this?

stein-j:
@lukinovec, the fresh commands drops all table and then re-run each migration.

This command is often used as a way to quickly reset your database environment by doing `php artisan migrate:fresh --seed`. So it might be expected by developers that the commande drops each tenant and then creates them again.

**But** this is complicated because each app might have custom logic for creating a tenants, and the command cannot know how to tenant should be create. So I would simply forget about creating tenants, developers can easily do `php artisan migrate:fresh --seed && php artisan tenants:create --domain=xx --seed`.

stein-j:
Also, it is important to note there are other commands :

### `migate:fresh`: 

This **drop all tables** and then run migrations.

> Because tables are dropped, should tenants databases be also dropped instead of deleting the model ? _Schema::dropIfExists($tenant->tenancy_db_name)_ ?

### `migrate:reset`: 

This **rollback** all database migrations. 

> Should this also delete tenants ?

### `migrate:refresh`:

This commands calls the `migrate:reset` command and then migrates.

_Because this calls the migrate:reset command, tenants will be delete so it doesn't need to be overriden_

stancl:
~~Good point about `--seed`, so we probably don't want to just `parent::handle(); Tenant::all()->each->delete()`, but we want to add the `delete()` call *in the middle of* the original logic. So we want to override some other method, ideally one that will require the least code duplication, and add our code there.~~ see below

stein-j:
@stancl I'm not sure I follow, why would you want to call the seeder during the orignal logic ? The seeder is called last, and the original logic handles that.

I think what lukinovec has done is enough `Tenant::all()->each->delete(); parent::handle();`, then if the user wants to pre-deploy a tenant they can simply create a seeder `TenantSeeder` which will be called when using the `--seeder` option.

stancl:
> I'm not sure I follow, why would you want to call the seeder during the orignal logic ? The seeder is called last, and the original logic handles that.

~~I'm not saying *we* should call the seeder. I'm saying that if we execute all of the normal logic (`parent::handle()` — which includes seeding **that may create tenants**), and then we delete all tenants, that it will not work as it should.~~ see below

stancl:
Actually nevermind, it's delete tenants first, and then run original logic.

stancl:
@stein-j 

> So it might be expected by developers that the commande drops each tenant and then creates them again.

No, the point of this feature is to specifically get rid of all tenant databases. If the developer wants to create DBs again, he should just use a seeder. Also, it's literally impossible for us to *know* how to re-create them, since we can't distinguish between "clean data" and data that was added during local testing (in the browser). For the clean data, they should just use seeders.

> Because tables are dropped, should tenants databases be also dropped **instead of deleting the model**

The point of deleting the model is that it deletes the database using event listeners. If we wanted to just delete the tenant as in the DB record, we wouldn't even need this feature, `migrate:fresh` would clear the `tenants` table.

And regarding the other commands, I don't think they're used that often, so it's probably not worth adding this feature for them. But maybe if the code for this feature ends up being simple, we could consider it.

That said, *rollbacks* come with nuances, so deleting tenant DBs might not be desired. `migrate:fresh` is more clean because it simply deletes all data.

Also, one thing I thought of @lukinovec — we should make sure that when `migrate:fresh` is ran for tenants, we don't run the call to delete all tenants.

But it seems that our implementation uses `db:wipe` and `tenants:migrate`, instead of extending/calling `migrate:fresh`, so that probably doesn't need any changes.

lukinovec:
> when migrate:fresh is ran for tenants, we don't run the call to delete all tenants

@stancl, do you mean `$tenant->run(fn() => Artisan::call('migrate:fresh'))` should not delete any tenants in the central DB's `tenants` table (as well as the tenant databases)?

If that's so, then I should look into that because doing that really deletes all tenants (from the central DBs `tenants` table + the tenant DBs get deleted)

EDIT: 
The issue should likely be solvable by checking if tenancy is initialized before deleting the tenants in the command – running only `parent::handle()` would make the comment behave as `migrate:fresh` would normally. But I tried running `$tenant->run(fn() => Artisan::call('migrate:fresh'))` in a simple app (Tenancy `v3.5`, without the changes from this PR), and that produced weird results: only the tenant for which I called the command would get migrated, and when I took a look at the tenant's DB, it had incorrect tables – **the freshly migrated tenant has tables from the central migrations instead of the ones in the `migrations/tenant` folder**.

Also, what's the use case of calling `migrate:fresh` for a tenant? I think if the above keeps happening, we'd best disable `migrate:fresh` in the tenant context and instead suggest `tenants:migrate` (maybe override the command always and throw an exception if the command was called when tenancy is initialized?)

stancl:
> do you mean

I mean the tenant-aware migrate fresh command — `MigrateFresh`. `$tenant->run()` is probably too much of an edge case, so we don't care about that. People should just use our command if they want to migrate fresh tenants.

And as I wrote:
> But it seems that our implementation uses db:wipe and tenants:migrate, instead of extending/calling migrate:fresh, so that probably doesn't need any changes.

src/TenancyServiceProvider.php
stancl: > Sometimes config can't be read this early in the request lifecycle, sometimes you need to do it at a later point.

Similarly to what I wrote here, can't we move the `if ()` check to the `extend()` closure?

If the config key is off, we'll just return the parameter passed to the closure (the original object), otherwise we'll return our own instance.

lukinovec:
I see, then yeah, it's fine

stancl:
Fixed it here https://github.com/archtechx/tenancy/pull/971/commits/607ea723e521a84d0ff0d50714380301c085cade

The solution is to always register the override, and then check the config inside the new class. I spent some time trying to find if we could update the `MigrateFresh` reference in artisan cache, but it seems like all solutions are very messy and complex, possibly depending on Reflection. And then if people wanted to use this logic in their app tests, they'd have to add similar workarounds.

In that commit, I also rewrote the test to be a bit shorter using datasets.

stancl:
Thanks for the work on this guys

PR #979, opened 2022-10-17
[4.x] Maintenance events

stein-j:
This PR adds events when maintenance mode are enabled and disabled.

Down for maintenance mode will fire : 
`\Stancl\Tenancy\Events\TenantMaintenanceModeEnabled::class`

Back from maintenance mode will fire : 
`\Stancl\Tenancy\Events\TenantMaintenanceModeDisabled::class`


stancl:
Looks great, thanks!

PR #980, opened 2022-10-18
Add cookie option on Initialize Tenancy by Request identification

### LINKED ISSUE #895: Add cookie option on Initialize Tenancy by Request identification
#### Description
Add Tenant Identification by **Request Cookie Identification** on `InitializeTenancyByRequestData` middleware payload which can work the same as **Request Header Identification**.

#### Why this should be added
When using `InitializeTenancyByRequestData` 
In a SPA front-end we need to set a header on every request that will be send to our backend or set query params with tenant id.
So in this approach When a user login will receive `X-Tenant` cookie with value `tenant id` from `Authentication Controller` which will be  use on the next requests to identify tenant.

I tested it on my project and it works:

### How it works

on Authentication controller
` $response->withCookie(cookie('X-Tenant', $tenant, 120));`

in `InitializeTenancyByRequestData` middleware from `stancl\tenancy\src\Middleware\InitializeTenancyByRequestData.php`

we add another condition to check if request cookie exist with key `X-Tenant` then we use the value 

```
 protected function getPayload(Request $request): ?string
    {
        $tenant = null;
        if (static::$header && $request->hasHeader(static::$header)) {
            $tenant = $request->header(static::$header);
        } elseif (static::$queryParameter && $request->has(static::$queryParameter)) {
            $tenant = $request->get(static::$queryParameter);
        } 

         // add below condition ...........
        elseif (static::$header && $request->hasCookie(static::$header)) {   // check if request has cookie with key X-Tenant
            $tenant = $request->cookie(static::$header);    
        }      

        return $tenant;
    }

``

### PR DESCRIPTION
**Pull request by abrardev99**
This PR adds the cookie identification using in `InitializeTenancyByRequestData` middleware. 

## Use case

When using `InitializeTenancyByRequestData`In a SPA front-end we need to set a header on every request sent to our backend or set query params with tenant id.
So in this approach, When a user login will receive an X-Tenant cookie with the value tenant id from Authentication Controller, which will be used on the next requests to identify the tenant.

## Usage 
Let's say you have SPA or using the API from any client. 

```php
Route::middleware(InitializeTenancyByRequestData::class)->get('/test', function () {
    return 'Tenant id: ' . tenant('id');
});
```

Now request should have a cookie. 

```php
$response->withCookie(cookie('X-Tenant', $tenant, 120));
```

Closes https://github.com/archtechx/tenancy/issues/895

### DIFF
--- a/tests/RequestDataIdentificationTest.php
+++ b/tests/RequestDataIdentificationTest.php
@@ -20,6 +20,7 @@ beforeEach(function () {
 
 afterEach(function () {
     InitializeTenancyByRequestData::$header = 'X-Tenant';
+    InitializeTenancyByRequestData::$cookie = 'X-Tenant';
     InitializeTenancyByRequestData::$queryParameter = 'tenant';
 });
 
@@ -46,7 +47,7 @@ test('query parameter identification works', function () {
 });
 
 test('cookie identification works', function () {
-    InitializeTenancyByRequestData::$header = 'X-Tenant';
+    InitializeTenancyByRequestData::$cookie = 'X-Tenant';
     $tenant = Tenant::create();
 
     $this
--- a/tests/RequestDataIdentificationTest.php
+++ b/tests/RequestDataIdentificationTest.php
@@ -26,13 +26,11 @@ afterEach(function () {
 test('header identification works', function () {
     InitializeTenancyByRequestData::$header = 'X-Tenant';
     $tenant = Tenant::create();
-    $tenant2 = Tenant::create();
 
     $this
         ->withoutExceptionHandling()
-        ->get('test', [
-            'X-Tenant' => $tenant->id,
-        ])
+        ->withHeader('X-Tenant', $tenant->id)
+        ->get('test')
         ->assertSee($tenant->id);
 });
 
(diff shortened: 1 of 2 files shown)


stancl:
When it comes to cookies, one thing we have to keep in mind is that they come with the problem of domain scoping — cookies set on `foo.com` won't work on `bar.com`.

But here it seems to be fine, since this is essentially meant to be used on a single domain, right?

E.g. `api.acme.com`, but instead of providing a header you provide a cookie. If you used multiple domains for the requests, it wouldn't work, but at that point you'd probably use domain identification anyway if you wanted to use multiple domains.

tests/RequestDataIdentificationTest.php
```diff
@@ -26,24 +26,32 @@
...
-    $tenant2 = Tenant::create();
 
     $this
         ->withoutExceptionHandling()
         ->get('test?tenant=' . $tenant->id)
         ->assertSee($tenant->id);
 });
+
+test('cookie identification works', function () {
+    InitializeTenancyByRequestData::$header = 'X-Tenant';
```
stancl: Why do we set the `$header` in a cookie-related test?

stancl: It'd make more sense to add a new parameter, just like the other two methods have:
https://github.com/archtechx/tenancy/blob/693e00b9be9daf7fb3ada777d5a510f7ec424950/src/Middleware/InitializeTenancyByRequestData.php#L14-L15

abrardev99: > It'd make more sense to add a new parameter, just like the other two methods have:

I was thinking of adding the separate property but then didn't and thought it would be unnecessary. But I think the separate property will make more sense.

abrardev99:
> But here it seems to be fine, since this is essentially meant to be used on a single domain, right?

Yes right. I think one should not use this approach for multiple domains. I agree with you on "but at that point you'd probably use domain identification anyway if you wanted to use multiple domains.". 
Do you think there be any valid use case for using cookies with multi domains?

stancl:
I don't think so. The only case is the issue with https://github.com/archtechx/tenancy/pull/917 which we discussed on BC, but that's for domain identification.

stancl:
Thanks for working on this!

Issue #978, opened 2022-10-15
migrate-fresh command don't consider mysql-schema.dump  file


arivelli:
### Bug description

When I run my migrations the schema files are not considered

### Steps to reproduce

sail artisan tenants:migrate-fresh

### Expected behavior

consider the file mysql-schema.dump

### Laravel version

^9.19

### stancl/tenancy version

3.6

stancl:
- Dumping tenant schemas is not a v3 feature
- We use a custom implementation for migrate:fresh. Why exactly is not using the dump file a problem?

arivelli:
Hey stancl,

Thanks for answer, it's beacuse I have my migration removed and I only have this mysql-schema.dump file, so, whenever I create a new domain I need to run the dump file manually instead of using artisan migrate as usual.

Regards

stancl:
I see, that's fair. We'll take a look at this.

@lukinovec as part of some of the tasks we're doing now, we should verify that all migration commands work even when schema is used instead of migration classes (since the migrations can be deleted after the schema dump is created). I *think* `migrate` can have parameters specified in the config, but `migrate:fresh` probably doesn't respect this.

I'm not sure if we want to use the schema automatically in `migrate`, probably not (so users who generate the schema should then configure it in the config), but `migrate:fresh` should take all of the parameters from the config for the `migrate` call it does.

arivelli:
Thanks Samuel,

In my humble opinion I think it should be as close as possible to how `php artisan migrate` works.

lukinovec:
@stancl, just to confirm, we're talking about `tenants:migrate-fresh` and `tenants:migrate`, not the Laravel commands (`migrate:fresh`) here, right?

> migrate:fresh should take all of the parameters from the config for the migrate call it does

Right, I see that `tenants:migrate` and `tenants:rollback` respect parameters from the `tenancy.migration_parameters` config, but `migrate:fresh` doesn't. `tenants:migrate-fresh` calls  `tenants:migrate` with `--tenants=[$tenantKey]` and `--force=true`, nothing else is passed to the command.

stancl:
Yeah, we probably just want to merge in the params from config

lukinovec:
> tenants:migrate-fresh calls tenants:migrate with --tenants=[$tenantKey] and --force=true, nothing else is passed to the command.

[...]

PR #985, opened 2022-10-25
[4.x] Make `tenants:migrate` default to configured schema path

lukinovec:
This PR solves https://github.com/archtechx/tenancy/issues/978 by adding the `--schema-path` migration command parameter to the `migration_parameters` config. This makes `tenants:migrate` use the dump file to which the path leads to (if the file exists – that's handled by the underlying Laravel command, `tenants:migrate` just passes the path to `migrate` along with the other `migration_parameters`).

The TenantDump command now defaults to the configured path. That solves the issue where the tenant dump file name could be the same as the central DB dump's name ([this BC to-do](https://3.basecamp.com/5170965/buckets/23651018/todos/5445406155#__recording_5466770611)).

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -258,6 +258,7 @@ return [
     'migration_parameters' => [
         '--force' => true, // This needs to be true to run migrations in production.
         '--path' => [database_path('migrations/tenant')],
+        '--schema-path' => database_path('schema/tenant-schema.dump'),
         '--realpath' => true,
     ],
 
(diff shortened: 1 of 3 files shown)


stancl:
Added a change - see the title of the commit. We generally don't want to read these config values as defaults, for two reasons:
- the config key may be missing, and then the value would be empty
- it's better to have the default value as a separate thing from what's defined in the config (or anywhere else). And then use that default in each place, including the config. This separates the default value as its own concept from the config (which only uses the default value)

PR #989, opened 2022-10-31
Make tenants able to have custom mail credentials

lukinovec:
Force reading the mail credentials from the config by replacing the MailManager singleton with a custom class that extends `MailManager` and overrides its `get()` method so that the mailers are always resolved instead of getting them from the `$mailers` property where the mailers are getting cached.

Only SMTP credentials are read from the tenant config by default (see `MailTenancyBootstrapper::$credentialsMap`). You can change that by updating the `MailTenancyBootstrapper::$credentialsMap` and `TenancyMailManager::$mailersToNotCache` static properties. 

Also, we should document that if the user wants to queue the mail, `QueueTenancyBootstrapper`'s `$forceRefresh` property has to be true for TenantConfig to update the config properly.

Another thing to document is that if you resolve the MailManager before the MailTenancyBootstrapper gets executed, the user should extend the MailManager in a service provider.  // I think this doesn't need to be documented now that the TenancyMailManager got removed

--
Docs

# MailTenancyBootstrapper {#mail-tenancy-bootstrapper}

Enabling MailTenancyBootstrapper allows you to use tenant attributes as mailer credentials. To enable the bootstrapper, uncomment it in your `tenancy.php` config file (`tenancy.bootstrappers`).

The bootstrapper maps tenant attributes specified in the customizable `$credentialsMap` static property to config. The following code updates the `mail.mailers.smtp.username` config key to the tenant's `smtp_username` attribute (if the tenant has that attribute) when Tenancy gets bootstrapped:

```php
// In TenancyServiceProvider
public function boot()
{
    ...
    MailTenancyBootstrapper::$credentialsMap = ['mail.mailers.smtp.username' => 'smtp_username',
    ]
}
```

You can also define map presets for different mailers in the static `$mapPresets` property. The property has a SMTP preset by default, but feel free to customize it.

```php
// In TenancyServiceProvider
public function boot()
{
    ...
    // MailTenancyBootstrapper::$mapPresets['mailer_name'] = ['config.key' => 'tenant_attribute_name']
    MailTenancyBootstrapper::$mapPresets['smtp'] = [
        'mail.mailers.smtp.host' => 'smtp_host',
        'mail.mailers.smtp.port' => 'smtp_port',
        'mail.mailers.smtp.username' => 'smtp_username',
        'mail.mailers.smtp.password' => 'smtp_password',
    ];
}
```

By default, the bootstrapper will automatically choose the map preset for your default mailer (`config('mail.default')`) from `$mapPresets` and merge that with `$credentialsMap`. You can make the bootstrapper use a different mailer by updating the static `$mailer` property.

## Queueing mail

If you want to queue the mail, `QueueTenancyBootstrapper`'s `$forceRefresh` property has to be set to `true`. This makes the config update properly and allows the mailer to use the correct credentials.


src/Bootstrappers/MailTenancyBootstrapper.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function bootstrap(Tenant $tenant): void
+    {
+        foreach (static::$credentialsMap as $storageKey => $configKey) {
+            /** @var Tenant&Model $tenant */
+            $override = Arr::get($tenant, $storageKey);
+
+            if (! is_null($override)) {
+                if (is_array($configKey)) {
```
stancl: Do we have a use case for this?

src/Bootstrappers/MailTenancyBootstrapper.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public static array $credentialsMap = [
+        'smtp_transport' => 'mail.mailers.smtp.transport',
+        'smtp_host' => 'mail.mailers.smtp.host',
+        'smtp_port' => 'mail.mailers.smtp.port',
+        'smtp_encryption' => 'mail.mailers.smtp.encryption',
+        'smtp_username' => 'mail.mailers.smtp.username',
+        'smtp_password' => 'mail.mailers.smtp.password',
+        'smtp_timeout' => 'mail.mailers.smtp.timeout',
+        'smtp_local_domain' => 'mail.mailers.smtp.local_domain',
```
stancl: Wouldn't it make more sense to have these in the opposite direction? People may want to customize the *tenant attribute* used for a certain config key, rather than vice versa.

src/TenancyMailManager.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    public static array $mailersToNotCache = [
+        'smtp',
+    ];
+
+    protected function get($name)
+    {
+        if (in_array($name, static::$mailersToNotCache)) {
+            return $this->resolve($name);
+        }
```
stancl: Is `$this->resolve` the logic that re-resolves the mailer?

It'd be good to add comments here, it's not very clear how this class works.

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: Good point, we don't. I think we can probably omit the whole `if (is_array($configKey))`.

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: That makes sense, but it could also confuse people who already know TenantConfig, which has similar logic. Though this doesn't exactly copy the TenantConfig logic – it's different (very slightly), so this might not be an issue

src/TenancyMailManager.php
lukinovec: > Is $this->resolve the logic that re-resolves the mailer?

Yeah.

> It'd be good to add comments here, it's not very clear how this class works.

Right, I'm going to add the comments

lukinovec: Also, `$mailersToNotCache` seems inaccurate. The mailers are still cached in the `$mailers` property, but even when they are cached, they get resolved again. So maybe something like `$mailersToAlwaysResolve` would be more accurate

lukinovec: https://github.com/archtechx/tenancy/pull/989/commits/64aa0c895b6656953a73066b300cb3974741c68b @stancl could you check out the changes I made in this commit?

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: Removed the check https://github.com/archtechx/tenancy/pull/989/commits/d895a30c30a65c7d875ed0f5a7ac5b14c8c3c038

stancl: Rather than copying code from that Feature and putting it here, I would just write the logic from scratch here. The logic in the Feature is *super* specific to support specific syntax etc. Here we just need a simple loop and calls to set config values.

src/Bootstrappers/MailTenancyBootstrapper.php
stancl: There's **tons** of use cases and approaches to changing config in multi-tenant apps. Our TenantConfig feature is *just an option*. It's not standard in any sense, it's not even necessarily recommended for anything, it just exists and can be useful for some things. So I wouldn't try to make it consistent with TenantConfig since that's an unrelated thing. And it just happens to work a certain way.

src/TenancyMailManager.php
stancl: `$mailersToRefresh` or something sounds reasonable. Or even `$tenantMailers`.

I agreed with your previous suggestion (mailers to not cache) since I haven't seen the code and it sounded logical. But here it'd make more sense to make it something like `$tenantMailers` since that's what makes sense to the developer.

"X to not cache" sounds like it's used for *exceptions* or making things exempt. Whereas `$tenantMailers` / `$refreshedMailers` / ... sound like they're used for the *main thing* that this feature does.

lukinovec: Yeah, I like `$tenantMailers`

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: I think simplifying the copied code is fine here – we need to 1) set the config values in `bootstrap`, and 2) revert the config values to the original ones in `revert`. I changed a small thing in the bootstrap method (https://github.com/archtechx/tenancy/pull/989/commits/04ada13584a4bbf1efea1dbe049bb4c8bd8d40a8), and now, I don't think I'd change anything else here

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: Right. I updated the array and the example https://github.com/archtechx/tenancy/commit/4016bc17ba75d89ad4e5344e7cbc31d8e238ee78

src/TenancyMailManager.php
lukinovec: I renamed the variable to `$tenantMailers`. Could you look at the comments in this PR and let me know if you'd change anything?

src/TenancyMailManager.php
```diff
@@ -0,0 +1,49 @@
...
+     * cached and available in the $mailers property).
+     */
+    public static array $tenantMailers = [
+        'smtp',
+    ];
+
+    /**
+     * Override the get method so that the mailers in $tenantMailers
+     * always get resolved, even when they're cached and available in the $mailers property
+     * so that the mailers have up-to-date mailer credentials.
```
stancl: This feels a bit off when you read it in full

CleanShot 2022-11-01 at 8 30 49@2x

The last line can probably be removed entirely

src/TenancyMailManager.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Tenancy swaps Laravel's MailManager singleton for an instance of this class,
+ * which overrides the manager's get method to always resolve
+ * the mailers specified in the static $tenantMailers property
+ * instead of getting them from the $mailers property where they're cached.
+ *
+ * This is mainly used to solve the issue where
+ * mail gets sent with the incorrect (old) mailer credentials
+ * due to the mailer being cached in the manager's $mailers property
+ * and not getting updated when the tenant's credentials change.
```
stancl: This part can probably be removed

src/Bootstrappers/MailTenancyBootstrapper.php
```diff
@@ -0,0 +1,57 @@
...
+    public static array $credentialsMap = [
+        'mail.mailers.smtp.transport' => 'smtp_transport',
+        'mail.mailers.smtp.host' => 'smtp_host',
+        'mail.mailers.smtp.port' => 'smtp_port',
+        'mail.mailers.smtp.encryption' => 'smtp_encryption',
+        'mail.mailers.smtp.username' => 'smtp_username',
+        'mail.mailers.smtp.password' => 'smtp_password',
+        'mail.mailers.smtp.timeout' => 'smtp_timeout',
+        'mail.mailers.smtp.local_domain' => 'smtp_local_domain',
+    ];
```
stancl: I'm looking at this again and I'm thinking that the previous approach could be a bit better for scenarios where you're setting completely new properties (e.g. because you're using a different mailer which has different properties). There setting `'aws_key' => 'mail.mailers.ses.whatever'` would make more sense.

**However**, I think that the scenario where you'd want to customize the properties easily, e.g. `$credentialsMap['mail.mailers.smtp.username'] = 'tenant_mail_username'`, makes more sense to prioritize.

So neither approach is perfect, but this one seems the most reasonable.

Another thing: I would simplify this array. We likely don't want to set things like `timeout` individually for each tenant. What happens if the property doesn't exist on the tenant and this bootstrapper is enabled? Does it simply get skipped (due to line 43)?

And I'm also thinking that it doesn't make sense to make this default to smtp since many people use different things. Maybe we could add static methods that would set the "preset", e.g. one for smtp, one for AWS SES etc.

lukinovec: > What happens if the property doesn't exist on the tenant and this bootstrapper is enabled? Does it simply get skipped (due to line 43)?

If the property is `null` on the tenant, we skip setting the config, so the config value remains the same instead of being mapped to the tenant property (meaning if the tenant doesn't have the property, the default config value, e.g. the one set in a config file will be used).

Maybe checking if the value is `null` isn't right – someone might like to override a config value to make it `null`, but because we're skipping `null` values, the default wouldn't get overridden.

lukinovec: > And I'm also thinking that it doesn't make sense to make this default to smtp since many people use different things. Maybe we could add static methods that would set the "preset", e.g. one for smtp, one for AWS SES etc.

Yeah, that makes sense.

stancl: > Maybe checking if the value is null isn't right – someone might like to override a config value to make it null, but because we're skipping null values, the default wouldn't get overridden.

Check if `hasAttribute()` (or something like that) behaves differently than `getAttribute($foo) !== null`. Ideally the former would return true even on *defined* null. And return false only when the attribute isn't present in $attributes at all.

lukinovec: I couldn't find a method similar to `hasAttribute()`. `array_key_exists($attributeName, $tenant->getAttributes())` works though

assets/config.php
```diff
@@ -83,6 +83,7 @@
         Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\QueueTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\BatchTenancyBootstrapper::class,
+        Stancl\Tenancy\Bootstrappers\MailTenancyBootstrapper::class, // Queueing mail requires using QueueTenancyBootstrapper with $forceRefresh set to true
```
lukinovec: I found out that while testing, TenancyServiceProvider was getting registered _before_ adding the bootstrapper to the `tenancy.bootstrappers` config. Because of that, the bootstrapper wasn't stateful (it didn't get bound as a singleton), and the config wasn't getting reverted properly.

I tried if uncommenting the bootstrapper in the config will fix that issue, and it did. I also couldn't think of a good reason to comment out `MailTenancyBootstrapper`.

@stancl, do you think there's a good reason to keep the bootstrapper commented out? Also, do you know of a solution for adding and registering the bootstrappers in tests properly? I thought about manually making the bootstrapper a singleton, but I wasn't sure if making the bootstrapper a singleton is everything that's happening to the bootstrappers in the provider's `register` method.

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: > We likely don't want to set things like timeout individually for each tenant

I agree with this. I made the SMTP preset have only `username`, `password`, `host`, and `port`. These looked like the ones that'd get mapped commonly (not sure about host and port, but I saw someone on Discord trying to map these, so I though these might be common).

> Maybe we could add static methods that would set the "preset", e.g. one for smtp, one for AWS SES etc.

[I updated the code related to this](https://github.com/archtechx/tenancy/pull/989/files#diff-421166a13b8e50c082716efce511eec8587457ce0cac287118d5de829b4a4980R23-R46), could you take a look? I think instead of having a method per mailer, I'd like a big array with the presets which would look like this:
```php
public static $mapPresets = [
    'smtp' => [
        'mail.mailers.smtp.host' => 'smtp_host',
        'mail.mailers.smtp.port' => 'smtp_port',
        'mail.mailers.smtp.username' => 'smtp_username',
        'mail.mailers.smtp.password' => 'smtp_password',
    ],
    'ses' => [
        ...
    ]
]
```
Then, we could delete the `pickMapPreset` method, and instead of using it, we could just do `static::$mapPresets[$config->get('mail.default')] ?? []` in the constructor.

src/TenancyMailManager.php
```diff
@@ -0,0 +1,41 @@
...
+     * even when they're cached and available in the $mailers property.
+     */
+    protected function get($name)
+    {
+        if (in_array($name, static::$tenantMailers)) {
+            return $this->resolve($name);
+        }
+
+        return parent::get($name);
+    }
```
lukinovec: I deleted the `smtp` default from `$tenantMailers` because if someone's app is only using e.g. `ses`, having `smtp` explicitly named as a tenant mailer by default wouldn't make much sense.

This made me think that the re-resolving could be done for all mailers by default, and to utilize caching of some mailers, we could be naming them explicitly in a `$mailersToGetFromCache` array. So it'd be the other way around – mailers in the array would not always get resolved, and the `get` method would look like this:
```php
if (in_array($name, static::$mailersToGetFromCache)) {
    return parent::get($name);
}

return $this->resolve($name);
```
(swap `return` in the `if` check with the other one)

@stancl, which approach seems better to you?

src/Bootstrappers/MailTenancyBootstrapper.php
stancl: I think this:

```php
public static function smtpPreset(): void
{
    $this->setMapping([
        'mail.mailers.smtp.host' => 'smtp_host',
        'mail.mailers.smtp.port' => 'smtp_port',
        'mail.mailers.smtp.username' => 'smtp_username',
        'mail.mailers.smtp.password' => 'smtp_password',
    ]);
}
```
would be better (with `setMapping()` being a simple method that sets the values; or maybe you can just directly modify the mapping as a property).

That way we don't make assumptions about the default mailer, and if people want to extend it to change the preset, they can easily do that.

But maybe a static property has its benefits too. That way people could just do `::$mapPresets['smtp'] = $theirPreset` without having to extend anything. So I guess the current approach is good, I'd just make it a bit less dependent on the default mailer. Maybe add an option to specify the mailer (probably some other static property).

assets/config.php
stancl: I think we bind the Redis bootstrapper as a singleton somewhere. Could that be a similar case to this?

src/TenancyMailManager.php
stancl: I think it's better to specify the mailers that *should* be re-resolved.

assets/config.php
lukinovec: I found where the Redis bootstrapper gets bound as a singleton (https://github.com/archtechx/tenancy/blob/master/tests/TestCase.php#L115 – the line also has a todo that says `use proper approach eg config for singleton registration`), and yeah, that seems like a very similar case

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: > So I guess the current approach is good, I'd just make it a bit less dependent on the default mailer. Maybe add an option to specify the mailer

I added the static property https://github.com/archtechx/tenancy/pull/989/commits/55aa49b18e7c2b227751a0e66bd0c0d35111cfb4

stancl: I wrote:
> But maybe a static property has its benefits too. That way people could just do ::$mapPresets['smtp'] = $theirPreset without having to extend anything. So I guess the current approach is good, I'd just make it a bit less dependent on the default mailer

And it seems like we're using the method approach now. Do you have some reason for preferring that over a property?

assets/config.php
stancl: Try using the same solution. And then I'll try to fix how those two work at once.

src/Bootstrappers/MailTenancyBootstrapper.php
lukinovec: I prefer the property, I didn't change the method approach because you wrote "So I guess the current approach is good, I'd just make it a bit less dependent on the default mailer". I see you probably meant the approach I was writing about [here](https://github.com/archtechx/tenancy/pull/989#discussion_r1011849029)

stancl: By current approach I meant the `$mapPresets` property, yeah. Didn't realize you changed it to a method before I described the method approach.

lukinovec: Updated the code to use the static property now https://github.com/archtechx/tenancy/pull/989/commits/a0ad69a95cbd68bc49d46966d7623f64927a8d2f

stancl:
@lukinovec is this fully ready for review now? If so, you can mark the open reviews as resolved.

assets/config.php
lukinovec: https://github.com/archtechx/tenancy/pull/989/commits/62619e6738d45e0dbd2f1de627ed86f40a0f05f8

src/TenancyServiceProvider.php
```diff
@@ -69,6 +70,12 @@ public function register(): void
             return new Commands\Seed($app['db']);
         });
 
+        // Use custom mail manager that resolves the mailers specified in its $tenantMailers static property
+        // Instead of getting the cached mailers from the $mailers property
+        $this->app->extend(MailManager::class, function (MailManager $mailManager) {
+            return new TenancyMailManager($this->app);
+        });
```
stancl: I wonder if this could be moved to the bootstrapper? So that we don't unnecessarily override the class (which could conflict with people's custom extended `MailManager`s when they don't want to use this feature)

tests/BootstrapperTest.php
```diff
@@ -326,6 +327,34 @@
...
+        expect(config('mail.mailers.smtp.username'))->toBe($defaultUsername);
+        expect(config('mail.mailers.smtp.password'))->toBe(tenant()->smtp_password);
+    });
+});
+
+test('MailTenancyBootstrapper reverts the config to default when tenancy ends', function() {
+    MailTenancyBootstrapper::$credentialsMap = ['mail.mailers.smtp.password' => 'smtp_password'];
+    config(['mail.mailers.smtp.password' => $defaultPassword = 'no password']);
+
+    Tenant::create(['smtp_password' => 'testing password'])->run(fn() => '');
```
stancl: Make the fact that Tenancy is initialized and ended more clear. The `run()` is easy to overlook.

Use `initialize()`, then assert that the tenant password is being used, and then call `end()` and assert that we're back to the central password.

tests/MailTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    $tenant = Tenant::create();
+
+    // Initialize tenancy as $tenant and assert that the smtp mailer's transport has the correct password
+    $assertTransportUsesPassword = function(string|null $password) use ($tenant) {
+        tenancy()->initialize($tenant);
+
+        $manager = app(MailManager::class);
+
+        $getMailerViaManager = new ReflectionMethod($manager::class, 'get');
```
stancl: I think I added `invade()` to the testing dependencies. Could that help you refactor this?

src/TenancyServiceProvider.php
lukinovec: I like that idea. It works with a small change – tenancy has to be initialized for the bootstrapper to be bootstrapped (so that MailManager resolves as TenancyMailManager), which makes sense to me.

https://github.com/archtechx/tenancy/pull/989/commits/d83cd1ddb482aa99c4d08d9d75c6a65a0b6bd20b

tests/BootstrapperTest.php
lukinovec: Updated the test https://github.com/archtechx/tenancy/pull/989/commits/b96ff262ef14de1fd8a570a2839ce877e6e955a2

tests/MailTest.php
lukinovec: Yeah, `invade()` made it much cleaner. I didn't know about that helper

https://github.com/archtechx/tenancy/pull/989/commits/b3cef0c100e4907c1f0d334059c8c2114b90c683

stancl: Yeah, it's from a Spatie package.

I think we can remove the Reflection-related imports now as well?

src/TenancyServiceProvider.php
stancl: Yeah, that is correct.

I guess moving it to the bootstrapper comes with a _small_ issue — there may be an edge case where people interact with the mailer (or try to inject it somewhere, thus triggering its instantiation from the container) before the bootstrapper is executed. But if that ever comes up, we can just tell people to put this in a service provider.

tests/MailTest.php
lukinovec: There are no imports, the Reflection classes didn't need to be imported, so I think we can resolve this

Also, we can refactor [the getDiskPrefix method in BootstrapperTest](https://github.com/archtechx/tenancy/blob/cf5bffea258b59a2c6a80426a371fcceaec30871/tests/BootstrapperTest.php#L362) to use `invade()` too (maybe in another PR)

src/TenancyServiceProvider.php
lukinovec: > if that ever comes up, we can just tell people to put this in a service provider

Agree

lukinovec: > if that ever comes up, we can just tell people to put this in a service provider

For the record, I'm resolving this because I added a note about mentioning that in the PR's description

tests/MailTest.php
stancl: Ah I see, tests are in the global namespace. And yeah, feel free to get rid of all Reflection code in tests in another PR 👍🏻

tests/BootstrapperTest.php
```diff
@@ -326,6 +327,38 @@
...
+    MailTenancyBootstrapper::$credentialsMap = ['mail.mailers.smtp.password' => 'smtp_password'];
+    config(['mail.mailers.smtp.password' => $defaultPassword = 'no password']);
+
+    tenancy()->initialize(Tenant::create(['smtp_password' => $tenantPassword = 'testing password']));
+
+    expect(config('mail.mailers.smtp.password'))->toBe($tenantPassword);
+
+    tenancy()->end();
+
+    expect(config('mail.mailers.smtp.password'))->toBe($defaultPassword);
```
stancl, line 368: Do you think we want to use `invade()` in this these tests as well? To ensure that it's not reflected just in the config, but also on the mailer?

stancl:
Could you write up a simple version of the docs draft and include that in the PR description? So that I can review the usage.

The code itself looks great now, I just want to be sure about the approach we used for specifying the config overrides.

tests/BootstrapperTest.php
lukinovec: Yeah I think that makes sense

stancl: Though then we should be clear about the difference between these tests and the tests in the newly added test file.

lukinovec: I think BootstrapperTest could test that using the MailTenancyBootstrapper makes the key things the bootstrapper provides work (mapping the tenant attributes to config and reverting the config when tenancy ends, making the mailer use the tenant credentials, and making `app(MailManager::class)` resolve TenancyMailManager). I agree with what you wrote on BC ("Usually the BootstrapperTest ensures that everything works okay from the end user's POV (so here we check the config & the mailer), and the dedicated test file checks for edge cases and low-level details to make sure that everything is perfectly covered.")

So because the current MailTest tests only test the core things that MailTenancyBootstrapper provides, we could move the tests to BootstrapperTest. We could also keep the MailTest file to test the more under-the-hood things, like ensuring that TenancyMailManager re-resolves the cached mailers correctly.

stancl: I guess having some duplication is fine if the version in the MailTest file is more in-depth. Meaning, some tests can be similar with the BootstrapperTest version only asserting that the config keys & mailer properties are correct. And the MailTest verifying all the low-level details. And possibly that reverting (or switching multiple times between tenant & central, or between different tenants directly) works well.

So the BootstrapperTest could be simple like the ones that are in the file, just checking *that it works*, and the MailTest would check all of the details of how it works if that makes sense.

lukinovec: I was testing the re-resolving and realized we actually don't even need that in the end.

I think the `get()` method overriding was primarily there to ensure the mailer credentials got updated in real-time (when the config and tenant credentials got updated).

Later, I learned that because the config updates only when tenancy gets bootstrapped, re-resolving should have no real effect, but because removing the `get()` method override would make the central app's mailer have the password of the previous tenant if tenancy got canceled using `tenancy()->end()`, I kept the override.

Now, I tried removing the override again, and I realized that after ending tenancy, the tenant credentials persist because the `TenancyMailManager` instance (which stores the mailers and their credentials) we bind in the mail bootstrapper stays bound to `MailManager::class` even after ending tenancy.

I suggest deleting the `TenancyMailManager` class and updating `MailTenancyBootstrapper` so that it just binds a new `MailManager` instance to `MailManager::class` (instead of new `TenancyMailManager`), and so that it does the same thing in `revert()` (`app()->forgetInstance(MailManager::class)` didn't work, but extending `MailManager::class` the same way as in `bootstrap()` did).

I pushed the changes https://github.com/archtechx/tenancy/pull/989/commits/fcd71f0b63bf1d850e8ed3ea8511d91c42301f50

lukinovec: I also updated the tests

lukinovec:
@stancl check the PR description, I added the documentation

tests/BootstrapperTest.php
stancl: So is everything here done now and ready for review?

One question about the implementation: couldn't we just use `forgetInstance()` in both places, like we use `bindNewMailManagerInstance()` now?

Using `extend()` repetitively seems like a bad way to implement this, I think each `extend()` registers a new "extension", so the closure will run several times, and also we're kind of abusing `extend()` here.

Getting rid of the override is definitely good, though I'd look for a better solution than registering `extend()` Closures an unlimited number of times for the same binding.

lukinovec: I realized I could try forgetting `mail.manager` instead of `MailManager::class`, and that worked. So I got rid of `bindNewMailManagerInstance()` and used `forgetInstance()` instead.

> So is everything here done now and ready for review?

Yeah

stancl:
Related task: Could you go through all of these PRs where we added docs in the PR description, and create issues in stancl/tenancy-docs that link to those PRs? So that we don't forget about any when we write the v4 docs.

lukinovec:
Created an issue with the docs for this PR https://github.com/stancl/tenancy-docs/issues/228

PR #992, opened 2022-11-04
[4.x] Resource syncing improvements 

abrardev99:
This PR adds minor improvements in resource syncing code. It continues with #915. 

As explained in this review https://github.com/archtechx/tenancy/pull/915#discussion_r1013168413

>So we want to:
Use Tenant instead of Resource in all suffixed (e.g. X Providing ...) models
Either keep ResourceUser as it is, or rename it to ResourceTenantUser. Or maybe just add a comment saying // Tenant users

So this PR does the followings: 

1. Rename all models starting with `Resource...` to `Tenant`. 
2. Update test names accordingly. 
3. ~~Also, rename ResourceUser to TenantUser.~~
4. ~~Improve some styling and remove redundant comments.~~


tests/ResourceSyncingTest.php
```diff
@@ -848,7 +849,7 @@ public function getSyncedAttributeNames(): array
     }
 }
 
-class ResourceUser extends Model implements Syncable
+class TenantUser extends Model implements Syncable
```
stancl: CleanShot 2022-11-04 at 1 23 47@2x

abrardev99: I read that. But I though renaming it to `TenantUser` making it more clear. Should I revert it?

stancl: Hmm, it depends. `TenantUser` is the most accurate naming, but the most problematic one since it has the highest chance of having conflicts. Seems like we're not using the class anywhere else though. So I guess we can keep it `TenantUser` for now and only deal with the "scoped" naming later if needed.

stancl: Actually, `TenantUser` sounds like a model for `tenant_users`, i.e. the pivot table. So maybe we want to avoid the name in either case.

So I guess `ResourceUser` with a `// Tenant users` comment above the class definition works best.

stancl: Abrar, why are you renaming the other classes as well? Can you please carefully read the screenshot I linked?

https://github.com/archtechx/tenancy/pull/992/commits/f9ba778e1b6da7f19a191e68b07248b4dee1ddc3#diff-0c32d140e95604d5ab8852beb5eba2255180184351adfaa0ff7b2895012d4d7dR889-R918

I said that all classes named `ResourceUserProviding...` should be `TenantUserProviding...`. And in the first highlighted paragraph I said that **only for the *one* case of `ResourceUser` we may consider a different approach**.

It seems like you basically reverted the entire PR in the last commit.

abrardev99: >Abrar, why are you renaming the other classes as well? Can you please carefully read the screenshot I linked?

I see. It's my fault. I used PHPStorm for refactoring, which did this. I ran tests and pushed the code. I thought PhpStorm renamed the class and its usage. Looked like it does rename everything starting with this name. Weird.

stancl: You can just always check the diff before pushing changes. It's a good practice in general since it shows if any unexpected files were created/changed, or if you have some debugging changes that you don't want to push.

stancl:
@abrardev99 The PR is showing that there are conflicts, but when you click "Resolve conflicts" it says that there are zero. Could you try merging `master` into this branch (using CLI) and seeing if that causes any conflicts?

abrardev99:
> Could you try merging master into this branch (using CLI) and seeing if that causes any conflicts?

Unable to merge using CLI. 
image

But when I try to merge using GUI app (GitHub desktop) it doesn't show any conflicts. 
image

abrardev99:
Looks like PhpStorm showed some conflicts: 
image

But when looking into them, I don't see any change in comparison.

stancl:
Try to resolve this locally then, run tests, and if it passes, push to this branch.

abrardev99:
> Try to resolve this locally then, run tests, and if it passes, push to this branch.

Tests are failing. I'm looking into them.

abrardev99:
@stancl Tests are failing in the `PendingTenantTest` file. It's totally irrelevant to PR conflicts because all of the conflicts were in (resource syncing tests). 
All of the tests show this error: 
image

Do you think this method got removed in `master`?

stancl:
Tests are passing after the merge, so that probably doesn't matter now?

And `getColumnForQuery()` was added in a dependency.

abrardev99:
Jeez. Tests are passing in CI. So mostly probably it is a cache/composer dump issue on my side.

stancl:
Yes, a `composer update` should fix it.

stancl:
Thanks! If you want to PR the formatting improvements as well, you can make another PR 👍🏻

PR #993, opened 2022-11-04
[4.x] Add feature to ignore the resource syncing based on provided condition

abrardev99:
Related PR #889. 

This PR adds the ability to ignore resource synchronization on demand. This adds the new method `shouldSync` which returns `true` by default. You can specify this method in a model like: 
```php
public function shouldSync(): bool
{
    return false;
}
```

You can specify any condition here: 
```php
public function shouldSync(): bool
{
    return $this->sync_me == 'yes'; // can be any condition
}
```

This is a useful feature when you want to ignore the sync on the go or based on condition. The attached PR has an excellent example.

### DIFF
--- a/tests/ResourceSyncingTest.php
+++ b/tests/ResourceSyncingTest.php
@@ -763,6 +763,25 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
     expect(ResourceUser::first()->role)->toBe('commenter');
 }
 
+test('resources are synced only when sync is enabled', function () {
+    [$tenant1, $tenant2] = createTenantsAndRunMigrations();
+
+    tenancy()->initialize($tenant1);
+
+    TenantUserWithDisabledSync::create([
+        'global_id' => 'absd',
+        'name' => 'John Doe',
+        'email' => 'john@localhost',
+        'password' => 'password',
+        'role' => 'commenter',
+    ]);
+
+    tenancy()->end();
+
+    expect(CentralUser::all())->toHaveCount(0);
+
+})->group('current');
+
 /**
  * Create two tenants and run migrations for those tenants.
  */
@@ -977,5 +996,21 @@ class ResourceUserProvidingMixture extends ResourceUser
             'role' => 'admin',
             'password' => 'secret',
         ];
+    }
+}
+
+class CentralUserWithDisabledSync extends CentralUser
+{
+    public function shouldSync(): bool
+    {
+        return false;
+    }
+}
+
+class TenantUserWithDisabledSync extends CentralUser
+{
+    public function shouldSync(): bool
+    {
+        return false;
     }
 }
--- a/tests/ResourceSyncingTest.php
+++ b/tests/ResourceSyncingTest.php
@@ -763,15 +763,15 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
     expect(ResourceUser::first()->role)->toBe('commenter');
 }
 
-test('resources are synced only when sync is enabled', function () {
+test('resources are synced only when sync is enabled', function (bool $enabled) {
+    app()->instance('_tenancy_test_shouldSync', $enabled);
+
     [$tenant1, $tenant2] = createTenantsAndRunMigrations();
     migrateUsersTableForTenants();
 
     tenancy()->initialize($tenant1);
 
-    // Tenant model returns false from `shouldSync`
-    // central model will not be created
-    TenantUserWithDisabledSync::create([
+    TenantUserWithConditionalSync::create([
         'global_id' => 'absd',
         'name' => 'John Doe',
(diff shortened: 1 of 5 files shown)


tests/ResourceSyncingTest.php
```diff
@@ -763,6 +763,45 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
...
+
+    // central model returns false from `shouldSync`
+    // resource model will not be created
+    $centralUser->tenants()->attach('t2');
+
+    $tenant2->run(function () {
+        expect(ResourceUser::all())->toHaveCount(0);
+        expect(ResourceUser::whereGlobalId('acme')->first())->toBeNull();
+    });
+})->group('current');
```
stancl: What's the point of the group?

tests/ResourceSyncingTest.php
```diff
@@ -763,6 +763,45 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
     expect(ResourceUser::first()->role)->toBe('commenter');
 }
 
+test('resources are synced only when sync is enabled', function () {
```
stancl: The previous PR had two tests. Did you combine their logic in this test? To test both scenarios.

abrardev99: > Did you combine their logic in this test? 

Yes.

tests/ResourceSyncingTest.php
abrardev99: Oh sorry. Debugging thing.

tests/ResourceSyncingTest.php
```diff
@@ -763,6 +763,45 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
...
+        'role' => 'commenter',
+    ]);
+
+    // central model returns false from `shouldSync`
+    // resource model will not be created
+    $centralUser->tenants()->attach('t2');
+
+    $tenant2->run(function () {
+        expect(ResourceUser::all())->toHaveCount(0);
+        expect(ResourceUser::whereGlobalId('acme')->first())->toBeNull();
```
stancl: I think it doesn't make sense to be interacting with `ResourceUser` and `CentralUser` in this test. You defined new models, so you should be using exclusively those. Or is there any reason to use these specifically?

(I just want an answer here, not code changes, I'm working on rewriting the tests myself)

abrardev99: > You defined new models, so you should be using exclusively those. Or is there any reason to use these specifically?

I'm using them for `creation` only.  There is no specific reason. We can use newly defined models here too.

stancl: I see. Try to default to making things consistent unless there's a **specific** reason to introduce inconsistency such as using different models for the same thing.

tests/ResourceSyncingTest.php
```diff
@@ -763,6 +763,43 @@ function creatingResourceInTenantDatabaseCreatesAndMapInCentralDatabase()
     expect(ResourceUser::first()->role)->toBe('commenter');
 }
 
+test('resources are synced only when sync is enabled', function (bool $enabled) {
+    app()->instance('_tenancy_test_shouldSync', $enabled);
+
+    [$tenant1, $tenant2] = createTenantsAndRunMigrations();
```
stancl, line 769: Is there a specific reason for using two tenants in this test? Instead of just one.

abrardev99: I created two tenants when I was testing both sides. Eg creating central and asserting tenant, then creating tenant and asserting central. 

But since you refactored it to the dataset, We can stay with one tenant. You can remove it.

stancl: I don't think the dataset changes anything here. The code is the same, I just run it with both settings.

But I'm asking about why two tenants are used. Because it seems like the code was using the second tenant just for the `attach()` call, but is there a reason to use tenant2 instead of tenant1 there?

The test in general doesn't test anything about syncing *between tenants*, it just tests syncing between central & tenant contexts. So I'm not sure why two tenants are used.

abrardev99: > but is there a reason to use tenant2 instead of tenant1 there?

Yes, there is. 
We are using different tenants for both cases. Because If we use the same tenant, CentralUser would have two records when using the tenant1 for both cases. 
For that, at the bottom of the test, we would have `expect(ResourceUser::all())->toHaveCount(1);` instead of `expect(ResourceUser::all())->toHaveCount(0);`.  It feels more natural to assert count 0 (which means resource not created) instead of  `expect(ResourceUser::all())->toHaveCount(1);`  which means the central user created previously and the count should be 2. Makes sense?

stancl: I see, yeah that makes sense.

PR #997, opened 2022-11-08
[4.x] Use polymorphic table for mapping resources to tenants

abrardev99:
This PR enables the use of the polymorphic relationship for mapping resources to tenants. See the docs on how resource syncing is working. 
Let's understand this with an example. Assume you have two resources to sync, e.g., `User` and `Company`. Before this PR you need two pivot tables. `tenant_users`, and `tenant_companies`, which are being used to set up relations.

This PR adds support to using the single table using a polymorphic approach instead of multiple tables. 

> Note: You can still go with the older approach. It's still supported. 

## Using polymorphic table/relationship approach 
The `ResourceSyncing` trait provides the implementation for the `tenants` method. So, you don't need to write the method in your models when using this approach. This method depends on the `Tenant` model fetched from the config or default to `Stancl/Database/Model/Tenant`. The developer can also override using a method `getResourceTenantModelName` to provide the tenant model if it is different from the default, which is what we are doing in the tests of this PR. 

## High-Level Idea
Considering the above User and Company example, the high-level structure of the example can be:
```php
Tenant
    id - string

User 
    id - integer
    global_id - string
    name - string
    email - string

Company 
    id - integer
    global_id - string
    name - string

tenant_resources
    id - integer
    tenant_id - string
    resource_global_id - string
    resource_type - string 
```

The mapping will be stored as follows: 
image

## Usage 

Tenant model
```php
class TenantUsingPolymorphic extends Tenant
{
    public function users(): MorphToMany
    {
        return $this->morphedByMany(CentralUserUsingPolymorphic::class, 'tenant_resources', 'tenant_resources', 'tenant_id', 'resource_global_id', 'id', 'global_id')
            ->using(TenantPivot::class);
    }

    public function companies(): MorphToMany
    {
        return $this->morphedByMany(CentralCompanyUsingPolymorphic::class, 'tenant_resources', 'tenant_resources', 'tenant_id', 'resource_global_id', 'id', 'global_id')
            ->using(TenantPivot::class);
    }
}
```

CentralUser model 
```php
class CentralUserUsingPolymorphic extends Model implements SyncMaster
{
    use ResourceSyncing, CentralConnection;

    protected $guarded = [];

    public $timestamps = false;

    public $table = 'users';

    {
        return TenantUserUsingPolymorphic::class;
    }

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return static::class;
    }

    {
        return [
            'global_id',
            'name',
            'password',
            'email',
        ];
    }
}
```

TenantUser model
```php
class TenantUserUsingPolymorphic extends Model implements Syncable
{
    use ResourceSyncing;

    protected $table = 'users';

    protected $guarded = [];

    public $timestamps = false;

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return CentralUserUsingPolymorphic::class;
    }

    {
        return [
            'global_id',
            'name',
            'password',
            'email',
        ];
    }
}
```

CentralCompany model 
```php
class CentralCompanyUsingPolymorphic extends Model implements SyncMaster
{
    use ResourceSyncing, CentralConnection;

    protected $guarded = [];

    public $timestamps = false;

    public $table = 'companies';

    {
        return TenantCompanyUsingPolymorphic::class;
    }

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return static::class;
    }

    {
        return [
            'global_id',
            'name',
            'email',
        ];
    }
}
```

ResourceCompany model
```php
class TenantCompanyUsingPolymorphic extends Model implements Syncable
{
    use ResourceSyncing;

    protected $table = 'companies';

    protected $guarded = [];

    public $timestamps = false;

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return CentralCompanyUsingPolymorphic::class;
    }

    {
        return [
            'global_id',
            'name',
            'email',
        ];
    }
}
```

The syntax for attaching central resources to tenants `$centralUser->tenants()->attach('t1');`. 

## Using separate table/relationship approach (old approach)
As the old approach is still supported and here is how you can use it using the same User and Company example. 

Tenant model
```php
class Tenant extends BaseTenant
{
    public function users(): BelongsToMany
    {
        return $this->belongsToMany(CentralUser::class, 'tenant_users', 'tenant_id', 'global_user_id', 'id', 'global_id');
    }

    public function companies(): BelongsToMany
    {
        return $this->belongsToMany(CentralCompany::class, 'tenant_companies', 'tenant_id', 'global_user_id', 'id', 'global_id');
    }
}
```

Central User model. 
```php
class CentralUser extends Model implements SyncMaster
{
    use ResourceSyncing, CentralConnection;

    protected $guarded = [];

    public $timestamps = false;

    public $table = 'users';

    // override tenants method 
    {
        return $this->belongsToMany(Tenant::class, 'tenant_users', 'global_user_id', 'tenant_id', 'global_id')
            ->using(TenantPivot::class);
    }

    {
        return TenantUser::class;
    }

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return static::class;
    }

    {
        return [
            'global_id',
            'name',
            'password',
            'email',
        ];
    }
}
```

Tenant User model 
```php
class TenantUser extends Model implements Syncable
{
    use ResourceSyncing;

    protected $table = 'users';

    protected $guarded = [];

    public $timestamps = false;

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return CentralUser::class;
    }

    {
        return [
            'global_id',
            'name',
            'password',
            'email',
        ];
    }
}
```

Central Company model. 
```php
class CentralCompany extends Model implements SyncMaster
{
    use ResourceSyncing, CentralConnection;

    protected $guarded = [];

    public $timestamps = false;

    public $table = 'users';

    // override tenants method 
    {
        return $this->belongsToMany(Tenant::class, 'tenant_companies', 'global_user_id', 'tenant_id', 'global_id')
            ->using(TenantPivot::class);
    }

    {
        return TenantUser::class;
    }

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return static::class;
    }

    {
        return [
            'global_id',
            'name',
            'email',
        ];
    }
}
```

Tenant Company model 
```php
class TenantCompany extends Model implements Syncable
{
    use ResourceSyncing;

    protected $table = 'users';

    protected $guarded = [];

    public $timestamps = false;

    public function getGlobalIdentifierKey(): string|int
    {
    }

    {
        return 'global_id';
    }

    {
        return CentralCompany::class;
    }

    {
        return [
            'global_id',
            'name',
            'email',
        ];
    }
}
```

Notice we are overriding the `tenants` method and providing our relationship using separate tables (`tenant_users`, `tenant_companies`) for each resource with `TenantPivot` model. 

#### Customization 
1. You can publish the pivot table migrations using the `php artisan vendor:publish --tag=resource-syncing-migrations` command. 
2. Package provides the relationship method `tenants()` out of the box using a single table approach. If you'd like to customize it, you can override it in your models. For example, you can see these [tests](https://github.com/archtechx/tenancy/blob/e2ca635ff89aec72270b927ba034ee0f370c1b6c/tests/ResourceSyncingTest.php#L843-L847) doing the same. 
3. `tenants()` method needs `Tenant` model. Out of the box, it uses the `tenancy.tenant_model` value.
4. PR added the new model `TenantMorphPivot`, used explicitly for this approach. There is still the `TenantPivot` model, which can be used with the older approach.


stancl:
Would it be possible to support both approaches? I.e. table for each relation, and a single polymorphic table.

tests/ResourceSyncingTest.php
```diff
@@ -149,7 +150,7 @@
     });
 
     // When central model provides the list of attributes, resource model will be created from the provided list of attributes' values
-    $centralUser->tenants()->attach('t1');
+    $centralUser->resources()->attach('t1');
```
stancl: The central user *is the resource*. Why would we attach the tenant to `resources()`? Is the name change needed?

It doesn't make sense to attach a tenant as a resource to a resource. We want to add a tenant to `tenants()`.

abrardev99: Yes, it doesn't make sense. I renamed it to `tenants`.

abrardev99:
> Would it be possible to support both approaches? I.e. table for each relation, and a single polymorphic table.

@stancl Yes, of course. 

Developers can always override the method in their model. 
Using a single polymorphic table: 
```php
public function tenants(): MorphToMany
{
    return $this->morphToMany($this->getResourceTenantModelName(), 'tenant_resources', 'tenant_resources', 'resource_global_id', 'tenant_id', 'global_id')
        ->using(TenantPivot::class);
}
```

Using the table for each relation: 
```php
{
    return $this->BelongsToMany($this->getResourceTenantModelName(),  'tenant_users', 'users_global_id', 'tenant_id', 'global_id')
        ->using(TenantPivot::class);
}
```

Here are a few considerations: 
1. Notice the method return type. So the method in the interface should allow returning both types, eg `BelongsToMany|MorphToMany`. 
2. `TenantPivot` model needs to extend from a different model. eg `Pivot` or `MorphPivot`. So we need to find a way how to achieve this. Maybe separate models for each approach.

abrardev99:
Questions: 
1. Should we ship the `tenant_resources ` migrations with the package (with the ability to publish)? 
2. I created a separate test file for this PR mainly because the ResourceSyncingTest file is getting too long. Is it okay?

stancl:
1. Yes, I think it's a good default (compared to the old approach), so we should also make the migration publishable. It shouldn't be published with the default migrations during installation though, there should be a special tag.
2. Yeah, the separate file is okay.

stancl:
And I was thinking about how we'll solve the typehints yesterday. I thought that if I remember correctly, `MorphToMany` extends `BelongsToMany`. I checked that now, and it indeed does. So we can just use the `BelongsToMany` type.

stancl:
And what do you mean by point 2) in https://github.com/archtechx/tenancy/pull/997#issuecomment-1308560827? Where would that be done?

FYI `MorphPivot` also extends `Pivot` if that's relevant here.

abrardev99:
> Where would that be done?

Here https://github.com/archtechx/tenancy/pull/997/files#diff-c2d7d418b5a69a37e68e93551a4a5e1bffc7bbd9190303ce327c5e507fde973cR10

As you are saying "MorphPivot also extends Pivot" then that's fine. We don't need to change anything. Let me test.

stancl:
Well I think if people used `TenantPivot` in their app, but used a polymorphic relationship, then they should be using something that extends `MorphPivot`. `MorphPivot` presumably has extra stuff that's important for the polymorphic relationship. So we can just provide two separate classes.

But yeah, test it first.

abrardev99:
> Well I think if people used `TenantPivot` in their app, but used a polymorphic relationship, then they should be using something that extends `MorphPivot`. `MorphPivot` presumably has extra stuff that's important for the polymorphic relationship. So we can just provide two separate classes.
> 
> But yeah, test it first.

Tests are passing with this single class extending from `MorphPivot`.

stancl:
Thanks for testing that. Though I guess wouldn't use a class extending `MorphPivot` for `Pivot` relationships because that's semantically wrong. It's probably best to just have two classes 👍🏻

abrardev99:
@stancl you can review PR.

stancl:
Add all usage-related info to the PR description, including publishing the migrations etc.

stancl:
@abrardev99 Add instructions for using the old approach to the PR description. Since the trait now uses the morph classes by default, it should be clearly explained how people can change this to use the old relationship structure.

src/Database/Models/TenantMorphPivot.php
```diff
@@ -0,0 +1,22 @@
...
+    public static function booted(): void
+    {
+        static::saved(function (self $pivot) {
+            $parent = $pivot->pivotParent;
+
+            if ($parent instanceof Syncable && $parent->shouldSync()) {
+                $parent->triggerSyncEvent();
+            }
+        });
+    }
```
stancl: If this code is the same as in `TenantPivot`, we may want to extract it to a trait

tests/ResourceSyncingPolymorphicTest.php
```diff
@@ -0,0 +1,334 @@
...
+        return [
+            'global_id',
+            'name',
+            'password',
+            'email',
+        ];
+    }
+}
+
+class ResourceUserForPolymorphic extends Model implements Syncable
```
stancl: Refer back to our conversation about `ResourceUser` and properly naming things in resource syncing tests.

abrardev99:
> @abrardev99 Add instructions for using the old approach to the PR description. Since the trait now uses the morph classes by default, it should be clearly explained how people can change this to use the old relationship structure.

Done.

stancl:
The code looks good now, I'll do a deeper review (especially of the tests) later this week but seems like everything is complete here.

tests/ResourceSyncingTest.php
```diff
@@ -284,14 +284,10 @@
...
-    $centralUser = CentralUserProvidingMixture::whereGlobalId('acmey')->first();
+    $centralUser = CentralUserProvidingMixture::whereGlobalId('acmey')->first()->only(['name', 'email', 'password', 'role']);
     expect($resourceUser->getSyncedCreationAttributes())->toBeNull();
 
-    $centralUser = $centralUser->toArray();
-    $resourceUser = $resourceUser->toArray();
-    unset($centralUser['id']);
-    unset($resourceUser['id']);
-
+    $resourceUser = $resourceUser->only(['name', 'email', 'password', 'role']);
```
stancl: What's the reason for this change?

abrardev99: Just cleaning/improvements. I used the same technique in newly added tests and thought I should refactor them too. Changes in `ResourceSyncingTest.php` are not related to PR/ or needed.

stancl: I don't think this approach is good. Below, we are asserting this:
```php
// Assert central user created as 1:1 copy of resource model except "id"
expect($centralUser)->toBe($resourceUser);
```

Which implies that the models should be ENTIRELY identical (except the id for obvious reasons).

Selecting only a few columns to compare defeats the point here — we want to check that the models are completely identical.

I think both in this test, and in the tests you added, you should be using the `unset()` approach because it makes it clear that we assert that the models are the same.

Using `only()` *might* be the same thing here, **but it's not obvious to people who read the test**. They'd need to check the migrations to see that. So it's better to make it explicit in the code.

tests/ResourceSyncingTest.php
```diff
@@ -832,10 +824,9 @@ function migrateUsersTableForTenants(): void
 // Tenant model used for resource syncing setup
 class ResourceTenant extends Tenant
 {
-    public function users()
+    public function users(): BelongsToMany
     {
-        return $this->belongsToMany(CentralUser::class, 'tenant_users', 'tenant_id', 'global_user_id', 'id', 'global_id')
-            ->using(TenantPivot::class);
```
stancl, line 838: Why was the `using()` removed?

src/Database/Concerns/ResourceSyncing.php
```diff
@@ -43,4 +46,15 @@ public function shouldSync(): bool
...
+    public function tenants(): MorphToMany
+    {
+        return $this->morphToMany($this->getResourceTenantModelName(), 'tenant_resources', 'tenant_resources', 'resource_global_id', 'tenant_id', 'global_id')
+            ->using(TenantMorphPivot::class);
+    }
+
+    public function getResourceTenantModelName(): string // todo better name
+    {
+        return config('tenancy.tenant_model', Tenant::class);
+    }
```
stancl: Why was this added again?

src/Database/Concerns/ResourceSyncing.php
```diff
@@ -43,4 +46,15 @@ public function shouldSync(): bool
     {
         return true;
     }
+
+    public function tenants(): MorphToMany
+    {
+        return $this->morphToMany($this->getResourceTenantModelName(), 'tenant_resources', 'tenant_resources', 'resource_global_id', 'tenant_id', 'global_id')
+            ->using(TenantMorphPivot::class);
+    }
```
stancl: Why does this forcibly use `MorphToMany`?

tests/ResourceSyncingUsingPolymorphicTest.php
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    public function companies(): MorphToMany
+    {
+        return $this->morphedByMany(CentralCompanyUsingPolymorphic::class, 'tenant_resources', 'tenant_resources', 'tenant_id', 'resource_global_id', 'id', 'global_id')
+            ->using(TenantPivot::class);
+    }
+}
+
+class CentralUserUsingPolymorphic extends Model implements SyncMaster
```
stancl, line 247: Can you also check that you can access the `tenants()` relationship from these models?

tests/ResourceSyncingTest.php
abrardev99: > They'd need to check the migrations to see that. So it's better to make it explicit in the code.

Makes sense. I'll update.

src/Database/Concerns/ResourceSyncing.php
abrardev99: In case the developer wants to provide a Tenant model other than the default one. A classic example is in our tests https://github.com/archtechx/tenancy/pull/997/files#diff-ff1377ee24fda8c295959b05c5156fda8ae36d131e56f20e220f153c4d19e8ccR181-R184

src/Database/Concerns/ResourceSyncing.php
abrardev99: Using the `polymorphic` approach is the default in this PR (We discussed this). So with this approach, I think it's okay to use `MorphToMany` here. And developers can override this method of course like 
```php
{
    return $this->belongsToMany(ResourceTenant::class, 'tenant_users', 'global_user_id', 'tenant_id', 'global_id')
        ->using(TenantPivot::class);
}
```

Is there something wrong with this return type?

abrardev99: I think we had a discussion somewhere where we decided that `MorphToMany` is fine. And it's `BelongsToMany` can be used too because the `parent` return type is the same.

tests/ResourceSyncingTest.php
abrardev99: I'm still looking at why it got removed. I think it's a mistake/leftover when I refactored the `polymorphic` test to a separate test suite. Tests are still passing though. So I'll add it back.

abrardev99: I believe this was a mistake. `using` has been added back.

tests/ResourceSyncingUsingPolymorphicTest.php
abrardev99: Assertions added.

tests/ResourceSyncingUsingPolymorphicTest.php
```diff
@@ -72,6 +72,8 @@
 
     $centralUser->tenants()->attach('t1');
 
+    expect($centralUser->tenants()->count())->toBe(1); // Assert `tenants` are accessible
```
stancl: In all of these assertions, probably assert that it's also the right tenant, not just the count.

tests/ResourceSyncingUsingPolymorphicTest.php
```diff
@@ -72,7 +72,10 @@
 
     $centralUser->tenants()->attach('t1');
 
-    expect($centralUser->tenants()->count())->toBe(1); // Assert `tenants` are accessible
+    // Assert `tenants` are accessible
+    expect($centralUser->tenants->first())
+        ->not()->toBeNull()
```
stancl: The `->not()->toBeNull()` assertion is redundant since on the next line you're asserting that we can access `->id`

abrardev99: Makes sense.

src/Database/Concerns/ResourceSyncing.php
stancl: The link doesn't work now - can you provide a permalink to some commit/lines of code, instead of the PR diff?

src/Database/Concerns/ResourceSyncing.php
stancl: > And it's BelongsToMany can be used too because the parent return type is the same.

What do you mean by this?

tests/ResourceSyncingUsingPolymorphicTest.php
abrardev99: I updated and it's just this now `expect($centralUser->tenants->first()->id)->toBe('t1');`

src/Database/Concerns/ResourceSyncing.php
abrardev99: > What do you mean by this?

This method has a `MorphToMany` return type. Its base class is `BelongsToMany`. So in case, someone wants to use the separate table approach (opposite of polymorphic), `BelongsToMany` can be specified in the overridden method.

src/Database/Concerns/ResourceSyncing.php
abrardev99: https://github.com/archtechx/tenancy/blob/ea3e90987b40c4025246d6e80b7487971d9c36e5/tests/ResourceSyncingUsingPolymorphicTest.php#L290-L293

tests/ResourceSyncingUsingPolymorphicTest.php
```diff
@@ -73,9 +73,7 @@
     $centralUser->tenants()->attach('t1');
 
     // Assert `tenants` are accessible
-    expect($centralUser->tenants->first())
-        ->not()->toBeNull()
-        ->id->toBe('t1');
+    expect($centralUser->tenants->first()->id)->toBe('t1');
```
stancl: One more change: maybe using `->tenants->pluck('id')` and `toBe(['t1'])` would also assert that there's only one tenant attached.

src/Database/Concerns/ResourceSyncing.php
stancl: A few things:
1. What's the actual point of this? Can't we just change the config in the tests?
2. Are tests the only use case or is there any other use case?
3. As we've discussed several times before, the term "resource tenant" doesn't really make sense.

abrardev99: > What's the actual point of this? Can't we just change the config in the tests?

We can. What if the developer wants to provide a different tenant model in a real app? 

> Are tests the only use case or is there any other use case?

I believe, there can be a real app case. I'm not sure what, but I believe there should be an option to provide a different tenant model. 

> As we've discussed several times before, the term "resource tenant" doesn't really make sense.

Yes, but there is already a method `getTenantModelName`, that's why the current name was feeling right.

stancl: > What if the developer wants to provide a different tenant model in a real app

I don't see why the developer would want to use a different model **just for this one thing**, but use the configured Tenant model for everything else. Am I missing something?

> I believe there should be an option to provide a different tenant model

In all other parts of the package we use the exact same tenant model — the configured one.

abrardev99: > In all other parts of the package we use the exact same tenant model — the configured one.

I see, that makes sense.

src/Database/Concerns/ResourceSyncing.php
stancl: I think the base class doesn't matter here in any way. When you have a case where "A extends B", your typehint says A, and you want to change it to B later on, then PHP won't let you. The child class can have additional things, it's a separate type. Using a "B" typehint would let you use both A and B.

That said, this is a trait so I see your point about MorphToMany being the default, and about people being able to override this.

Do we override the method with a BelongsToMany in any tests? We should cover this, so that we know that overriding the method works fine.

abrardev99: > I think the base class doesn't matter here in any way. When you have a case where "A extends B", your typehint says A, and you want to change it to B later on, then PHP won't let you. The child class can have additional things, it's a separate type. Using a "B" typehint would let you use both A and B.

Yes, that makes sense. So trait method type should be `BelongsToManay`, then we can use any type between two. 

> Do we override the method with a BelongsToMany in any tests? We should cover this, so that we know that overriding the method works fine.

Yes, we are overriding this method and providing the `BelongsToMany` return type in `ResourceSyncingTest` tests. 
https://github.com/archtechx/tenancy/blob/144df028a0a21e0c3098d9549604379a0a9571cb/tests/ResourceSyncingTest.php#L852-L856


It's working because our interface specifies the `BelongsToMany` type. That's why we are able to use both. It doesn't matter what type we use in the trait. It just matters if you hold the contract with the interface.

abrardev99: I've updated the trait method return type to `BelongsToMany`, which makes sense. As `BelongsToMany` is the parent type, developer can use both `BelongsToMany` or `MorphToMany`

abrardev99: Anyway, I've updated the trait method return type to `BelongsToMany` too.

stancl: Why? I said that this is a trait so there should be no issues overriding it. And if:
- the default is `MorphToMany`
- to use `BelongsToMany`, people are expected to override the method

Then it makes no sense to have a `$this->morphTomany()` method that returns `BelongsToMany`.

abrardev99: Makes sense.

tests/ResourceSyncingUsingPolymorphicTest.php
```diff
@@ -0,0 +1,349 @@
...
+use Stancl\Tenancy\Listeners\RevertToCentralContext;
+use Stancl\Tenancy\Listeners\UpdateSyncedResource;
+use Stancl\Tenancy\Tests\Etc\Tenant;
+
+beforeEach(function () {
+    config(['tenancy.bootstrappers' => [
+        DatabaseTenancyBootstrapper::class,
+    ],
+        'tenancy.tenant_model' => ResourceTenantUsingPolymorphic::class,
+    ]);
```
stancl: Can you:
1. Reformat this to be a normal array
2. Update the config key? https://github.com/archtechx/tenancy/blob/master/assets/config.php#L13

It's strange that the tests are passing even though the model is not actually set correctly.

Are you using `$tenant->users` or `$tenant->companies` anywhere? It seems that the model isn't actually used here.

We should be making assertions that we can access the right users & companies *from* the tenant.

abrardev99: > It's strange that the tests are passing even though the model is not actually set correctly.

Tests are passing because the config key is correct. In this PR, config has `tenancy.tenant_model`. The link you provide is in the master. So, I'm going to merge the master branch into this branch, and then we'll see if tests pass with same code.

abrardev99: > Are you using $tenant->users or $tenant->companies anywhere? It seems that the model isn't actually used here.

You are right. `$tenant->companies` is not being used anywhere. Tests are passing because TestCase.php set the `Tenant` model. Test fails when I try to access `$tenant->companies` because right Tenant model is not being used.

abrardev99: Added the assertions and a test for accessing the right resources from the Tenant model.

tests/ResourceSyncingUsingPolymorphicTest.php
```diff
@@ -0,0 +1,400 @@
...
+    expect($tenant1->users()->pluck('email')->toArray())->toBe(['john@localhost']);
+
+    $centralUser = $centralUser->withoutRelations()->toArray();
+    $tenantUser = $tenantUser->toArray();
+    unset($centralUser['id'], $tenantUser['id']);
+
+    // reorder global_id in $centralUser to put it at the last position so we can match the exact array
+    $centralUser['global_id'] = array_shift($centralUser);
+
+    expect($tenantUser)->toBe($centralUser);
```
stancl:
```suggestion
    // array keys use a different order here
    expect($tenantUser)->toEqualCanonicalizing($centralUser);
```

PR #1005, opened 2022-11-17
CachedTenantResolver missing cache case crash fix.

Trimod:
**Issue**: In case cache is broken or invalidated in parallel with request Resolver might crash because `has->($key)` will return `true`, but `$tenant` var will be `null`.

I do understand that this might not be the main issue, but still it`s better to find tenant without cache, then crash :)

### DIFF
--- a/src/Resolvers/Contracts/CachedTenantResolver.php
+++ b/src/Resolvers/Contracts/CachedTenantResolver.php
@@ -27,9 +27,8 @@ abstract class CachedTenantResolver implements TenantResolver
 
         $key = $this->getCacheKey(...$args);
 
-        if ($this->cache->has($key)) {
-            $tenant = $this->cache->get($key);
-
+        $tenant = $this->cache->has($key) ? $this->cache->get($key) : null;
+        if ($tenant) {
             $this->resolved($tenant, ...$args);
 
             return $tenant;
--- a/src/Resolvers/Contracts/CachedTenantResolver.php
+++ b/src/Resolvers/Contracts/CachedTenantResolver.php
@@ -27,8 +27,7 @@ abstract class CachedTenantResolver implements TenantResolver
 
         $key = $this->getCacheKey(...$args);
 
-        $tenant = $this->cache->has($key) ? $this->cache->get($key) : null;
-        if ($tenant) {
+        if ($tenant = $this->cache->get($key)) {
             $this->resolved($tenant, ...$args);
 
             return $tenant;


stancl:
I don't get how this change fixes it. It seems that the code is just moved from separate lines to a single line, but what does that actually fix?

If I understand your comment correctly, then the issue would happen if the cache is cleared in between the `has()` and `get()` calls, correct?

But using a ternary instead of multiple lines doesn't change anything about this. It's not like lines of code are atomic. It's just syntax.

stancl:
Or is the issue that `has()` would return `true` because the **key** exists but `get()` would return a falsy value?

Still, I don't see when that would happen. If you clear a cache key properly then `has()` should return false.

Trimod:
> Or is the issue that `has()` would return `true` because the **key** exists but `get()` would return a falsy value?
> 
> Still, I don't see when that would happen. If you clear a cache key properly then `has()` should return false.

Issue is that `has()` would return `true` because the key exists but `get()` would return `null`.

With my fix we check `$tenant` after all of that, and we skip using cache in case we were not able to retrieve it.

As the code is right now, we pass `null` to `$this->resolved($tenant, ...$args);` which crashes the request.

stancl:
How does the key become `null` in your case?

Trimod:
> How does the key become `null` in your case?

I have this logged in sentry as a single case, and everything worked on the next request. 
So my bet is that the cache was cleaned up/invalidated right in the middle of the request.

I will investigate that separetely. Just wanted to help avoid crashes even in such rare cases :)

stancl:
If cache were cleared then I think the key would be removed as well. Can't think of a case when `has()` would return `true` but `get()` a falsy value.

Trimod:
> If cache were cleared then I think the key would be removed as well. Can't think of a case when `has()` would return `true` but `get()` a falsy value.

Cache expires in the middle of the request.

Anyway, this fix does not brake anything, but rather covers for an additional case (super rare, i agree. Though i had it in production).

Do you have arguments against it ? Will you approve ?

stancl:
We only merge bug fixes if the bugs can be reproduced. The issue with merging a minor code change "just because it doesn't hurt" is that we might change the code in the future and not notice. So all changes should be intentional.

I'd need to see a reproducible case where the key remains but the value is null. If that exists then we can add a regression test, if it doesn't exist then it's likely that this is caused by some other issue on your end.

drfraker:
FWIW, I have been seeing this off and on in my app too. I think this PR might resolve the problem. I don't know how this bug could be reproduced in a test, but I think I might have an explanation as to what is happening.

1. Tenant is cached.
2. Tenant is checked on line 30 `if ($this->cache->has($key))`
3. The check on line 2 returns true.
4. The cache expires.
5. We attempt to retrieve the tenant from the cache `$tenant = $this->cache->get($key);`
6. Since the cache expired between step 2 and 5 `null` is returned.


Maybe in your test you could mock the cache to make it return the values above.
`$cacheMock->shouldReceive('has')->once()->andReturn(true); $cacheMock->shouldReceive('get')->once()->andReturn(null);`

stancl:
I see, thanks for the additional info Dustin. I'll check if we can make the cache calls atomic somehow, or we'll just use an approach similar to what's used here (checking that `get()` also returns a truthy value).

And good suggestion with the mocks, returning `true` for `has()` and `false` for `get()` seems indeed like the most straightforward way to reproduce this in tests.

codebyoul:
I think the $this->cache->get($key) is synchronised, when we run in multithread environnement, in the same tenant there is a chance that another thread change the cache and invalidate it

stancl:
Will release this as-is now to push out a fix, maybe we'll write the tests in a separate PR (probably per @drfraker's suggestion). @lukinovec you can make a BC task for that

stancl:
Just noticed that this went to `master` but should've gone to `3.x`. Will push it there as well and release

fulopattila122:
FYI This issue has happened to our app as well with v3.5.6. We've also only seen the reports in Bugsnag and had a rare occurrence, but it happened a couple of times. Thanks @Trimod for pushing this through :pray:

PR #1014, opened 2022-11-21
[4.x] Cache prefix mode for separating tenant caches

### LINKED ISSUE #531: Cache prefix mode for the cache bootstrapper
#### Description

Support cache prefix mode for separating tenant caches.

#### Why this should be added

To support more drivers than Redis.

abrardev99:
closes #531 

This PR adds support for cache prefix mode for separating tenant caches using a bootstrapper that can be enabled. The bootstrapper simply adds the prefix for cache keys. 

> Note: It does not work for the `file` driver. 

## Usage 

enable bootstrapper in the `tenancy.php` config. 
```php
'bootstrappers' => [
        Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper::class,
        // Stancl\Tenancy\Bootstrappers\CacheTenancyBootstrapper::class,
        Stancl\Tenancy\Bootstrappers\BatchTenancyBootstrapper::class,
        Stancl\Tenancy\Bootstrappers\PrefixCacheTenancyBootstrapper::class, // prefix cache keys
        // Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper::class, // Note: phpredis is needed
    ],
```
That's it. Your cache keys are prefixed now. 

The cache will behave like this. 
```php
cache()->put('key', 'original-value');

tenancy()->initialize($tenant1);
cache('key') // returns null;
cache()->put('key', 'value-for-tenant1');

tenancy()->initialize($tenant2);
cache('key') // returns null;
cache()->put('key', 'value-for-tenant2');

tenancy()->end();
cache('key') // returns 'original-value';

tenancy()->initialize($tenant1);
cache('key') // returns 'value-for-tenant1'

tenancy()->initialize($tenant2);
cache('key') // // returns 'value-for-tenant2'
```


assets/config.php
```diff
@@ -102,6 +102,7 @@
         Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\QueueTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\BatchTenancyBootstrapper::class,
+        // Stancl\Tenancy\Bootstrappers\PrefixCacheTenancyBootstrapper::class, // prefix cache keys
```
stancl: Remove the comment to make this consistent with the other bootstrappers

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,52 @@
...
+
+class PrefixCacheTenancyBootstrapper implements TenancyBootstrapper
+{
+    protected null|string $originalPrefix = null;
+    protected string $storeName;
+
+    public function bootstrap(Tenant $tenant): void
+    {
+        $this->originalPrefix = config('cache.prefix');
+        $this->storeName = config('cache.default');
```
stancl: Use proper DI in classes like this

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,52 @@
...
+
+        $this->setCachePrefix('tenant_id_' . $tenant->id);
```
stancl: The prefix should be customizable in the config. Take a look at the config file and see how all of the other prefixes we use are handled 👍🏻

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,52 @@
...
+        // the `$store` but doesn't update the `$app['config']`.
+        app()->forgetInstance('cache');
+
+        //This is important because the Cache Repository is using an old version of the CacheManager
+        app()->forgetInstance('cache.store');
+
+        // Forget the cache repository in the container
+        app()->forgetInstance(Repository::class);
+
+        Cache::clearResolvedInstances();
```
stancl: Do we need **all** of these 4 calls?

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->storeName = config('cache.default');
+
+        $this->setCachePrefix('tenant_id_' . $tenant->id);
+    }
+
+    public function revert(): void
+    {
+        if ($this->originalPrefix) {
+            $this->setCachePrefix($this->originalPrefix);
+            $this->originalPrefix = null;
```
stancl: Why `null` when in `bootstrap()` we use `config('cache.prefix')`? This means that initializing and ending tenancy will break people's central cache.

Also, before fixing this, reproduce this in tests first.

stancl: I see that there aren't any tests specific to this bootstrapper, just the `cache data is separated` test.

Try to write a proper suite of tests for this bootstrapper. Every other bootstrapper (or its related logic) has them, e.g. just for caching:
- https://github.com/archtechx/tenancy/blob/d6da626f7371b4aa10fab38c61fc6ea2c6065485/tests/CacheManagerTest.php
- https://github.com/archtechx/tenancy/blob/d6da626f7371b4aa10fab38c61fc6ea2c6065485/tests/GlobalCacheTest.php

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: The current test is passing without these four calls. 
But I think we should keep the `Cache::clearResolvedInstances();` one and remove the other three.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: > Why null when in bootstrap() we use config('cache.prefix')? This means that initializing and ending tenancy will break people's central cache.

I tried removing the `null` if check, looks like tests are passing without this.

abrardev99: I added test suite https://github.com/archtechx/tenancy/commit/c4f9324aa85d79e94212ddb48fea626784a0653e. 

But to be honest, I don't think we need a separate test suite. Bootstrapper is simple enough. We already tested that it separates data. Most of the tests in this suite are inspired by the CacheManagerTest.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: So after adding the new test suite, a test failed if we remove `$this->app->forgetInstance('cache.store');`. For now, I've removed the following. 
```php
$this->app->forgetInstance('cache');

$this->app->forgetInstance(Repository::class);
```

And current tests are passing.

stancl: We should know very clearly what each binding is used for and if it needs to be cleared, rather than trying to randomly clear bindings. If you miss some edge case in tests, there will be bugs, so it'd be better to be more intentional about what the code is doing here.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
stancl: Things like this:

> Why null when in bootstrap() we use config('cache.prefix')? This means that initializing and ending tenancy will break people's central cache.

Are what the dedicated test file should cover. Things specific to the **implementation** that we want to make sure work correctly, even in edge cases.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: > We should know very clearly what each binding is used for and if it needs to be cleared, rather than trying to randomly clear bindings. 

I agree. So I researched these bindings and why the Spatie package adds these. 

### Reset Facade cache `Cache::clearResolvedInstances()`
It is needed when a call to the facade has been made before bootstrapping tenancy. The facade has its own cache, separate from the container. That's why we need to clear facade cache. 

### Forget cache store manager `$this->app->forgetInstance('cache.store')`
It is needed because the Cache Repository uses an old version of the CacheManager so we need to forget it. Here is the PR someone sent to Spatie package and described the issue and reason for adding this https://github.com/spatie/laravel-multitenancy/pull/374

### Forget cache instance `$this->app->forgetInstance('cache')`
Simply calling `forgetDriver()` isn't enough to update cache prefixes everywhere. The `CacheManager` also keeps a copy of the old config in `cache`.  These should be cleared and forgotten after switching cache configs. Here is the related PR https://github.com/spatie/laravel-multitenancy/pull/355

### Forget cache repository in the container `$this->app->forgetInstance(Repository::class)`
I'm not really sure if we needed this. Spatie package added this recently. 
image
Don't you think forgetting the `cache.store` should also forget the `Repository`? 


So currently, I'm not sure if we should keep all these calls. As I added a separate test suite, I'm thinking of edge cases for tests. And maybe there, I can actually feel the need for these calls. 
But from the Spatie package PRs, it looks like people faced issues, so they contributed. We might also want to keep all these 4 calls.

stancl: Thanks a lot for the detailed write-up.

I guess we can keep all of them then, just add a concise comment above each of those lines explaining in short what it's used for/why it's needed.

And regarding `Repository`, I guess we can just keep it there and mention in the comment that it's added just to cover edge-cases.

I guess you don't need to cover **low-level things** in tests if they work well in the Spatie package.

But we likely still want some specific tests for this bootstrapper that make sure it doesn't affect other caches, that central cache still works well, that everything works well after tenancy is ended (and also ended & reinitialized) etc. Just try to cover *usage edge-cases*, and stuff that's specific to the implementation that you think should be tested. Look at the other bootstrapper tests for inspiration.

abrardev99: Ok, I added all four calls and added the comments on each call. 

As for the tests, I added a few tests (which I can think of). Most of them are inspired by the existing bootstrapper test suite.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: Added https://github.com/archtechx/tenancy/pull/1014/files#diff-77383c3312698f402a6bd8bf0d47ef7622c5f0fe67954a5f055539650025a72d

assets/config.php
```diff
@@ -187,6 +188,7 @@
      */
     'cache' => [
         'tag_base' => 'tenant', // This tag_base, followed by the tenant_id, will form a tag that will be applied on each cache call.
+        'prefix_base' => 'tenant_' // this prefix_base, followed by the tenant_id, will form a cache prefix that will be applied on each cache call
```
stancl:
```suggestion
        'prefix_base' => 'tenant_' // This prefix_base, followed by the tenant_id, will form a cache prefix that will be applied on each cache call.
```

Consistency with the above line.

Also, maybe `that will be used for every cache key` might make more sense? I used "applied on each *call*" for the `tag_base` because tags are applied on calls. But prefixes are used for keys.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
stancl: @abrardev99 By proper DI I didn't mean inject `Application` and then use `$this->app['config']`: that has no benefits of DI — it doesn't list the dependencies, and it doesn't provide any meaningful IDE support or information that phpstan could use.

I mean actually injecting *the dependencies that we use*, not just `Application`.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->originalPrefix = $this->app['config']['cache.prefix'];
+        $this->storeName = $this->app['config']['cache.default'];
+
+        $this->setCachePrefix($this->app['config']['tenancy.cache.prefix_base'] . $tenant->id);
+    }
+
+    public function revert(): void
+    {
+        $this->setCachePrefix($this->originalPrefix);
+        $this->originalPrefix = null;
```
stancl: What's the reason for clearing the `originalPrefix`? I thought I asked this in a review already but seems like I didn't.

stancl: I can see how that makes sense here but would like to read your reasoning first in case you have other things in mind as well. These things can be a bit tricky in bootstrappers so they need to be done carefully.

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,107 @@
...
+    $tenant2 = Tenant::create();
+    tenancy()->initialize($tenant2);
+
+    pest()->assertNotSame('bar', cache()->get('foo'));
+
+    cache()->put('foo', 'xyz', 1);
+    expect(cache()->get('foo'))->toBe('xyz');
+});
+
+test('central cache is not broke', function () {
```
stancl: Fix grammar on all of the test names here

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+    $tenant1 = Tenant::create();
+    $tenant2 = Tenant::create();
+
+    $tenantOnePrefix = 'tenant_' . $tenant1->id . ':';
+
+    tenancy()->initialize($tenant1);
+    expect($tenantOnePrefix)
+        ->toBe(app('cache')->getPrefix())
+        ->toBe(app('cache.store')->getPrefix());
```
stancl: This test should also try to write something to cache and then assert that the key is prefixed. Is there a way to fetch all keys from cache?

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: > What's the reason for clearing the originalPrefix?

There is no specific reason. Tests are passing without this. We always set the prefix in bootstrap so clearing here doesn't need. But it also makes sense to clean it after it is being used.

abrardev99: TBH, I added it because I faced failures in tests where bootstrapper used the previous tenant prefix. But the actual issue was bootstrapper was not registering as a singleton. So I register it the same in TestCase (same as RedisTenanc...). And I left this line as an extra precautionary step.

tests/PrefixCacheBootstrapperTest.php
abrardev99: > This test should also try to write something to cache and then assert that the key is prefixed.

Don't you think other tests are doing this where we are asserting cache data is separated? 

> Is there a way to fetch all keys from cache?

Not easy or Laravel provided but yes we can.

assets/config.php
abrardev99: Makes sense. I updated https://github.com/archtechx/tenancy/pull/1014/commits/41a6f5b074d92a30b76c2cd233a100be12d5228e.

tests/PrefixCacheBootstrapperTest.php
stancl: > where we are asserting cache data is separated

Those are high-level tests — they assert that the data is separated, but we want to assert the low-level details as well (that the prefixes are correctly set).

abrardev99: Looks like there is no easy way to get the keys. We need to get the store to get the keys. In our, I'm fetching from Redis `Redis::connection('cache')->keys('*')`. But these keys are not accurate. These also include the redis prefix.

stancl: I see. Then I guess we can just try building the prefix manually and then try accessing the tenant's data from the central context and assert that it's the correct value

abrardev99: Added https://github.com/archtechx/tenancy/pull/1014/commits/b920a2905ea6c0d3bab19c393e87fca0ab930323

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,119 @@
...
+        'cache.default' => 'redis',
+    ]);
+
+    Event::listen(TenancyInitialized::class, BootstrapTenancy::class);
+    Event::listen(TenancyEnded::class, RevertToCentralContext::class);
+});
+
+test('cache prefix is separate for each tenant', function () {
+    $originalPrefix = config('cache.prefix') . ':';
+    cache()->set('key', 'original-value');
```
stancl: Is this used for anything?

stancl:
How does this feature behave when someone configures a prefix for the central cache?

abrardev99:
> How does this feature behave when someone configures a prefix for the central cache?

it will remain intact. When tenancy revert, We reset back the original prefix.

tests/PrefixCacheBootstrapperTest.php
abrardev99: No. Removed

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: Makes sense. I DI the Config class. And I think that's all the dependencies we are depending on.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,59 @@
...
+
+    protected function setCachePrefix(null|string $prefix): void
+    {
+        $this->config->set('cache.prefix', $prefix);
+
+        $this->app['cache']->forgetDriver($this->storeName);
+
+        // The CacheManager will have the $app['config'] array cached with old prefixes on the 'cache' instance
+        // This call will forget the 'cache' instance
+        $this->app->forgetInstance('cache');
```
stancl: If `'cache'` is forgotten completely, what's the point of line 43?

Is the driver stored as a singleton instance in the container or something like that?

stancl:
> > How does this feature behave when someone configures a prefix for the central cache?
> 
> it will remain intact. When tenancy revert, We reset back the original prefix.

I think the prefixes used for the tenant keys should also include the original prefix. It should be `$originalPrefix . $tenantPrefix . $key` (pseudocode).

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
abrardev99: > Is the driver stored as a singleton instance in the container or something like that?

Yes. https://github.com/laravel/framework/blob/9.x/src/Illuminate/Cache/CacheServiceProvider.php#L22

abrardev99:
> I think the prefixes used for the tenant keys should also include the original prefix. It should be $originalPrefix . $tenantPrefix . $key (pseudocode).

Done.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
stancl: I don't think that's related to `forgetDriver()` — we remove `cache.store` on line 50, so what exactly does `forgetDriver()` do here?

stancl: Seems like the method itself only calls `unset()` **on the `CacheManager`**, so it doesn't interact with the container. Therefore this is false:

> Is the driver stored as a singleton instance in the container or something like that?

And if we forget the instance of `CacheManager` completely on line 47, I don't think we need to bother cleaning up the state of the class?

Or is this due to existing classes that may already have the `CacheManager` injected?

abrardev99: > so it doesn't interact with the container. Therefore this is false:

Ah I see. 

> And if we forget the instance of CacheManager completely on line 47, I don't think we need to bother cleaning up the state of the class?

Logically it makes sense. I tried removing it tests are still passing.

stancl: Presumably because we aren't testing a case where `CacheManager` is injected into a service class, and then tenancy ended and reinitialized for another tenant.

abrardev99: What are you suggesting here? Should we write a test?

stancl: Probably yeah. And check if that verifies that the `forgetDriver()` call here is used for that.

abrardev99: @stancl I wrote this test https://github.com/archtechx/tenancy/pull/1014/commits/222686ec1d6772048ded1b6de8e73fac96347bbf, and it looks like `CacheManager` injected into a service class working fine without `$this->app['cache']->forgetDriver($this->storeName);`

stancl: - Try to use a second tenant as well, as I mentioned here
- The service should be a singleton. You're creating a fresh instance every time, which misses the point of what we want to test. We want to have a single instance of the service and then see if its dependencies **that were already injected in the past** are updated. Here you're re-creating the instance and re-injecting a fresh `CacheManager`

abrardev99: > The service should be a singleton. 

That is a very good point. The test fails when I register it as a singleton. So I enabled this line `$this->app['cache']->forgetDriver($this->storeName);` and the test is passing.

abrardev99: Looks like it's failing for `tenant2`. Looking into it.

tests/CacheManagerTest.php
```diff
@@ -2,17 +2,20 @@
...
 use Stancl\Tenancy\Listeners\BootstrapTenancy;
-use Stancl\Tenancy\Tests\Etc\Tenant;
+use Stancl\Tenancy\Bootstrappers\CacheTenancyBootstrapper;
 
 beforeEach(function () {
     config(['tenancy.bootstrappers' => [
         CacheTenancyBootstrapper::class,
     ]]);
 
+    CacheManager::$addTags = true;
```
stancl: Can't this default to `true` and we'd just disable it in the tests specific to the cache prefix? Would require changing much fewer files I think. Remember that it's only a temporary thing we're adding before we send the Laravel PR, then we'll get rid of this completely.

lukinovec: Yeah, that makes a lot more sense. Updated that

lukinovec:
[Opened an issue with the rough documentation of the bootstrapper added by this PR ](https://github.com/stancl/tenancy-docs/issues/230)@stancl (it's the description of this PR, but modified)

stancl:
Could we also add a test where we use a different cache repo in a service? Probably one that's always central regardless of tenancy. If we want to use DI then I think the right way is to use `CacheManager` in the constructor *parameters* but then use `$this->cache = $cacheManager->driver('foo')` (or some call like that) to inject the actual *repository*?

We could also inject CacheManager, but assuming we never destroy it and only make changes *on the single instance that's always kept in the container*, that's probably not needed since it has predictable behavior. Injecting a specific repository is a bit more unpredictable so it'd be good to have it covered in tests. I wonder how the current implementation will handle that.

lukinovec:
> Could we also add a test where we use a different cache repo in a service?

Good idea. We should definitely cover that.

Sorry for responding so late @stancl, it took me a while to understand the comment. Just to be clear – what did you mean by "injecting" in the second paragraph?  From the context, I figure "inject it and save it in the `$cache` property", is that right? Asking because, in the first paragraph, you wrote about using CacheManager as a constructor parameter, which means injecting CacheManager. But after that, you wrote "We could also inject CacheManager...".

lukinovec:
> Probably one that's always central regardless of tenancy

With the current code, only the default cache driver gets prefixed. The default driver is the one CacheManager uses for cache operations – if we want a different driver to perform the cache operations, we'd have to make it default. This means that any driver we use other than the default will be central (= won't get any prefix).

I pushed [a test](https://github.com/archtechx/tenancy/pull/1014/commits/a59d5a1069e9cc9c5b78126dd0a9719263aa0a18) that tests that behavior

tests/Etc/CacheManagerService.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Tests\Etc;
+
+use Illuminate\Cache\CacheManager;
+use Illuminate\Cache\Repository;
+
+class CacheManagerService
+{
+    public Repository|null $cache = null;
```
stancl: I think the `null` can be removed here
```suggestion
    public Repository $cache;
```

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,217 @@
...
+        $redis = $this->app['redis'];
+
+        $connection = $config['connection'] ?? 'default';
+
+        $store = new RedisStore($redis, $this->getPrefix($config), $connection);
+
+        return $this->repository(
+            $store->setLockConnection($config['lock_connection'] ?? $connection)
+        );
+    });
```
stancl: Is this part needed? Setting the config should be all that's needed.

stancl:
> This means that any driver we use other than the default will be central (= won't get any prefix).

Could the list of tenant-aware caches be configurable?

lukinovec:
> Could the list of tenant-aware caches be configurable?

Yeah. I think I'd add a static property (`$tenantCacheDrivers` array) to `PrefixCacheTenancyBootstrapper`, and prefix the cache only if the current driver is present in the property.

But I also thought that since the users probably mostly want to prefix all drivers by default, it could be better to specify `$nonTenantCacheDrivers` (drivers specified in the property wouldn't get prefixed – so the reverse version of what I wrote in the first paragraph).

I'd go for `$nonTenantCacheDrivers` because I think it's more likely people would want to have all drivers prefixed, and make exceptions for a few specified drivers in the property. Let me know which approach you think is more reasonable

tests/PrefixCacheBootstrapperTest.php
lukinovec: This is needed because CacheManager doesn't have a method for creating a `redis2` repository (see [this](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Cache/CacheManager.php#L102-L105) – CacheManager would have to have a `createRedis2Driver`).

Also if you mean setting the `cache.default` config, the cache would be prefixed – unless we make the tenant-aware caches configurable. I'll make them configurable now

stancl:
Can you resolve conflicts so that CI can run?

> But I also thought that since the users probably mostly want to prefix all drivers by default

Don't think so because most people don't use multiple drivers. They use just one cache. And if they create a new dedicated one, the odds that it's a central one are decently high.

Specifying the drivers in a static property seems like the right approach.

tests/PrefixCacheBootstrapperTest.php
stancl: The cache config looks like this:

CleanShot 2023-01-05 at 2 53 32@2x

Can't we create a `redis2` store that uses a `redis` driver?

lukinovec: I see, that should work, I'll try removing the `extend` code

lukinovec: Yeah you were right, it works without the code

lukinovec:
I updated the tests @stancl. I'm still not sure about [this one](https://github.com/archtechx/tenancy/pull/1014/files#diff-77383c3312698f402a6bd8bf0d47ef7622c5f0fe67954a5f055539650025a72dR181) (the name is misleading – it tests if the different, non-tenant repo obtained using the injected CacheManager stays unprefixed).

> Could we also add a test where we use a different cache repo in a service? Probably one that's always central regardless of tenancy. 

We want to test if the prefixing works correctly when using a different repository obtained through the injected CacheManager, right? But if the repo is always central, it won't get prefixed at all – shouldn't we test if a tenant cache repo _gets_ prefixed instead?

stancl:
No, I think testing that it doesn't get prefixed makes sense here because the point of the test is to verify that people *can use a central cache inside their services*. If we tested that it gets prefixed we'd be testing the same thing we're already testing in a different way, but we wouldn't cover the use case where people want to explicitly use the central cache at all.

lukinovec:
I see, thanks for the clarification

lukinovec:
I updated the test and the service, what do you think now @stancl?

https://github.com/archtechx/tenancy/pull/1014/files/52db2a0c06ccb11ccf673d0a3e518cdea250895a..8deeef25beb2bd549aecf264484dccef29020232

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
lukinovec: I think we can resolve this because the code is different now https://github.com/archtechx/tenancy/pull/1014/files#diff-41b4083e1eb3004d0f552f7805c28a2c2e479078ed592be5d955a90788f486d5R41

stancl:
> I updated the test and the service, what do you think now @stancl?
> 
> https://github.com/archtechx/tenancy/pull/1014/files/52db2a0c06ccb11ccf673d0a3e518cdea250895a..8deeef25beb2bd549aecf264484dccef29020232

Is the `$expectPrefixToBe($originalPrefix);` assertion also used after ending tenancy? To check that after reverting, the cache is correct.

lukinovec:
> Is the `$expectPrefixToBe($originalPrefix);` assertion also used after ending tenancy? To check that after reverting, the cache is correct.

It wasn't used @stancl. I updated the test so that tenancy is ended and it's checked if the prefix gets reverted correctly [here](https://github.com/archtechx/tenancy/pull/1014/commits/c3e3a33ed0a02fdb182007dbfa9efb76078adda2)

stancl:
Linking this here for easy access: [`cache-methods`](https://github.com/lukinovec/framework/tree/cache-methods) ([compare](https://github.com/laravel/framework/compare/10.x...lukinovec:framework:cache-methods))

Merged here: https://github.com/laravel/framework/pull/46203

stancl:
Reverted 5f3079d2ff62979e79ab2cfd94f45f917f66b317 since the matrix is defined *inside* the `tests` job. So the phpstan job will be failing here but we can ignore that.

lukinovec:
@stancl, there's a resource syncing test that began failing after a recent Laravel release. Locally, the tests run fine when using specifically v10.1.1 (there's just the one cache prefixing test that I expect to fail – should have marked that one as skip), but using ^10.0 installs v10.1.4, and after that, the resource syncing test starts failing. So somewhere after Laravel v10.1.1, there was a change that broke our test.

stancl:
So it worked in v10.1.1 but is failing in v10.1.4?

If so, we can check the diff here https://github.com/laravel/framework/compare/v10.1.1...v10.1.4

lukinovec:
> So it worked in v10.1.1 but is failing in v10.1.4?

Yeah, [this change](https://github.com/laravel/framework/compare/v10.1.1...v10.1.4#diff-b656de2aea3c48b173198c2ab34fa1d255f99291225778406f777d191f12a22cR50-R57) – "Don't fake facades if they've already been faked" caused that (https://github.com/laravel/framework/pull/46188). When we call the second `Event::fake()` to clear the faked events, they don't get cleared now

tests/BootstrapperTest.php
```diff
@@ -82,11 +84,11 @@
     expect(DB::table('users')->first()->name)->toBe('Foo');
 });
 
-test('cache data is separated', function () {
+test('cache data is separated', function (string $bootstrapper) {
+    CacheManager::$addTags = true;
```
stancl: Do we reset this static property anywhere? Tests should do that, see what we do with other static properties

src/CacheManager.php
```diff
@@ -10,6 +10,8 @@
 
 class CacheManager extends BaseCacheManager
 {
+    public static bool $addTags = true;
```
stancl: Wouldn't the fact that:
1. The CacheManager is now always replaced in the TenancyServiceProvider
2. This defaults to `true`

Lead to this feature being always enabled by default? What is the purpose of `CacheTenancyBootstrapper` now and how does it being enabled/disabled affect the actual behavior?

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public function getStorePrefix(string $store, Tenant $tenant): string
+    {
+        if (isset(static::$prefixGenerators[$store])) {
+            return static::$prefixGenerators[$store]($tenant);
+        }
+
+        // Use default generator if the store doesn't have a custom generator
+        return static::$prefixGenerators['default']($tenant);
```
stancl: Instead of using a custom key called `default` and having to set it in `bootstrap()`, wouldn't it be better to just have a method we call here? Or inline the "default" logic here instead of moving it to a `default` prefix generator?

Or is there a point to having a `default` prefix generator?

lukinovec: The only reason for having the `default` key is that the user can change the default prefix generator. But I think it's not very likely for someone to do that – the user can use `generatePrefixUsing` for the exact store(s) they want to prefix using their own methods.

I'd inline the logic like you're suggesting since it's simple (updated the code)

src/CacheManager.php
lukinovec: > What is the purpose of CacheTenancyBootstrapper now and how does it being enabled/disabled affect the actual behavior?

Since the CacheManager is replaced in the SP, the bootstrapper seems redundant. Using it together with PrefixCacheTenancyBootstrapper doesn't seem to affect the behavior in any way.

The feature (cache prefixing) is not enabled by default – because the `$addTags` property is `true`, the default behavior of the cache manager is basically the same as it was before, but now, we replace the CacheManager instance in the SP instead of the bootstrapper.

Caching should work the same way as it worked before (except the need for using CacheTenancyBootstrapper) – the cache gets scoped by the tags (only when tenancy is initialized), and to change that behavior so that prefixing is used instead of tagging, we'd have to 1) change `$addTags` to `false` and 2) use PrefixCacheTenancyBootstrapper.

tests/BootstrapperTest.php
lukinovec: Like this? https://github.com/archtechx/tenancy/pull/1014/commits/66669934f91556802d89ef27c0f667ad0308cda7

src/CacheManager.php
lukinovec: I just deleted CacheTenancyBootstrapper, and everything seems to be working fine

tests/BootstrapperTest.php
stancl: Yeah, just to the default values of the class. Should be done in both `beforeEach()` (so that the state is cleaned up before running the test) and in `afterEach()` (so that the state is cleaned up before running _other_ tests).

stancl: And make sure to do it in every test file where the static properties get changed in any way.

assets/config.php
```diff
@@ -98,10 +98,10 @@
      */
     'bootstrappers' => [
         Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper::class,
-        Stancl\Tenancy\Bootstrappers\CacheTenancyBootstrapper::class,
```
stancl, line 101: Did we get rid of the tags-based bootstrapper completely? If so, why?

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,86 @@
...
+        Cache::clearResolvedInstances();
+    }
+
+    public function getStorePrefix(string $store, Tenant $tenant): string
+    {
+        if (isset(static::$prefixGenerators[$store])) {
+            return static::$prefixGenerators[$store]($tenant);
+        }
+
+        return $this->originalPrefixes[$store] . config('tenancy.cache.prefix_base') . $tenant->getTenantKey();
```
stancl: We can use `$this->config` instead of `config()`

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
stancl: Yeah, just inline it

src/CacheManager.php
stancl: As I wrote in a review just now, we don't want to remove it. We need backwards compatibility. And we still want to offer both approaches to separating tenant caches.

assets/config.php
lukinovec: There isn't any use for it anymore. It just swapped the CacheManager instance, but since we're doing that in TSP, the bootstrapper doesn't really do anything (explained here: https://github.com/archtechx/tenancy/pull/1014/files#r1154378154)

src/CacheManager.php
lukinovec: > we still want to offer both approaches to separating tenant caches

There's probably a misunderstanding – as I wrote, both approaches are still supported, it's just that CacheTenancyBootstrapper is now useless because its point was replacing the cache manager, and that's done in the SP now ("Caching should work the same way as it worked before (except the need for using CacheTenancyBootstrapper) – the cache gets scoped by the tags (only when tenancy is initialized), and to change that behavior so that prefixing is used instead of tagging, we'd have to 1) change $addTags to false and 2) use PrefixCacheTenancyBootstrapper.").

> we don't want to remove it. We need backwards compatibility

We're replacing the cache manager in the SP, so I don't think we can make this completely backwards compatible (removed cache bootstrapper can exist, but its only purpose would be just that – to be there so that users won't be getting an exception).

So backwards compatibility seems like the only reason to keep the bootstrapper. But couldn't we mention that the bootstrapper got removed and that it should be removed from the `tenancy.bootstrappers` config, since now, Tenancy replaces the cache manager in the SP?

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
lukinovec: I think we can resolve this since it's inline now

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
lukinovec: Right, changed that

tests/BootstrapperTest.php
lukinovec: Seems like a lot of test files need this treatment – I'd probably do that in a separate PR. What do you think?

stancl: What do you mean by that?

src/CacheManager.php
stancl: I see, so controlling whether tags are used is now done purely using the static property?

If so, maybe the bootstrapper could be used for that? It'd feel kind of in line with how other types of "tenancy logic" are enabled by the user. Also, enabling cache tags by default **without the user seeing it anywhere** (as they do with bootstrappers — they can see which ones are not commented) doesn't seem ideal.

Also, what again was the reason for moving the logic to the SP?

> So backwards compatibility seems like the only reason to keep the bootstrapper

When I refer to backwards compatibility I only mean the behavior. Syntax can change in any way 👍🏻

lukinovec: > what again was the reason for moving the logic to the SP?

If the logic was in a bootstrapper, changing the cache manager instance scopes the cache to the current tenant completely.

E.g., do `cache(['key' => 'central'])` in the central context, switch to the tenant context, do `cache('key')`. If the prefix bootstrapper is used and it changes the cache manager instead of doing that in the SP, we'd expect the unprefixed stores to return 'central'. But because we're swapping the cache manager entirely, we'd get `null`.

I think for prefixing, we can't swap the cache manager instances in the bootstrappers. We have to do it only once at some point (TSP) so that the cache only gets prefixed (`cache('key')` => `cache('tenant.key')`), not _fully_ scoped if that makes sense. And if we're using the prefix bootstrapper, but we're using a cache store that we don't want to prefix, we expect the cache calls to work as if they were central (unprefixed, not scoped), so we can't really bind a new CacheManager instance in the bootstrapper.

tests/BootstrapperTest.php
lukinovec: > make sure to do it in every test file where the static properties get changed in any way

There are more test files (unrelated to this PR) in which the static properties get changed, but we don't reset them properly. I thought that since these files aren't related to this PR's changes, I'd add the proper static property resetting in another PR.

src/CacheManager.php
lukinovec: > I see, so controlling whether tags are used is now done purely using the static property?
>
> If so, maybe the bootstrapper could be used for that?

Added a bootstrapper for enabling the tags here https://github.com/archtechx/tenancy/pull/1014/commits/7ebfc375d6ef0f43311544b26750020b9f905a6f

tests/BootstrapperTest.php
```diff
@@ -85,7 +85,7 @@
 test('cache data is separated', function () {
     config([
         'tenancy.bootstrappers' => [
-            CacheTenancyBootstrapper::class,
+            CacheTaggingBootstrapper::class,
```
stancl: In an earlier version of this PR we used the prefix bootstrapper here, now it's changed "back" to the tagging bootstrapper. Maybe it'd be reasonable to test this with both bootstrappers (perhaps using a data provider with the bootstrapper class?)

stancl: Same in `GlobalCacheTest` maybe

tests/BootstrapperTest.php
stancl: Can you give some examples? We generally don't need to reset everything, it's usually just things we change in tests. You can see that we reset static properties in tests where we change them.

assets/config.php
stancl: Marking as resolved since we continued the convo in the other thread & Discord.

stancl:
Not sure about the `CacheTaggingBootstrapper`, `CacheTagsBootstrapper` might be better but neither sounds perfect. Maybe go with the latter for now and we can change it later.

tests/BootstrapperTest.php
lukinovec: E.g. in BootstrapperTest

https://github.com/archtechx/tenancy/blob/master/tests/BootstrapperTest.php#L384
https://github.com/archtechx/tenancy/blob/master/tests/BootstrapperTest.php#L385
https://github.com/archtechx/tenancy/blob/master/tests/BootstrapperTest.php#L415

tests/BootstrapperTest.php
lukinovec: > Maybe it'd be reasonable to test this with both bootstrappers

Agree

lukinovec: Made the changes here: https://github.com/archtechx/tenancy/compare/651302943f2b7bc626b76463166688423a640318..3481e24c7d36dbdcc5e010a995f367ba38a86c04

tests/BootstrapperTest.php
```diff
@@ -28,17 +28,20 @@
...
 use Stancl\Tenancy\Middleware\InitializeTenancyBySubdomain;
 use Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\BroadcastTenancyBootstrapper;
 use Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper;
+use Stancl\Tenancy\Bootstrappers\PrefixCacheTenancyBootstrapper;
+use Stancl\Tenancy\CacheManager;
 
 beforeEach(function () {
     $this->mockConsoleOutput = false;
+    CacheManager::$addTags = false;
```
stancl: Probably not needed anymore?

stancl: Actually might be a good way of cleaning things up. Could also be done in `afterEach()` and then we could remove it from the `cache data is separated` test.

tests/BootstrapperTest.php
```diff
@@ -82,12 +85,16 @@
     expect(DB::table('users')->first()->name)->toBe('Foo');
 });
 
-test('cache data is separated', function () {
+test('cache data is separated', function (string $bootstrapper) {
+    $cacheDriver = 'redis';
+
+    if ($bootstrapper === PrefixCacheTenancyBootstrapper::class) {
+        PrefixCacheTenancyBootstrapper::$tenantCacheStores = [$cacheDriver];
```
stancl: Could we move this to `beforeEach()` and get rid of the `if` completely? And then we could get rid of the `$cacheDriver` variable as well.

We'd undo the change to `tenantCacheStores` in `afterEach()`.

Not 100% sure about this change, but it'd clean up this test so it would actually just accept the `$bootstrapper` and work with it without having any logic coupled to the values that are actually passed.

tests/BootstrapperTest.php
```diff
@@ -121,7 +128,13 @@
...
     // Asset central is still the same
     expect(Cache::get('foo'))->toBe('central');
-});
+
+    // Reset the static property
+    CacheManager::$addTags = false;
+})->with([
+    'tagging' => CacheTagsBootstrapper::class,
+    'prefixing' => PrefixCacheTenancyBootstrapper::class,
+])->group('bootstrapper');
```
stancl: You can remove this I think. Or if it's used for anything rename it to something better since it's related to cache, not "bootstrapper" (that's the whole file).

tests/BootstrapperTest.php
```diff
@@ -121,7 +128,13 @@
 
     // Asset central is still the same
     expect(Cache::get('foo'))->toBe('central');
-});
+
+    // Reset the static property
+    CacheManager::$addTags = false;
+})->with([
+    'tagging' => CacheTagsBootstrapper::class,
+    'prefixing' => PrefixCacheTenancyBootstrapper::class,
```
stancl: Are the keys used in any way? If not, we can simplify this into a normal array.

stancl: Same in GlobalCacheTest

tests/BootstrapperTest.php
lukinovec: Yeah, forgot to delete this

tests/BootstrapperTest.php
lukinovec: They aren't, so yeah, we can just use a normal array

tests/BootstrapperTest.php
lukinovec: Yeah, we can do that. The only thing that'd be worse would be that there won't be the `$cacheDriver` variable, so in the test, we'd have to assume 'redis' is in the tenant cache stores property

stancl: Are we using the cache driver variable anywhere in the actual test logic? As in after configuring these classes/the config,

lukinovec: Yeah, after this line, we used the variable to set the config ([`'cache.default'`](https://github.com/archtechx/tenancy/pull/1014/files/3481e24c7d36dbdcc5e010a995f367ba38a86c04#diff-b22931f4c76926f01513d299f2f87fb1528e0d3a92e539ef66cb4675b4342f4bR97))

stancl: Well that's still in what I meant by the config. My point is that if we move the `PrefixCacheTenancyBootstrapper::$tenantCacheStores` logic elsewhere, we don't need the variable anymore probably.

> so in the test, we'd have to assume 'redis' is in the tenant cache stores property

We'd probably set this in the `beforeEach()`. Tests can make assumptions about the environment like that.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public function bootstrap(Tenant $tenant): void
+    {
+        $this->originalPrefix = $this->config->get('cache.prefix');
+
+        // Use default prefix generator if the prefix generator isn't set
+        static::$prefixGenerator ??= function (Tenant $tenant) {
+            return $this->originalPrefix . $this->config->get('tenancy.cache.prefix_base') . $tenant->getTenantKey();
+        };
```
stancl: Was that in this PR or elsewhere that I wrote that it might be cleaner to keep the variable always null if it's not user-defined, and using the default logic when that property is null? Can't find it now but I think I mentioned this in some context.

stancl: If we made that change, we'd also remove any `static::$prefixGenerator = null` calls to avoid overriding the user-defined prefix generator. In fact, it seems like the code already overrides it. Initialize -> revert -> initialize and it will be gone (due to line 53).

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    Event::listen(TenancyInitialized::class, BootstrapTenancy::class);
+    Event::listen(TenancyEnded::class, RevertToCentralContext::class);
+});
+
+afterEach(function () {
+    PrefixCacheTenancyBootstrapper::$tenantCacheStores = [];
+    PrefixCacheTenancyBootstrapper::$prefixGenerator = null;
+
+    TenancyCacheManager::$addTags = false;
```
stancl: Is this set to `true` anywhere in this test? If not, we can remove this line.

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    tenancy()->initialize(Tenant::create(['id' => 'second']));
+
+    expect(app('cache')::class)->toBe(TenancyCacheManager::class);
+    expect(app(CacheManager::class)::class)->toBe(TenancyCacheManager::class);
+
+    tenancy()->end();
+
+    expect(app('cache')::class)->toBe(TenancyCacheManager::class);
+    expect(app(CacheManager::class)::class)->toBe(TenancyCacheManager::class);
```
stancl: Is the point of these repeated assertions ensuring that the cache manager is always overridden, no matter what happens?

And on a related note: is there an easy way for people to disable the cache manager override (in case they want to use a custom CacheManager)? Should be a static property/config key probably. Maybe putting it inside https://github.com/archtechx/tenancy/blob/6a26f712e797aa9ed79c54be281c4fb1069328af/assets/config.php#L191 would work well.

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
stancl: Also: this in the `PrefixCacheBootstrapperTest` would have to be updated:
```php
$prefixGenerator = PrefixCacheTenancyBootstrapper::$prefixGenerator;

cache()->set('key', 'tenantone-value');

$expectPrefixToBe($prefixGenerator($tenant1));
```

Probably add a public method to the bootstrapper for generating prefixes 👍🏻

stancl: Or you could manually define the generator in the test, but adding a public method is simpler and it can be a helpful thing to be available for people to use. Can assist with debugging, for example.

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+test('cache is persisted when reidentification is used', function () {
+    $tenant1 = Tenant::create();
+    $tenant2 = Tenant::create();
+    tenancy()->initialize($tenant1);
+
+    cache(['foo' => 'bar'], 10);
+    expect(cache('foo'))->toBe('bar');
+
+    tenancy()->initialize($tenant2);
+    tenancy()->end();
```
stancl: Maybe in between these two lines add an assertion that `cache('foo')` is **not** `bar`. It's probably tested in other tests already, but it can add a bit more assurance in this test so that we know that the `cache('foo')` didn't persist the value due to some bug.

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+    $tenant1 = Tenant::create();
+    tenancy()->initialize($tenant1);
+
+    cache()->put('foo', 'bar', 1);
+    expect(cache()->get('foo'))->toBe('bar');
+
+    $tenant2 = Tenant::create();
+    tenancy()->initialize($tenant2);
+
+    pest()->assertNotSame('bar', cache()->get('foo'));
```
stancl: Instead of `pest()->assertNotSame()` you can use "expect not to be"

stancl: Either `notToBe()` or `->not()->toBe()`, not sure if the former method exists

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+    expect(cache()->get('foo'))->toBe('xyz');
+});
+
+test('central cache is persisted', function () {
+    cache()->put('key', 'central');
+
+    $tenant1 = Tenant::create();
+    tenancy()->initialize($tenant1);
+
+    cache()->put('key', 'tenant');
```
stancl, line 121: Above this also assert that `cache('key')` is null

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    tenancy()->initialize($tenant1);
+    expect(cache('foo'))->toBe('bar');
+});
+
+test('prefixing separates the cache', function () {
+    $tenant1 = Tenant::create();
+    tenancy()->initialize($tenant1);
+
+    cache()->put('foo', 'bar', 1);
```
stancl: What's the `1` for?

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+    expect(cache()->get('foo'))->toBe('bar');
+
+    $tenant2 = Tenant::create();
+    tenancy()->initialize($tenant2);
+
+    pest()->assertNotSame('bar', cache()->get('foo'));
+
+    cache()->put('foo', 'xyz', 1);
+    expect(cache()->get('foo'))->toBe('xyz');
```
stancl, line 108: After this also go back to tenant 1 and check that `foo` is `bar`

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    expect($originalPrefix . $prefixBase . $tenant1->getTenantKey() . ':')
+        ->toBe(app('cache')->getPrefix())
+        ->toBe(app('cache.store')->getPrefix());
+});
+
+test('cache is prefixed correctly when using a repository injected in a singleton', function () {
+    $this->app->singleton(CacheService::class);
+
+    $this->app->make(CacheService::class)->handle();
```
stancl, line 158: Above this line maybe add a starting assertion that `cache('key')` is null

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+    expect(cache()->store($cacheStore)->get('key'))->toBe($tenant1->getTenantKey());
+
+    tenancy()->initialize($tenant2);
+
+    expect(cache()->store($cacheStore)->get('key'))->toBe($tenant1->getTenantKey());
+    $this->app->make(SpecificCacheStoreService::class)->handle();
+    expect(cache()->store($cacheStore)->get('key'))->toBe($tenant2->getTenantKey());
+
+    tenancy()->end();
+    expect(cache()->store($cacheStore)->get('key'))->toBe($tenant2->getTenantKey());
```
stancl: This basically checks that the value persists from one context to another, and is only updated upon the `handle()` call?

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+});
+
+test('stores specified in tenantCacheStores get prefixed', function() {
+    // Make the currently used store ('redis') the only store in $tenantCacheStores
+    PrefixCacheTenancyBootstrapper::$tenantCacheStores = ['redis'];
```
stancl: How does this test assert that this is happening specifically with the `redis` cache store?

stancl: The test seems pretty similar to the other tests above. Does it do anything extra? Or is it just present as a "contrast" to the next test (`stores not specified in tenantCacheStores do not get prefixed`)?

If that's the case, maybe you could merge those into one test that's named something like `only stores specified in tenantCacheStores get prefixed` where you'd assert that the specified stores do get prefixed, and the ones not specified don't.

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    expect(cache('key'))->toBe($tenant1->getTenantKey());
+    $this->app->make(CacheService::class)->handle();
+    expect(cache('key'))->toBe($tenant2->getTenantKey());
+
+    tenancy()->end();
+    expect(cache('key'))->toBe($tenant2->getTenantKey());
+});
+
+test('non default stores get prefixed too', function () {
```
stancl: What exactly does this test check? That even stores that aren't default, **but are specified in `tenantCacheStores`**, will get prefixed? If so, the name should be updated since right now it sounds like they get prefixed automatically.

Either something like "non default stores can be prefixed as well" or "non default stores get prefixed too when listed in tenantCacheStores"

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,345 @@
...
+
+    // We didn't add a prefix generator for our 'redis' driver, so we expect the prefix to be generated using the 'default' generator
+    expect(cache()->store()->getPrefix())->toBe($expectedPrefix . ':');
+    // Non-default store
+    expect(cache()->store('redis')->getPrefix())->toBe($expectedPrefix . ':');
+
+    tenancy()->end();
+});
+
+test('cache store prefix generation can be customized', function() {
```
stancl, line 285: This test and the next one seem to be using the outdated concept of specific prefix generators for individual cache stores. Now we only have one generator, so probably:
- update the names of the tests
- instead of `expect(cache()->store('redis')->getPrefix())` (direct `store()` call) do what you're doing in `cache base prefix is customizable` where you check both the cache prefix & the cache store prefix.

Also might be worth using multiple cache stores in these tests (including `cache base prefix is customizable`) because I think calling `cache()->getPrefix()` is just an indirect way of getting the prefix from the default store, so the direct call to the store is redundant. Instead it'd make more sense to have multiple stores and to check that the prefix is set on `cache()->getPrefix()` **and each store**. Not 100% sure but I think it'd make more sense that way.

tests/BootstrapperTest.php
lukinovec: Do you mean doing it like this? https://github.com/archtechx/tenancy/pull/1014/commits/13b85a2d3be9041d069f00f6e7306d4a9ad32290

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
lukinovec: > Initialize -> revert -> initialize and it will be gone (due to line 53).

You're right, I didn't think about that. I'll fix this

tests/PrefixCacheBootstrapperTest.php
lukinovec: It's not set to `true` anywhere in the test, so yeah, I'll  delete this

tests/PrefixCacheBootstrapperTest.php
lukinovec: > Is the point of these repeated assertions ensuring that the cache manager is always overridden, no matter what happens?

Yes

> is there an easy way for people to disable the cache manager override (in case they want to use a custom CacheManager)?

No. The config key idea seems good, though we'll need to mention that CacheTagsBootstrapper only works with our custom CacheManager. Or we can make the CacheManager implement some interface that requires the manager to have the static `$addTags` property (e.g. UsesTags) and require the CacheManager that implements that interface in the bootstrapper.

tests/PrefixCacheBootstrapperTest.php
lukinovec: Right. It's the latter

tests/PrefixCacheBootstrapperTest.php
lukinovec: Good question, it's TTL, but I don't see a reason for setting TTL here (also in some other `cache()->put()` calls in the tests)

lukinovec: I deleted that

tests/PrefixCacheBootstrapperTest.php
lukinovec: Makes sense, added

tests/PrefixCacheBootstrapperTest.php
lukinovec: Yeah, and that the service's `handle()` sets the cache correctly, even though the service is a singleton and the same cache repository is used (the `handle()` call sets `key` to the tenant key)

tests/PrefixCacheBootstrapperTest.php
lukinovec: > is it just present as a "contrast" to the next test

That's right

> maybe you could merge those into one test that's named something like only stores specified in tenantCacheStores get prefixed where you'd assert that the specified stores do get prefixed, and the ones not specified don't

Okay, that makes sense. I'll merge them

tests/PrefixCacheBootstrapperTest.php
lukinovec: Yeah, you got it right. I agree with the name change – I'd go with something like the latter suggestion

tests/PrefixCacheBootstrapperTest.php
stancl: > though we'll need to mention that CacheTagsBootstrapper only works with our custom CacheManager

That's fine.

> Or we can make the CacheManager implement some interface that requires the manager to have the static $addTags property (e.g. UsesTags)

Interfaces can't have properties, only methods. And sounds unnecessarily complex. Leaving a note that X feature only works with that config enabled is completely fine. 99% of people won't ever disable it, it's just to give people who need more customization the option to do so.

lukinovec: Good points, I'll add the config

tests/PrefixCacheBootstrapperTest.php
lukinovec: Something like this should do it: https://github.com/archtechx/tenancy/pull/1014/commits/4c9cb7cf51b9d3002b61b567d8df39a34d4ad7cb

Maybe I'd extract the code into the closure to make the test cleaner/reduce duplicity (tomorrow)

tests/PrefixCacheBootstrapperTest.php
lukinovec: > update the names of the tests

The current name of the first test ("cache store prefix generation can be customized") seems fine to me, agree with the next test's name being inaccurate though

> I think calling cache()->getPrefix() is just an indirect way of getting the prefix from the default store, so the direct call to the store is redundant

Right – `cache()->store()->getPrefix()` is the same as `cache()->getPrefix()`

lukinovec: What do you think about these changes? https://github.com/archtechx/tenancy/pull/1014/commits/ca907e68bed47cf44809ed721ee7d112422997d8

src/Bootstrappers/PrefixCacheTenancyBootstrapper.php
lukinovec: There's now a public `generatePrefix()` method, and the defaulting is inline

tests/PrefixCacheBootstrapperTest.php
lukinovec: The config key (`tenancy.cache.manager`) is there and it's used now, but I didn't add any tests for it yet – I'll finish this tomorrow

tests/PrefixCacheBootstrapperTest.php
stancl: Right, then maybe just add a comment mentioning that (above the lines I selected in the review). So that the point of the code is faster to understand.

tests/PrefixCacheBootstrapperTest.php
lukinovec: Added a todo for the test https://github.com/archtechx/tenancy/pull/1014/commits/effb6a81a1899aa54aa844be8be2ade4d3e1b1bf

tests/PrefixCacheBootstrapperTest.php
lukinovec: I think we can resolve this now, @stancl responded to this on Discord and I updated the code accordingly

assets/config.php
```diff
@@ -189,7 +190,9 @@
      * You can clear cache selectively by specifying the tag.
      */
     'cache' => [
+        'manager' => CacheManager::class,
```
stancl: Is there some benefit to using a class here instead of just a boolean `override_manager` config?

If we did go this route, I'd also inline the full class so that it's clear *which* CacheManager is being used here.

From the POV of developers disabling this to use a custom class, I think it might be a bit confusing to have to remember to go to the tenancy config file to specify which cache manager you want to use. So it might make more sense to have a boolean option here. People would set that to false and then register their own override wherever they want.

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,382 @@
...
+            PrefixCacheTenancyBootstrapper::class
+        ],
+        'cache.default' => $cacheDriver = 'redis',
+        'cache.stores.redis2' => config('cache.stores.redis'),
+    ]);
+
+    PrefixCacheTenancyBootstrapper::$tenantCacheStores = [$cacheDriver];
+    PrefixCacheTenancyBootstrapper::$prefixGenerator = null;
+
+    config('tenancy.cache.manager')::$addTags = false;
```
stancl: This line doesn't really make sense to me. As I wrote in the other review, I'd get rid of this approach to the config completely, but even if we did specify the cache manager in the config, using it like this seems wrong. **Our class** has the `addTags` property. Dynamically fetching the class from the config doesn't seem to add anything here, it just hides the `CacheManager` class (which will always be used here I think?) behind an unnecessary abstraction layer.

stancl: And in the `Tenancy overrides CacheManager` test, just reference the two classes directly. Laravel's CacheManager and our CacheManager (aliased to `TenantCacheManager` or something like that).

tests/PrefixCacheBootstrapperTest.php
```diff
@@ -0,0 +1,382 @@
...
+    $prefixBase = config('tenancy.cache.prefix_base');
+    $getDefaultPrefixForTenant = fn (Tenant $tenant) => $originalPrefix . $prefixBase . $tenant->getTenantKey();
+    $generatePrefixForTenant = function (Tenant $tenant) {
+        return app(PrefixCacheTenancyBootstrapper::class)->generatePrefix($tenant);
+    };
+
+    $expectCachePrefixToBe = function (string $prefix) {
+        expect($prefix . ':') // RedisStore suffixes prefix with ':'
+            ->toBe(app('cache')->getPrefix())
+            ->toBe(app('cache.store')->getPrefix());
```
stancl: Should we use more stores here like we're using in `cache store prefix generation can be customized`?

tests/BootstrapperTest.php
stancl: I see, those are all new features. Should ideally be added there as well. Just a `beforeEach()` and `afterEach()` to set all static properties manipulated in that test file to their default values.

tests/BootstrapperTest.php
stancl: Yeah. Though if `cache.default` is never set to anything except `redis`, there's no need to make this assertion at the start of the file:
```php
expect(config('cache.default'))->toBe('redis');
```

since we can assume that it will have the value defined in `beforeEach()`. The line doesn't hurt, but doesn't seem to add anything either right now, so we can probably remove it.

assets/config.php
lukinovec: Good point. I thought about the whole overriding thing, and I think that in the end, we only need to use the custom CacheManager when using cache tags, so we shouldn't have to override the manager in the TSP – doing so in the CacheTagsBootstrapper should do it (basically the same way [CacheTenancyBootstrapper](https://github.com/archtechx/tenancy/pull/1014/files#diff-7c9f10f1416181e2f3d8294c109feba0e9816a29fdac6974c44443004448394e) was doing it before).

So when we move the 'cache' overriding from TSP to the tag bootstrapper, the user should be able to override the cache manager:
1) whenever he wants, as long as it's before handling the prefix bootstrapper (e.g. in TSP) – if he overrides the manager after the bootstrapper runs, the new manager won't have the prefixes
2) in the CacheTagsBootstrapper by setting a static property (which would be our custom cache manager by default) – since the bootstrapper 

So I think I wouldn't use the config key in the end

tests/PrefixCacheBootstrapperTest.php
lukinovec: I changed the approach so that the config isn't used, and we override CacheManager only when using CacheTagsBootstrapper

https://github.com/archtechx/tenancy/pull/1014#discussion_r1170915056

tests/PrefixCacheBootstrapperTest.php
lukinovec: Makes sense. Also, there was another test that could be using more stores, so I added that. Now, it also makes sense to have both `redis` and `redis2` in `$tenantCacheStores` by default because we mostly want to check if prefixing works for multiple stores, and when we want to check if it works just for specific stores (e.g. in `only the stores specified in tenantCacheStores get prefixed`), we set `$tenantCacheStores` in that test – we end up with less `PrefixCacheTenancyBootstrapper::$tenantCacheStores = [...` calls, so the tests are a bit cleaner in the end (https://github.com/archtechx/tenancy/pull/1014/commits/52d10d36f96c31be590eb4b9d40163e4ab189cd2)

stancl: Set the two stores in `beforeEach()`, and `[]` in `afterEach()`

assets/config.php
lukinovec: I ended up using the `override_manager` config + `$addTags` approach instead

tests/BootstrapperTest.php
lukinovec: https://github.com/archtechx/tenancy/pull/1014/commits/53b2181779baa8541400dd014043636e213e6259

src/CacheManager.php
```diff
@@ -18,7 +18,7 @@ class CacheManager extends BaseCacheManager
      */
     public function __call($method, $parameters)
     {
-        $tags = [config('tenancy.cache.tag_base') . tenant()->getTenantKey()];
+        $tags = [config('tenancy.cache.tag_base') . tenant()?->getTenantKey()];
```
stancl: Any reason for this change?

lukinovec: I don't think there was a reason for this, let me check

lukinovec: Yeah, tests run fine and I can't think of a reason for keeping this change. Reverted this

stancl:
Everything seems to be completed here, so a summary of the changes:
- A new cache bootstrapper is added (prefix). This supports all cache drivers, not just Redis
- The original approach to cache tenancy is **completely unchanged**, however it is not recommended anymore and will likely be deprecated in the next major version (v5)
- The prefix bootstrapper supports all approaches to dependency injection
    - Injecting `CacheManager` and manually fetching the cache store/repository every time it needs to be used (this is identical to calling `cache()` / `Cache::`)
    - Injecting `CacheManager` and storing a specific store/repository in a property
        - Tenancy will update the prefix of these repositories even if they're persisted anywhere in memory
- The tags bootstrapper doesn't support **any** dependency injection properly:
    - The actual tenancy logic is in the CacheManager (`__call()`), not the repositories, so they cannot be injected. All calls must go through the CacheManager
    - The bootstrapper replaces the CacheManager in the container, so it cannot be injected
        - If a specific request happened entirely in the tenant context, without switching back to the central context, you could in theory inject the CacheManager. It will just not update when you revert back to the central context or switch to another tenant
        - For this reason, it's recommended to avoid using *any* cache-related dependency injection with the tags approach

therecluse26:
@stancl does this change fix issues related to why Laravel officially undocumented cache tags (due to memory leaks, according to them)?

stancl:
No, it just lets us support many more drivers more reliably.

PR #1015, opened 2022-11-22
[4.x] Make `TenantDump` work when called outside the tenant context

lukinovec:
Calling the `TenantDump` command outside the tenant context (e.g., `artisan tenants:dump --tenant=acme` in the terminal) dumps the **central** database. This happens because the command doesn't run the parent as the tenant passed in the `--tenant` option.

This PR fixes the problem by moving the command's `parent::handle()` call inside `$tenants->run()`.

### DIFF
--- a/tests/CommandsTest.php
+++ b/tests/CommandsTest.php
@@ -134,7 +134,7 @@ test('dump command generates dump at the passed path', function() {
     expect($schemaPath)->toBeFile();
 });
 
-test('migrate command correctly uses the dump located at the configured schema path by default', function () {
+test('migrate command correctly uses the schema dump located at the configured schema path by default', function () {
     config(['tenancy.migration_parameters.--schema-path' => 'tests/Etc/tenant-schema.dump']);
     $tenant = Tenant::create();
 
--- a/tests/CommandsTest.php
+++ b/tests/CommandsTest.php
@@ -29,10 +29,6 @@ beforeEach(function () {
         unlink($schemaPath);
     }
 
-    if (file_exists($schemaPath = database_path('tenant-schema.dump'))) {
-        unlink($schemaPath);
-    }
-
     Event::listen(TenantCreated::class, JobPipeline::make([CreateDatabase::class])->send(function (TenantCreated $event) {
         return $event->tenant;
     })->toListener());
--- a/tests/CommandsTest.php
+++ b/tests/CommandsTest.php
@@ -25,7 +25,7 @@ use Stancl\Tenancy\Listeners\RevertToCentralContext;
 use Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper;
 
 beforeEach(function () {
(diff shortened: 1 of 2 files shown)


stancl:
Why did it work in tinker? Even without the `$tenant->run()`

lukinovec:
> Why did it work in tinker? Even without the `$tenant->run()`

Did it work for you @stancl? For me, it only works in tinker with `tenancy()->initialize()` or `$tenant->run()`

stancl:
I haven't tested it, but from the title `when called using the terminal` I assumed it does work? I mostly meant to say artisan though, not tinker.

And if we have tests for this, and they're properly written (i.e. context shouldn't be tenant when we're running commands like these), then I'd assume it works well in CLI?

Or does the artisan command work now, but it doesn't work in tinker or when called from an app?

lukinovec:
> I haven't tested it, but from the title `when called using the terminal` I assumed it does work

I see. I renamed the PR

> Or does the artisan command work now, but it doesn't work in tinker or when called from an app?

The command works correctly when called in the tenant context, but outside the tenant context, it dumps the central DB schema which shouldn't be happening.

> if we have tests for this, and they're properly written (i.e. context shouldn't be tenant when we're running commands like these)

I just checked this, and in some tests, the command is called in the tenant context. Also, the dump generated in the tests doesn't get removed after the test stops running, so some tests succeed falsely. 

I'll update the tests in his PR as well

tests/CommandsTest.php
```diff
@@ -111,28 +111,43 @@
...
+    Artisan::call('tenants:migrate');
+
+    expect($schemaPath = 'tests/Etc/tenant-schema-test.dump')->not()->toBeFile();
+
+    Artisan::call("tenants:dump --tenant='$tenant->id' --path='$schemaPath'");
 
+    expect($schemaPath)->toBeFile();
+});
+
+test('tenant dump file gets created at the path specified in the tenancy migration parameters config', function() {
```
stancl: How does this test assert that?

stancl: Seems like the change you made undid that. You should be setting the config key, not passing it via `--path`.

stancl: After this is resolved, double check the diff that everything makes sense and you can squash merge (if GH lets you) since you mentioned that you need this for the testing stuff. If GH doesn't let you squash merge it, you can use the `dev-fix-tenant-dump-command` version constraint locally 👍

lukinovec: Yeah, the test doesn't make sense, and it asserts the same things as the previous one, so I'll delete it.

For the record, I updated the command so that if `--path` isn't passed, `config('tenancy.migration_parameters.--schema-path')` is used instead of the hardcoded default. I updated the test and then remembered that [the default was hardcoded on purpose](https://github.com/archtechx/tenancy/pull/985#issuecomment-1290798819), so I undid the changes, but I forgot to delete this test.

lukinovec: Merged

PR #1017, opened 2022-11-22
Migrate Refresh

i74ifa:
Just add Migrate Refresh
We need a way to refresh migrations files and option a specific file with ```--path``` to refresh it

### DIFF
--- a/src/Commands/MigrateRefresh.php
+++ b/src/Commands/MigrateRefresh.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Commands;
+
+use Illuminate\Console\Command;
+use Stancl\Tenancy\Concerns\DealsWithMigrations;
+use Stancl\Tenancy\Concerns\HasATenantsOption;
+use Symfony\Component\Console\Input\InputOption;
+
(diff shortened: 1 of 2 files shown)


stancl:
1. Why isn't `migrate:fresh` enough for your use case?
2. What's the point of the `--path` argument?

i74ifa:
Migrate fresh will fresh all the files 

But in my case, I want to refresh only one file
I don't know why `tenants:rollback` not accurate when i use `--path`

stancl:
Write that more clearly and in more detail.

i74ifa:
I now have` create_users_table.php` file
  I modified or added some columns
When I want to rollback using a command
`php artisan tenants:rollback --path=database/migration/tenants/create_users_table.php`
It recognizes the file but rollback does not work
This is my problem with rollback

But migrate:refresh

According to Laravel
> The `migrate:refresh `command will roll back all of your migrations and then execute the migrate command. This command effectively re-creates your entire database.


It is working properly

stancl:
I don't think we'll want to include this.

You should make your development environment stable enough so that you can run `migrate:fresh --seed` and be able to test the app, without having to roll back and re-run individual migrations.

PR #1021, opened 2022-11-25
Tenant unaware feature and overall "features" feature improvements 

### LINKED ISSUE #949: [4.x] Features are not registered before Tenancy is resolved from the container
### Bug description

I am trying to redirect the user to the tenant subdomain from the central domain.

I wrote `return redirect()->route('tenant::home')->domain('sub1.dev.local');` in my central controller, but this will raise an exception `Call to undefined method Illuminate\Http\RedirectResponse::domain()`.

After researching, I found that the relevant feature is registered after Tenancy is called, so if Tenancy is never called, the `domain` method will not be added.
https://github.com/archtechx/tenancy/blob/1ea3cefa1d1499211e8659a0b2e050b05ff0c4fe/src/TenancyServiceProvider.php#L31-L37

### Steps to reproduce

```
// routes/web.php
Route::get('/', function () {
//    tenancy();
    return redirect()->route('tenant::home')->domain('sub1.dev.local');
})->name('home');
```
It should raise a BadMethodCallException when you access it.

After uncomment the `// tenancy()`, it should works.

### Expected behavior

Optional features should be bootstrapped no matter the state of Tenancy initialization.

Or we should provide options for choosing which features to initialize ahead of time.

### Laravel version

9.30.1

### stancl/tenancy version

3.5.9

abrardev99:
> todo update description after resolving reviews.

This PR adds the ability to bootstrap features early in the request life cycle. In simple words, You can bootstrap the feature without waiting for tenancy initialization. We are calling these **tenant unaware** feature. 

## Solution 
PR introduces the new config key, which is an array of tenant-unaware features. For now, `CrossDomainRedirect` is added to this array. This also closes https://github.com/archtechx/tenancy/issues/949

## Usage 
You add your own tenant unaware feature in tenancy.php. 
```php
'tenant_unaware_features' => [
        Stancl\Tenancy\Features\CrossDomainRedirect::class, // https://tenancyforlaravel.com/docs/v3/features/cross-domain-redirect
       // ... add yours
    ],
```

## Improvements 
With the old approach, I noticed that the "$tenancy" object is passed but the tenancy is not initialized at this point. This happening because tenancy is not initiated yet. Maybe that's why current feature classes are not making use of the `tenancy` parameter. One feature was [using actually](https://github.com/archtechx/tenancy/pull/1021/files#diff-d9a6d8e4939efecbe554aa8da4fef8eb793bc685541455b28f5e049f4e72e874R21), to add macro, but macro can be added statically like `Tenancy::macro`. 

So the following got changed. 
- Removed the `$tenancy` parameter from the `Feature` interface `bootstrap` method. Now all tenant `unaware` & `aware` features have the same abstractions. 
- Tenant-aware features are being bootstrapped in the `->resolving(Tenancy)` event. That means we can use DI. 

The question can be how to develop tenant-aware features. Most probably using the `Tenancy events`.


tests/Features/RedirectTest.php
```diff
@@ -34,3 +34,20 @@
     expect(tenant_route('foo.localhost', 'foo', ['a' => 'as', 'b' => 'df']))->toBe('http://foo.localhost/abcdef/as/df');
     expect(tenant_route('foo.localhost', 'foo', []))->toBe('http://foo.localhost/abcdef');
 });
+
+test('redirect from central to tenant works fine', function () {
```
stancl: Does this test fail when the feature isn't in the config? It might be a good idea to assert this using a dataset.

Regarding the name:
1. The name should be more descriptive. We don't really care about central/tenant here, we care that it can be used before `tenancy()` is instantiated
2. So I guess we should also check that `Tenancy` is not resolved in the container (not sure what the default lifecycle is for tests here)
3. I wouldn't use `fine` in a test name
4. About point 1): Since this is in `RedirectTest`, it might make sense to use language that's more specific to the feature (`redirect from central to tenant`), rather than what I suggested in 1). But we do want a test that verifies that *the bug itself* is resolved.

So maybe a good solution could be to use a name that's specific to this feature, and add a comment explaining that this fixes a bug?
```suggestion
// Check that `domain()` can be called on a redirect before tenancy is used (regression test for #949)
test('redirect from central to tenant works', function () {
```

src/TenancyServiceProvider.php
```diff
@@ -27,15 +27,16 @@ public function register(): void
...
-                $this->app[$feature]->bootstrap($tenancy);
+                $this->app[$feature]->bootstrap();
             }
-
-            return $tenancy;
         });
 
+        foreach ($this->app['config']['tenancy.tenant_unaware_features'] ?? [] as $feature) {
+            $this->app[$feature]->bootstrap();
+        }
```
stancl: This is basically the main code of this PR, right?

We bootstrap tenant-aware features when `Tenancy` is actually used, and unaware features are bootstrapped **always**.

Could we get rid of the config and use a static method (maybe something like `alwaysBootstrap(): bool`) on the classes instead?

Also, what's the reason for the `extend` -> `resolving` change?

tests/Features/RedirectTest.php
abrardev99: > Does this test fail when the feature isn't in the config?

Yes, it does. 
I see TenancyServiceProvider runs before we add this feature in the config array _in_ tests. So feature never bootstraps. That's why I have to keep the feature in the config already. 

> It might be a good idea to assert this using a dataset.

Do you mean testing both sides? central to the tenant and vice versa?

src/TenancyServiceProvider.php
abrardev99: > This is basically the main code of this PR, right?

Yes. 

> We bootstrap tenant-aware features when Tenancy is actually used, and unaware features are bootstrapped always.

Correct. 

> Could we get rid of the config and use a static method (maybe something like alwaysBootstrap(): bool) on the classes instead?

Yes, this can be an option. I'd try it. 

> Also, what's the reason for the extend -> resolving change?

Mainly, extend doesn't allow using DI while the `resolving` event allows using DI. With `extend` we are going into a never-ending loop but with `resolving` it works fine. I think the core logic is `extend` is called every time a dependency is resolved in a request life cycle while the `resolving` event is fired only once when a class is resolved (Not sure though). That's we are able to DI the `Tenancy` class.

abrardev99: > Could we get rid of the config and use a static method (maybe something like alwaysBootstrap(): bool) on the classes instead?

I tried this and it's working https://github.com/archtechx/tenancy/pull/1021/commits/be6b2f82486d06db635008b823d1030137942ee0. I added this method in the interface so every class implement it.

tests/Features/RedirectTest.php
stancl: > Does this test **fail when the feature isn't in the config**? It might be a good idea to assert `this` using a dataset.

`this` being the highlighted part

tests/Pest.php
```diff
@@ -2,9 +2,16 @@
 
 use Stancl\Tenancy\Tests\TestCase;
 
-uses(TestCase::class)->in(__DIR__);
+uses(TestCase::class)->in(...filesAndFoldersExcluding(['WithoutTenancy']));
 
-function pest(): TestCase
+function pest(): \Orchestra\Testbench\TestCase
```
stancl: This removes IDE support for the methods we've added in our `TestCase`

tests/Pest.php
```diff
@@ -2,9 +2,16 @@
 
 use Stancl\Tenancy\Tests\TestCase;
 
-uses(TestCase::class)->in(__DIR__);
+uses(TestCase::class)->in(...filesAndFoldersExcluding(['WithoutTenancy']));
```
stancl: Can't we just hardcode the directories with normal tests instead of running code to scan them?

assets/config.php
```diff
@@ -283,7 +283,7 @@
         // Stancl\Tenancy\Features\TelescopeTags::class,
         // Stancl\Tenancy\Features\UniversalRoutes::class,
         // Stancl\Tenancy\Features\TenantConfig::class, // https://tenancyforlaravel.com/docs/v3/features/tenant-config
-        // Stancl\Tenancy\Features\CrossDomainRedirect::class, // https://tenancyforlaravel.com/docs/v3/features/cross-domain-redirect
+        Stancl\Tenancy\Features\CrossDomainRedirect::class, // https://tenancyforlaravel.com/docs/v3/features/cross-domain-redirect
```
stancl: Probably wouldn't change the default here — the features are *optional features*, so that implies they should all be disabled by default.

If you need to change something in the config for tests, use the `TestCase` for that.

tests/Pest.php
abrardev99: But we get the return type error when using `WithoutTenancy` TestCase.

tests/Pest.php
abrardev99: Is it okay to hardcode folders and files (~20+)?

assets/config.php
abrardev99: > the features are optional features, so that implies they should all be disabled by default.

I thought that's wht we moved to a separate TestCase thing. 

> If you need to change something in the config for tests, use the TestCase for that.

it has the same issue. SP runs before TestCase. always bootstrap feature never registers in this way. That's why I moved the whole RedirecTest suite into without tenancy.

tests/Pest.php
stancl: Ah I thought that they're all in some folder already. You can stick with the code you have here then, and add a `// todo` to move all tests to a separate folder.

assets/config.php
stancl: Don't really get your comment.

> I thought that's wht we moved to a separate TestCase thing.

And here you're saying that it's **not** the reason. That it has to be moved anyway:
> it has the same issue. SP runs before TestCase. always bootstrap feature never registers in this way. That's why I moved the whole RedirecTest suite into without tenancy.

So that seems contradictory. Either way, it's not what I'm suggesting — **I'm not talking about tests at all here, or the reason for moving anything**. I'm talking about the fact that it should remain disabled by default.

I don't think the test being separate has anything to do with whether it's enabled or disabled by default. We're using a dataset so both cases are covered. The test is separate to avoid running the service provider. I'm not suggesting that this change has anything to do with the tests, I'm simply telling you that the feature should remain optional like it was in v3.

So I don't get what you mean by your comment.

stancl: What I meant by this:

> If you need to change something in the config for tests, use the TestCase for that.

Is that **if you made this change because of tests**, change the config that's set inside `TestCase` (open the file to see what I mean).

If you made this change for any other reason, then you can share it or revert the change.

I'm not saying that this is because of tests. I just went a step ahead and said that **if** this is due to tests, then you can set the config in the `TestSuite`.

Or is your entire point that `getEnvironmentSetUp()` won't work because the SP runs before that, so to even test the redirect in the **normal** tests we need this to be enabled in the config? I guess that is what you mean here, but your comment is pretty confusing so it only occurred to me now that this may be what you mean. Please confirm if that's what you were trying to say.

If that's the case, then you might be able to set the config using some `resolving()` hook or something like that in `setUp()` of that test. If that won't work, then maybe changing the features to be bootstrapped using `extend()` instead of `resolving()` might make `resolving()` work for changing the config.

Regarding the `extend()` / `resolving()`, I think I had another review for that. We should be intentional about whether we use `extend()` or `resolving()`, it should have a reason.

tests/Features/RedirectTest.php
abrardev99: I believe this review is resolved now.

assets/config.php
abrardev99: > Or is your entire point that getEnvironmentSetUp() won't work because the SP runs before that, so to even test the redirect in the normal tests we need this to be enabled in the config? I guess that is what you mean here, but your comment is pretty confusing so it only occurred to me now that this may be what you mean. Please confirm if that's what you were trying to say.

Yes that's what I was trying to say. 

> If that won't work, then maybe changing the features to be bootstrapped using extend() instead of resolving() might make resolving() work for changing the config

`resolving` container event works but we want to always bootstrap feature. The bootstrapping feature in `resolving` means the feature is not always bootstrap. The issue attached to this PR explains that OP wants to use the feature without using `Tenancy`.

stancl: Ah right, I remembered that we had this code in an earlier version of the PR but forgot that it was changed later because of this.

So do you not see any solutions for enabling the feature in the config, in tests? If you don't, I'll find one and finish this PR myself.

abrardev99: Enabling the feature in `getPackageProviders`  works but not sure if this is clean/right approach. 

```php
    protected function getPackageProviders($app)
    {
        config()->set('tenancy.features', [CrossDomainRedirect::class]);

        return [
            TenancyServiceProvider::class,
        ];
    }
```

I'm leaving this review open. If you can find a better approach, Please finish PR.

tests/Pest.php
abrardev99: Looks like returning `TestCase` works now. Previously, the `WithoutTenancy` TestCase was extending `\Orchestra\Testbench\TestCase` but now I'm extending it from our `Stancl/Tests/TestCase`.

assets/config.php
stancl: I think I'll add a static property for disabling parts of the Tenancy SP, and a method for running them manually at a point that suits us in tests. Will complete this tomorrow probably.

lukinovec: I added the property, and I updated the tests so that the features get bootstrapped (manually). I think we can resolve this after your review @stancl

stancl:
Resubmitted in v4 repo

PR #1025, opened 2022-11-28
Improve code of pending tenants

lukinovec:
While documenting the pending tenants feature, I discovered that the `ClearPendingTenants` command probably doesn't need the `--all` option.

The option was there so that the time constraint options (e.g. `--older-than-days`) from config could have been overridden. Now that the options can only be passed, I don't think the option is needed.

### DIFF
--- a/src/Commands/ClearPendingTenants.php
+++ b/src/Commands/ClearPendingTenants.php
@@ -23,8 +23,8 @@ class ClearPendingTenants extends Command
         // We compare the original expiration date to the new one to check if the new one is different later
         $originalExpirationDate = $expirationDate->copy()->toImmutable();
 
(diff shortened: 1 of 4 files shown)


stancl:
Why did we get rid of the config again? Was the reason that people can just specify the constraints in their Console Kernel where they schedule the jobs?

lukinovec:
> Why did we get rid of the config again? Was the reason that people can just specify the constraints in their Console Kernel where they schedule the jobs?

Yeah, here's the conversation where we discussed this: https://github.com/archtechx/tenancy/pull/869/files/4d7831d6f6a172760dd54f64e527ea9895768e02#r930881310

PR #1027, opened 2022-12-07
[4.x] Make broadcasting work with Tenancy

lukinovec:
This PR makes broadcasting work with Tenancy using public, private, and presence channels. It also allows using different broadcaster credentials for each tenant.

`BroadcastTenancyBootstrapper` maps the tenant attributes to config so that the broadcasters use the credentials of the current tenant. It also overrides the `BroadcastManager` singleton with an instance of `TenancyBroadcastManager`.

`TenancyBroadcastManager` re-resolves the broadcasters instead of getting them from the `$drivers` property so that the broadcasters always use credentials from the current config.

### The issue with re-resolving the broadcasters

We need to make the broadcasters use the credentials from the config every time we initialize tenancy. Also, the freshly resolved broadcasters need to have all the channels registered in `routes/channels.php`, or else broadcasting won't work.

Ideally, we'd just forget the `BroadcastManager` instance without overriding it, [similarly to the mail credentials PR](https://github.com/archtechx/tenancy/pull/989/files#diff-421166a13b8e50c082716efce511eec8587457ce0cac287118d5de829b4a4980R47), but we'd lose the cached channels. So instead of forgetting `BroadcastManager`, I'm extending it with the custom `TenancyBroadcastManager`, and while re-resolving the broadcasters, I'm also ["passing" the channels of the original broadcaster to the new broadcaster instance](https://github.com/archtechx/tenancy/pull/1027/files#diff-bdb4a81bb3878579e3dc2a44ac8d532baa93e9f82eb525eb6f3cc2d7bdf42077R42) so that the newly resolved broadcasters have the channels too.


stancl:
The implementation looks reasonable at a quick glance. I guess you can write tests (that might help improve some details in the implementation) and then I'll do an in-depth review of all code 👍🏻

lukinovec:
@stancl I wrote the docs [here](https://github.com/stancl/tenancy-docs/issues/229). They may be a bit rough, but things might change in this PR, so I'll revisit the docs after you review it

danish1658:
@lukinovec great work man, public and private channels do work but have you tried presence channel with this, as it’s throwing 500 error

lukinovec:
Hey @danish1658, presence channels were working fine when I was trying them. Could you give me some more details please? What does the 500 error say in the failing request's preview in the "Network" tab?

I tried presence channels again now, and I can't reproduce the 500 error

tests/BootstrapperTest.php
```diff
@@ -328,6 +332,82 @@
     expect(File::isDirectory($tenantStoragePath))->toBeFalse();
 });
 
+test('BroadcastTenancyBootstrapper binds TenancyBroadcastManager to BroadcastManager::class and reverts the binding when tenancy is ended', function() {
```
stancl:
```suggestion
test('BroadcastTenancyBootstrapper binds TenancyBroadcastManager to BroadcastManager and reverts the binding when tenancy is ended', function() {
```

The `::class` here is redundant, it's just a way of telling PHP to convert something to a string. But the actual class name is `Namespace\BroadcastManager`, without the `::class`.

Also, we try to avoid using special characters in test names.

src/TenancyBroadcastManager.php
```diff
@@ -0,0 +1,65 @@
...
+
+        return parent::get($name);
+    }
+
+    // Because, unlike the original broadcaster, the newly resolved broadcaster won't have the channels registered using routes/channels.php
+    // Using it for broadcasting won't work, unless we make it have the original broadcaster's channels
+    protected function passChannelsFromOriginalBroadcaster(Broadcaster $originalBroadcaster, Broadcaster $newBroadcaster): void
+    {
+        // invade() because channels can't be retrieved through any of the broadcaster's public methods
+        $originalBroadcaster = invade($originalBroadcaster);
```
stancl, line 59: `invade()` is only in our dev dependencies, so this won't work in real applications. I think we should just make it a normal dependency

stancl:
Other than https://github.com/archtechx/tenancy/pull/1027#discussion_r1110985068, the PR now looks ready. I'll fix that and merge.

PR #1029, opened 2022-12-13
Add session state when impersonating tenant 

### LINKED ISSUE #548: Highlight when Impersonating

abrardev99:
Closes #548 

# Problem 
When impersonating a tenant, there is no indication of whether we are impersonating. A common use case shows a banner saying "Impersonating" and a button to stop/leave impersonation. 

# Solution 
This PR adds the session key `tenancy_impersonation` when impersonating a tenant. Now this session key can be used to show banner etc. There are also new methods added. 

- `isImpersonating()` to check if the tenant is being impersonated
- `leave()` to leave impersonation. It logout the user and remove the session impersonation state. **You need to handle redirection here**, `leave()` doesn't enforce any redirection. 

## Usage 
[Setup Impersonation](https://tenancyforlaravel.com/docs/v3/features/user-impersonation) same as you do in V3. You can use the impersonation state as follows: 
```php
// TRUE if you are impersonating.
UserImpersonation::isImpersonating();

// Logout from the current domain and forget impersonation session.
UserImpersonation::leave();
```
You need to handle redirection when leaving impersonation. Something like this in tenant routes: 
```php
Route::get('impersonantion/leave', function() {
  UserImpersonantion::leave();
  
  return redirect()->to(config('app.url') . '/dashboard') // redirect to central domain dashboard. 
}) 
```

## Important points you should know about the Impersonation feature in this package. 

- When using domain/subdomain identification, the central or admin user auth session will be preserved because each tenant has a different domain. I'm simple words, You can have two tabs side by side one with an admin logged-in user and one with tenant impersonation. 
- When using path/request data identification, the central or admin user auth session will be destroyed because it's on the same domain. In simple words, the admin will be logged out because we are starting another auth session on the same domain. 
- Setting the `SESSION_DOMAIN` env var can change the behavior of this feature, e.g., sharing sessions with subdomains will destroy auth sessions from other domains. 

### Working when using domain/subdomain

https://user-images.githubusercontent.com/54532330/207320271-65501f21-751d-4b8c-b5dd-d8cdb744f653.mov

### Working when using path identification

https://user-images.githubusercontent.com/54532330/207322104-656cd932-ad90-419d-a6e7-c3c36d7c2817.mov

### DIFF
--- a/src/Features/UserImpersonation.php
+++ b/src/Features/UserImpersonation.php
@@ -48,6 +48,23 @@ class UserImpersonation implements Feature
 
         $token->delete();
 
+        session()->put('tenancy_impersonation', true);
+
         return redirect($token->redirect_url);
+    }
+
+    public static function isImpersonating(): bool
+    {
+        return session()->has('tenancy_impersonation');
+    }
+
+    /**
+     * Logout and forget session
+     */
+    public static function stop(): void
+    {
+        auth()->logout();
+
+        session()->forget('tenancy_impersonation');
     }
 }
--- a/src/Features/UserImpersonation.php
+++ b/src/Features/UserImpersonation.php
@@ -59,7 +59,7 @@ class UserImpersonation implements Feature
     }
 
     /**
-     * Logout and forget session
+     * Logout and forget session.
      */
     public static function stop(): void
     {
--- a/src/Features/UserImpersonation.php
+++ b/src/Features/UserImpersonation.php
@@ -48,14 +48,14 @@ class UserImpersonation implements Feature
 
         $token->delete();
 
-        session()->put('tenancy_impersonation', true);
+        session()->put('tenancy_impersonating', true);
 
         return redirect($token->redirect_url);
     }
 
     public static function isImpersonating(): bool
     {
-        return session()->has('tenancy_impersonation');
+        return session()->has('tenancy_impersonating');
     }
 
     /**
@@ -65,6 +65,6 @@ class UserImpersonation implements Feature
     {
         auth()->logout();
 
-        session()->forget('tenancy_impersonation');
+        session()->forget('tenancy_impersonating');
     }
 }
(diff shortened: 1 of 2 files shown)


src/Features/UserImpersonation.php
```diff
@@ -48,6 +48,23 @@ public static function makeResponse(string|ImpersonationToken $token, int $ttl =
 
         $token->delete();
 
+        session()->put('tenancy_impersonation', true);
```
stancl: should be `impersonating` since it's a bool

src/Features/UserImpersonation.php
```diff
@@ -48,6 +48,23 @@ public static function makeResponse(string|ImpersonationToken $token, int $ttl =
...
         return redirect($token->redirect_url);
     }
+
+    public static function isImpersonating(): bool
+    {
+        return session()->has('tenancy_impersonation');
+    }
+
+    /**
+     * Logout and forget session.
```
stancl:
```suggestion
     * Logout from the tenant domain and forget session.
```
Would this be more accurate?

abrardev99: > Would this be more accurate?

It can't be domain always. This action can be performed when using other identification methods.

stancl: Will just go with this then, per our BC convo:
```suggestion
     * Logout from the current domain and forget impersonation session.
```

abrardev99:
@stancl Related PR and issue should be closed. 
https://github.com/archtechx/tenancy/issues/548
https://github.com/archtechx/tenancy/pull/917

PR #1030, opened 2022-12-14
Improve commands CLI output

abrardev99:
This is a simple PR that improves the CLI output using the `components->info/error` methods. Here are some examples. 

Before
Screenshot 2022-12-14 at 3 02 08 PM
Screenshot 2022-12-14 at 3 04 26 PM

After 
Screenshot 2022-12-14 at 3 03 08 PM
Screenshot 2022-12-14 at 3 05 15 PM

### DIFF
--- a/src/Commands/ClearPendingTenants.php
+++ b/src/Commands/ClearPendingTenants.php
@@ -17,7 +17,7 @@ class ClearPendingTenants extends Command
 
     public function handle(): int
     {
-        $this->info('Removing pending tenants.');
+        $this->components->info('Removing pending tenants.');
 
         $expirationDate = now();
         // We compare the original expiration date to the new one to check if the new one is different later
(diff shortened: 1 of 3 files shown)


src/Commands/ClearPendingTenants.php
```diff
@@ -27,8 +27,8 @@ public function handle(): int
         $olderThanHours = (int) $this->option('older-than-hours');
 
         if ($olderThanDays && $olderThanHours) {
-            $this->line("<options=bold,reverse;fg=red> Cannot use '--older-than-days' and '--older-than-hours' together \n"); // todo@cli refactor all of these styled command outputs to use $this->components
-            $this->line('Please, choose only one of these options.');
+            $this->components->error("Cannot use '--older-than-days' and '--older-than-hours' together");
+            $this->components->error('Please, choose only one of these options.');
```
lukinovec:
```suggestion
            $this->components->error("Cannot use '--older-than-days' and '--older-than-hours' together. Please, choose only one of these options.");
```

This is just a small detail, but I don't like displaying "Please, choose..." as a separate error because it isn't an error. Maybe there's a better `$this->components` method for this (maybe `info` would be okay?), but I think merging these lines would be even better

lukinovec (commented):
The PR looks good @abrardev99. I only left a small suggestion

src/Commands/ClearPendingTenants.php
abrardev99: Merging them looks better

stancl:
Thanks for the PR!

PR #1040, opened 2022-12-29
Allow defining the tenant connection template using array syntax in config

### LINKED ISSUE #529: Allow defining the tenant connection template using array syntax in config
#### Description

Allow defining the template connection using an array, like this:

```php
'template_tenant_connection' => [
    'driver' => 'mysql',
    // ...
],
```

And maybe (?) merge it with the central connection? Because if a completely custom connection was needed, we'd be using one created in the database config.

#### Why this should be added

To avoid having to create a template connection in `config/database.php`.

abrardev99:
Closes https://github.com/archtechx/tenancy/issues/529

This PR allows defining the template connection using an array or partial array (partial database configuration).  When the configuration is partial, it will use the rest from the central connection.

## Usages 

1. Using full database configuration array like this: 
```php
'template_tenant_connection' => [
    'driver' => 'mysql',
    'url' => null,
    'host' => 'mysql2',
    'port' => '3306',
    'database' => 'main',
    'username' => 'root',
    'password' => 'password',
    'unix_socket' => '',
    'charset' => 'utf8mb4',
    'collation' => 'utf8mb4_unicode_ci',
    'prefix' => '',
    'prefix_indexes' => true,
    'strict' => true,
    'engine' => null,
    'options' => [],
],
```

2. Using partial database configuration. The rest will be picked from the central connection.
```php
'template_tenant_connection' => [
    'host' => 'mysql2',
],
```

3. You can still use the connection name from `config/database.php` like this: 
```php
'template_tenant_connection' => 'mysql'
```

### DIFF
--- a/src/Database/DatabaseConfig.php
+++ b/src/Database/DatabaseConfig.php
@@ -87,7 +87,7 @@ class DatabaseConfig
     {
         $this->tenant->setInternal('db_name', $this->getName());
 
-        if ($this->connectionDriverManager($this->getTemplateConnectionName()) instanceof Contracts\ManagesDatabaseUsers) {
+        if ($this->connectionDriverManager($this->getTemplateConnectionDriver()) instanceof Contracts\ManagesDatabaseUsers) {
             $this->tenant->setInternal('db_username', $this->getUsername() ?? (static::$usernameGenerator)($this->tenant));
             $this->tenant->setInternal('db_password', $this->getPassword() ?? (static::$passwordGenerator)($this->tenant));
         }
@@ -97,11 +97,24 @@ class DatabaseConfig
         }
     }
 
-    public function getTemplateConnectionName(): string
+    public function getTemplateConnectionDriver(): string
     {
-        return $this->tenant->getInternal('db_connection')
-            ?? config('tenancy.database.template_tenant_connection')
-            ?? config('tenancy.database.central_connection');
+        return $this->getTemplateConnection()['driver'];
+    }
+
+    public function getTemplateConnection(): array
+    {
+        if ($template = $this->tenant->getInternal('db_connection')) {
+            return config("database.connections.{$template}");
+        }
+
+        if ($template = config('tenancy.database.template_tenant_connection')) {
+            return is_array($template) ? $template : config("database.connections.{$template}");
+        }
+
+        $template = config('tenancy.database.central_connection');
+
+        return config("database.connections.{$template}");
     }
 
     public function getTenantHostConnectionName(): string
@@ -114,8 +127,7 @@ class DatabaseConfig
      */
     public function connection(): array
     {
-        $template = $this->getTemplateConnectionName();
-        $templateConnection = config("database.connections.{$template}");
+        $templateConnection = $this->getTemplateConnection();
 
         return $this->manager()->makeConnectionConfig(
             array_merge($templateConnection, $this->tenantConfig()),
@@ -129,10 +141,9 @@ class DatabaseConfig
     public function hostConnection(): array
     {
         $config = $this->tenantConfig();
-        $template = $this->getTemplateConnectionName();
-        $templateConnection = config("database.connections.{$template}");
+        $templateConnection = $this->getTemplateConnection();
 
-        if ($this->connectionDriverManager($template) instanceof Contracts\ManagesDatabaseUsers) {
+        if ($this->connectionDriverManager($this->getTemplateConnectionDriver()) instanceof Contracts\ManagesDatabaseUsers) {
             // We're removing the username and password because user with these credentials is not created yet
             // If you need to provide username and password when using PermissionControlledMySQLDatabaseManager,
             // consider creating a new connection and use it as `tenancy_db_connection` tenant config key
--- a/src/Database/DatabaseConfig.php
+++ b/src/Database/DatabaseConfig.php
@@ -109,12 +109,17 @@ class DatabaseConfig
         }
 
         if ($template = config('tenancy.database.template_tenant_connection')) {
-            return is_array($template) ? $template : config("database.connections.{$template}");
+            return is_array($template) ? array_merge($this->getCentralConnection(), $template) : config("database.connections.{$template}");
         }
 
-        $template = config('tenancy.database.central_connection');
+        return $this->getCentralConnection();
+    }
 
-        return config("database.connections.{$template}");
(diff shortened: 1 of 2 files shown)


src/Database/DatabaseConfig.php
```diff
@@ -97,11 +97,24 @@ public function makeCredentials(): void
...
+    }
+
+    public function getTemplateConnection(): array
+    {
+        if ($template = $this->tenant->getInternal('db_connection')) {
+            return config("database.connections.{$template}");
+        }
+
+        if ($template = config('tenancy.database.template_tenant_connection')) {
+            return is_array($template) ? $template : config("database.connections.{$template}");
```
stancl: I'm thinking that it should be possible to define the template connection only partially and use the rest from the central connection. So it'd essentially be `array_merge($template, $centralConnection)`.

However, that might lead to some issues e.g. when the central connection has a `url` defined and we don't want to merge that with the central connection.

But maybe in those scenarios the values could be explicitly defined in the template, e.g. `'url' => null`.

If we use the `array_merge()` I suggested above, the central connection has a `url`, and the template has `'url' => null`, will the resulting connection have `'url' => null`?

abrardev99: > will the resulting connection have 'url' => null?

No. The result will be `url` not `null`. This is because `array_merge` merges the second array into first array. For example
```php
array_merge(['url' => null], ['url' => 'example.com'])

// output 
[
  "url" => "example.com",
]
```

Now if we switch sides: 
```php
array_merge(['url' => 'example.com'], ['url' => null]);

// output 
[
  "url" => null,
]
```

I can explore/write tests for these and implement them this way if it works as expected. With the second approach, we need to define every field in the `$template` we don't want to use from `$centralConnection`.

abrardev99: Implemented the ability to define the connection partial configuration. `array_merge($centralConnection, $template)` works as expected. Added a test and update the PR description.

stancl: Right, I meant opposite — `array_merge($centralConnection, $template)`. I typed that part of the code wrong but it should be clear what I meant from the rest of the comment.

tests/TenantDatabaseManagerTest.php
```diff
@@ -390,6 +390,81 @@
     expect(file_exists($customPath . '/' . $name))->toBeTrue();
 });
 
+test('template tenant connection value can be connection name or connection array', function () {
```
lukinovec: I think the name should suggest that the `connection` can be either the name of an existing configured connection template, and that instead of using that existing template, you can also define a new one

```suggestion
test('you can define the connection template array in template_tenant_connection or use a name of an existing connection template', function () {
```

Something like this. Do you agree?

tests/TenantDatabaseManagerTest.php
```diff
@@ -390,6 +390,81 @@
...
+        'tenancy_db_name' => $name,
+    ]);
+
+    /** @var MySQLDatabaseManager $manager */
+    $manager = $tenant->database()->manager();
+    expect($manager->databaseExists($name))->toBeTrue();
+    expect($manager->database()->getConfig('host'))->toBe('mysql2');
+});
+
+test('you can define the partial connection template array in template_tenant_connection and it will be completed by merging into the central connection', function () {
```
lukinovec: I don't really like the term "partial" here, it makes things a bit less concise and harder to understand in my opinion. I'd count on people understanding`array_merge` because it's a pretty known PHP thing – the information that the template is obtained like `array_merge($central, $tenant)` (wrote concisely) is the important thing to test I think.

The main thing to improve here is the conciseness – the template is used like `array_merge($centralTemplate, $tenantTemplate)`, which the test name now does indicate, but it could be worded better.

```suggestion
test('template_tenant_connection will be merged into the central connection template using array_merge($central, $tenant)', function () {
```
My suggestion may not be the best, but I'd use something similar to it

abrardev99: Partially agree with you. But the feature we want to test is `we can define partial connection config`. I don't think this name describes that.

lukinovec: I see what you mean. I'll try thinking of a better name

lukinovec: > the feature we want to test is we can define partial connection config

So to confirm – we want to test that if `template_tenant_connection` overrides the central connection template, and if `template_tenant_connection` is missing some things that the central template has, `template_tenant_connection` gets them?

lukinovec: Btw apart from this detail, I think the PR is good

abrardev99: > So to confirm – we want to test that if template_tenant_connection overrides the central connection template, and if template_tenant_connection is missing some things that the central template has, template_tenant_connection gets them?

Correct. But I'm not sure if we can say it _override_. Its actually merging `template` into `central`. Override gives a feeling of _replacing_ everything.

lukinovec: Right. But I don't think I can come up with a better name than the one I suggested [here](https://github.com/archtechx/tenancy/pull/1040#discussion_r1069483913) – maybe we could even omit the `array_merge` part. @stancl, what do you think?

stancl: Test names shouldn't include special characters or code

stancl: Or really implementation details, just describe what is being done. We're merging the connection into the central connection. That's fine I think.

lukinovec:
```suggestion
test('configured tenant connection template gets merged into the central connection template', function () {
```

@stancl is this fine?

stancl: `gets merged` instead of `merges`, otherwise it's okay

tests/TenantDatabaseManagerTest.php
```diff
@@ -390,6 +390,81 @@
     expect(file_exists($customPath . '/' . $name))->toBeTrue();
 });
 
+test('you can define the connection template array in template_tenant_connection or use a name of an existing connection template', function () {
```
stancl: @lukinovec make a similar naming change here as well — `you can` isn't ideal

stancl: actually I'll rename the tests myself, nevermind

Issue #1018, opened 2022-11-23
Queued Notifications are not tenant aware and produces wrong Action links 


chibyk101:
### Bug description

I want to notify a user after registering to follow a link in the email to get to the tenant's application. And I want to queue this notification but notifications sent in a queue by implementing the ShouldQueue interface will return a weird action link that looks so  

localhost/{central_domin}/......

Instead of the tenant's domain, Localhost is treated as the tenant domain and the central domain is like a path

In the config/tenancy.php I removed every other domain in the central domain key

```php 
'central_domains' => [
      'mydomain.test',
  ],
```
  
Nothing changed.

```php
/**
     * Tenancy bootstrappers are executed when tenancy is initialized.
     * Their responsibility is making Laravel features tenant-aware.
     *
     * To configure their behavior, see the config keys below.
     */
    'bootstrappers' => [
        Stancl\Tenancy\Bootstrappers\DatabaseTenancyBootstrapper::class,
        Stancl\Tenancy\Bootstrappers\CacheTenancyBootstrapper::class,
        Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper::class,
        Stancl\Tenancy\Bootstrappers\QueueTenancyBootstrapper::class,
        // Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper::class, // Note: phpredis is needed
    ],
```

### Steps to reproduce

- Create a tenant

- Register a user for a tenant

- notify the user

```PHP
$user->notify(new WelcomeNotification)
```

- In the notification class implement the ShouldQueue Interface

```PHP
use Illuminate\Bus\Queueable;
use Illuminate\Notifications\Notification;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Notifications\Messages\MailMessage;

class NewStudentRegistration extends Notification implements ShouldQueue
{
  use Queueable;

  public $password;
   ...
}
```
- Listen for queues with PHP artisan queue:listen

I'll also love to mention that i use nwidart/laravel-modules just to be sure it's not the cause

### Expected behavior

Emails sent to users should include the correct Tenant URL

like:  http:\\\tenant-domain.com/login

### Laravel version

9.19

### stancl/tenancy version

3.6

stancl:
This isn't a bug because there's no such thing as a "tenant URL". A tenant can have 10 different domains. There's no way for the package to build a tenant URL.

[...]

PR #1044, opened 2023-01-12
[4.x] Allow user to customize tenant's URL root in CLI

lukinovec:
solves https://github.com/archtechx/tenancy/issues/1018

This PR adds a bootstrapper that lets the user define a  closure that returns the URL root they want, and if the closure exists, override the root url by basically calling `app()->make('url')->forceRootUrl($url)`


assets/TenancyServiceProvider.stub.php
```diff
@@ -118,7 +118,20 @@ public function events()
 
     public function register()
     {
-        //
+        /**
```
stancl, line 123: 1. I'd move this to a separate method and call that from here
2. Can we use `boot()` instead?

lukinovec: Good idea with the separate method. Calling the method in `boot()` seems to be working fine

lukinovec: I updated the code and tested it. I also tested if the URLs are generated correctly in the emails – they are. So I think you can do a final review now @stancl

tests/BootstrapperTest.php
```diff
@@ -380,3 +381,30 @@ function getDiskPrefix(string $disk): string
...
+    config(['tenancy.bootstrappers.url' => UrlTenancyBootstrapper::class]);
+
+    UrlTenancyBootstrapper::$rootUrlOverride = function (Tenant $tenant) {
+        $baseUrl = URL::to('/');
+        $scheme = str($baseUrl)->before('://') . '://';
+
+        return str($baseUrl)
+            ->after($scheme)
+            ->prepend($tenant->getTenantKey() . '.')
+            ->prepend($scheme)
```
stancl: This syntax is a bit confusing, so for the docs we may want to think of something else (e.g. `replaceFirst($scheme, $scheme . "://$subdomain.")` or something like that)

stancl: Actually it's not just in the docs, it's also in the code example. So try to rewrite it into something that's intuitively understandable and then we can merge the PR. The implementation logic looks good to me.

tests/BootstrapperTest.php
```diff
@@ -380,3 +381,30 @@ function getDiskPrefix(string $disk): string
...
+
+     expect(URL::to('/'))->not()->toContain($tenant->getTenantKey());
+
+     tenancy()->initialize($tenant);
+
+     expect(URL::to('/'))->toContain($tenant->getTenantKey());
+
+     tenancy()->end();
+
+     expect(URL::to('/'))->not()->toContain($tenant->getTenantKey());
```
stancl: Maybe add direct assertions about the routes, so that we know that the generated URLs don't have e.g. duplicate `/`s etc. Rather than just "contains" assertions.

src/Bootstrappers/UrlTenancyBootstrapper.php
```diff
@@ -0,0 +1,34 @@
...
+use Stancl\Tenancy\Contracts\Tenant;
+
+class UrlTenancyBootstrapper implements TenancyBootstrapper
+{
+    public static Closure|null $rootUrlOverride = null;
+    protected string|null $originalRootUrl = null;
+
+    public function __construct(protected UrlGenerator $urlGenerator)
+    {
+    }
```
stancl:
```suggestion
    public function __construct(
        protected UrlGenerator $urlGenerator,
    ) {}
```
We put the parameters on separate lines when using DI using parameters

lukinovec: Right, I keep forgetting this. Though multiple lines feel natural when there are multiple parameters

tests/BootstrapperTest.php
lukinovec: I updated the test. I also thought it was a good idea to check if `url(route())` generates the URL correctly. Could you take a look at [the commit](https://github.com/archtechx/tenancy/pull/1044/commits/3fbf6b4a45932a0d4ccbae3c363d2198fa9e4648) and let me know what you think now?

tests/BootstrapperTest.php
lukinovec: Updated this https://github.com/archtechx/tenancy/pull/1044/commits/3fbf6b4a45932a0d4ccbae3c363d2198fa9e4648#diff-b22931f4c76926f01513d299f2f87fb1528e0d3a92e539ef66cb4675b4342f4bR401-R406

src/Bootstrappers/UrlTenancyBootstrapper.php
stancl: It's not just about multiple parameters. It's so that it's clear that we're using this for **creating the properties as well**, and not just injecting dependencies. The `protected` part can be easy to miss when it's all on one line.

tests/BootstrapperTest.php
```diff
@@ -380,3 +383,45 @@ function getDiskPrefix(string $disk): string
 
     return $prefix;
 }
+
+test('url bootstrapper overrides the root url when tenancy gets initialized and reverts the url to the central one after tenancy ends', function() {
+    config(['tenancy.bootstrappers.url' => UrlTenancyBootstrapper::class]);
+
+    $routeName = 'home';
+    $routeUri = '/';
```
stancl: If these never change, the code might be more understandable with those values hardcoded rather than referencing a variable.

tests/BootstrapperTest.php
stancl: Should be updated in the comment example as well.

stancl: And you can probably keep all of it inside the closure, that would be cleaner. Just avoid using `after()->prepend()->prepend()`

tests/BootstrapperTest.php
lukinovec: I hardcoded the values here: https://github.com/archtechx/tenancy/pull/1044/commits/28adb5692d26046e48c5da15371ef1bb6097d9a9

tests/BootstrapperTest.php
lukinovec: [I moved the variables to the closure](https://github.com/archtechx/tenancy/pull/1044/commits/7bcf9d329ac91d83d0f01ce4a157b9f2dd8f8f72) (except for `$baseUrl`), and [updated the comment example](https://github.com/archtechx/tenancy/pull/1044/commits/5e78d03ea40fa05310863bedadb20a8774ff18a6)

assets/TenancyServiceProvider.stub.php
lukinovec: @stancl could you please take a look at the TenancyServiceProvider stub again now that it's updated? 
(the code example and the `overrideUrlInTenantContext` method + its call) https://github.com/archtechx/tenancy/pull/1044/files#diff-6ca30424f3bab5bd30feedc72b4d4e930244bc8f73fc6828396beb574df18fd0

tests/BootstrapperTest.php
stancl: I was thinking about using literal strings (`http://tenant1.localhost/foo`), but checking that it equals to the value returned from the closure is probably good enough since it checks directly that the closure is being used 👍🏻

PR #1057, opened 2023-02-08
Add AuthTenancyBootstrapper.

### LINKED ISSUE #1052: Error when issuing a password reset for multiple tenants
### Bug description

When a password reset is issued in tenant context ~~inside a queued job after a tenant is created~~, an error is thrown the second time a tenant switch happens. 

### Steps to reproduce

Given the following command, the command fails on the second tenant in the system.

```php
<?php

namespace App\Console\Commands;

use App\Models\Tenant;
use App\Models\User;
use Illuminate\Auth\Notifications\ResetPassword;
use Illuminate\Support\Facades\Password;
use Stancl\Tenancy\Concerns\TenantAwareCommand;

class CreateDummyUser extends Command
{
    use TenantAwareCommand;

    protected $signature = 'tenants:create:dummy-user';

    public function handle(Tenant $tenant): int {
        $user = User::factory()->create();
        Password::sendResetLink(['email' => $user->email], function (User $user, $token){
            $user->notify(new ResetPassword($token));
        });

        return Command::SUCCESS;
    }
    protected function getTenants()
    {
        return Tenant::all();
    }
}
```

The error is:
image

### Expected behavior

I expect to be able to use the password reset for more than one tenant consecutively 

### Laravel version

9.x

### stancl/tenancy version

3.6.0

megawubs:
Remove the 'auth.password' singleton from the container. This will ensure a new instance is created when another tenant needs it in the same app lifecycle.

Fixes #1052

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -32,6 +32,7 @@ return [
         Stancl\Tenancy\Bootstrappers\CacheTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\FilesystemTenancyBootstrapper::class,
         Stancl\Tenancy\Bootstrappers\QueueTenancyBootstrapper::class,
+        Stancl\Tenancy\Bootstrappers\AuthTenancyBootstrapper::class,
         // Stancl\Tenancy\Bootstrappers\RedisTenancyBootstrapper::class, // Note: phpredis is needed
     ],
 
--- a/src/Bootstrappers/AuthTenancyBootstrapper.php
+++ b/src/Bootstrappers/AuthTenancyBootstrapper.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Stancl\Tenancy\Bootstrappers;
+
+use Illuminate\Support\Facades\App;
+use Stancl\Tenancy\Contracts\TenancyBootstrapper;
+use Stancl\Tenancy\Contracts\Tenant;
+
+class AuthTenancyBootstrapper implements TenancyBootstrapper
+{
+
+    public function bootstrap(Tenant $tenant)
+    {
+        // empty
+    }
+
+    public function revert()
+    {
+        App::forgetInstance('auth.password');
+    }
+}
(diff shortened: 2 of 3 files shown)


stancl:
I'm still not sure what the issue is in #1052. The test here should be higher level, rather than just checking that the managers aren't equal, it should make real world assertions.

megawubs:
This is te underlying issue to what's described in #1052. The provided command now runs without any problem.

Also, disabling the new bootstrapper makes the test fail. Providing proof that the instance given back is the same as made inside the other tenant context. 

The problem here is quite low level, That's why the test is low level. 

larvel registers a singelton here:

https://github.com/laravel/framework/blob/df696e9963903172ea218bc97d00b72ce738c0ee/src/Illuminate/Auth/Passwords/PasswordResetServiceProvider.php#L27

Note that the above service provider is a deferred service provider. This means it is only executed when `auth.password` or `auth.password.broker` are resolved from the container. 

Because of the singleton, the `PasswordBrokerManager` returned on line 28 stays active inside the container. This is no big problem. Only thing is that inside the PasswordBrokerManager, it keeps a list of brokers that, once resolved, will be reused. These brokers get a reference to the ConnectionInterface. But, when tenancy ends, the PDO on the ConnectionInterface is set to `null`.  the PasswordBroker that was made once is kept in memory, with a reverence to the now incorrect ConnectionInterface. Removing the singleton from the container forces Laravel to re-evaluate it and make a correct connection instance to give to the `PasswordBroker`

stancl:
We couldn't reproduce this, so I'll close the issue and the PR. If the issue exists in your code for whatever reason, you can always use this bootstrapper in your own app.

megawubs:
> We couldn't reproduce this, so I'll close the issue and the PR. If the issue exists in your code for whatever reason, you can always use this bootstrapper in your own app.

That's interesting. Not even with the example command provided in the issue?

PR #1062, opened 2023-02-14
Feature/update resource in central database

ovidiu-serafim:


stancl:
1. Changes to resource syncing should be targeting `master`, not `3.x`, since resource syncing got significantly changed in v4
2. Provide a detailed description for a PR like this

PR #1068, opened 2023-02-17
[4.x] Set `app.url` config in UrlTenancyBootstrapper

lukinovec:
This PR expands https://github.com/archtechx/tenancy/pull/1044. Instead of updating just the URL generator's root URL, we also update the `app.url` config, so that the app URL override is also reflected there.


stancl:
Can you also add assertions for this?

lukinovec:
Just pushed them  (they're failing – I'll fix that)

PR #1076, opened 2023-02-23
Feature/update resource in central database

ovidiu-serafim:
This PR introduces the possibility of creating a resource in the central database related to a resource from the tenant database only with synced attributes if this resource doesn't exist.

### Solution

PR introduces the new config key, which is a boolean with a "false" default value.

### Usage

Set the value of the following config key to "true" for enabling this possibility.

`'created_only_with_synced_attributes' => true,`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,9 @@ env:
 
 on:
   push:
-    branches: [ 3.x, 2.x, master ]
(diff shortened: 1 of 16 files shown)


stancl:
I recommend opening a feature request issue describing what you're trying to achieve first, before opening a PR. As I mentioned, we changed resource syncing a lot in v4.

PR #1085, opened 2023-03-09
SqlServer Support

AmrIbrahem99:
Added SqlSrv support (including permission control)


stancl:
I think this was added for v4 in #715

stancl:
Hmm maybe we don't have the permission controlled version?

AmrIbrahem99:
yes ! I think so,  there is no SchemaManager too 
i mean i didnt work on that but i think its applicable

stancl:
What do you mean by "no SchemaManager"?

AmrIbrahem99:
like pg SchemaManager

AmrIbrahem99:
did you take a look on permission controller  yet?

AmrIbrahem99:
I think this pull request been dead for a while XD

> Hmm maybe we don't have the permission controlled version?

stancl:
We have reimplemented something similar in the v4 repo. Only adding the permission controlled one now, though it's a bit different and tests are included.

Appreciate the contribution but will have to close this in favor of the other PR.

PR #1091, opened 2023-03-18
Tenancy Fake method added to Tenancy Facade.

chinmaypurav:
The aim of this PR is to allow faking of tenancy when it comes to testing.

We can call the fake method on Tenancy facade to resolve the `TenancyFake` class in subsequent calls.

```php
use Stancl\Tenancy\Database\Models\Tenant;
use Stancl\Tenancy\Facades\Tenancy;

Tenancy::fake();

tenancy()->initialize('tenant_string');

echo tenant('id');
// 'tenant_string'



tenancy()->initialize(new Tenant(['id' => 'tenant_object']));

echo tenant('id');
// 'tenant_object'

```

### DIFF
--- a/src/Facades/Tenancy.php
+++ b/src/Facades/Tenancy.php
@@ -15,6 +15,8 @@ class Tenancy extends Facade
 
     public static function fake()
     {
-        static::swap(new TenancyFake);
+        static::swap($fake = new TenancyFake);
+
+        return $fake;
     }
 }
--- a/src/Facades/Tenancy.php
+++ b/src/Facades/Tenancy.php
@@ -12,4 +12,9 @@ class Tenancy extends Facade
     {
         return \Stancl\Tenancy\Tenancy::class;
     }
+
+    public static function fake()
+    {
+        static::swap(new TenancyFake);
+    }
 }
--- a/src/Facades/TenancyFake.php
+++ b/src/Facades/TenancyFake.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Stancl\Tenancy\Facades;
+
+use Stancl\Tenancy\Database\Models\Tenant;
+use Stancl\Tenancy\Tenancy;
+
+class TenancyFake extends Tenancy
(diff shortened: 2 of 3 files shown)


stancl:
Could you mention the use cases of this in depth in the PR description?

Also, have you actually tested this code in an app? There's a lot of things to consider when adding a fake for the Tenancy object.

And if we implement this, we probably don't want it just for the facade. The `tenancy()` helper is used more than the facade.

chinmaypurav:
1. I am working on a client project, which at times demand certain checks depending on a specific tenant. The fake method allows me to set the tenancy instance to anything I want without the need for a central DB. I have recently started to write tests for all the new features I am working upon, So earlier I was using the methodology given in the V2 docs, I create a central DB, a tenant, I run the migrations and then I test. That approach was costing me too much time. So I took this approach to solve my use case.
2.  I have already implemented the code in my project by creating a custom facade class extending to the package facade class.
3. The tenancy helper and the Facade resolve the same object (look screenshot)
image
Since we are using service containers, once we use the fake method on the facade, the subsequent calls via Facade or the `tenancy()` helper will return the `TenancyFake`  instance instead of the `Stancl\Tenancy\Tenancy` instance. I reverse engineered how the `fake()` method was used on other Laravel facades.

Additional side notes:
~I was not able to figure out how to write tests for the fake method (I can figure it out given some time), but for now the code is working for me in my client project.~
Update: Basic Tests added. Can discuss if we need to cover more scenarios.

chinmaypurav:
@stancl Any further thoughts for this PR?

stancl:
I don't really see much point to adding this. Like I said, actually implementing the fake would be a lot more work, we'd need the fake to work with the service provider too (and that's more complicated because the Tenancy object is injected in many places), we'd need some assertion methods on the fake to make it more valuable. And still I don't think this solves much, the package as it is is very testable and there are many approaches to improving test performance if that's a concern.

PR #1094, opened 2023-03-24
Refactor checking host is subdomain

suleymanozev:
When use endsWith throw 500 server error with is not subdomain host is not subdomain.

I check subdomain really exists.


stancl:
The PR description should explain why this change is beneficial to the package.

suleymanozev:
Because, wrong way checking host is a subdomain.

suleymanozev:
Please review code. My english is bad.

stancl:
The code change looks like it goes from checking that the **hostname ends with a central domain** to simply checking that **the central domain is included in the hostname anywhere**?

I don't understand why we would make this change.

PR #1105, opened 2023-04-21
[4.x] Postgres RLS support

### LINKED ISSUE #769: Add support for Postgres RLS
https://github.com/archtechx/tenancy/issues/654#issuecomment-848209019

lukinovec:
(Note before merging: sync the Post/Comment/ScopedComment changes – use the dedicated model files.)

**Implementation**
To make Postgres RLS work with Tenancy, we create and use policies for scoping the queries performed on the tables of tenant-related models using the CreateRLSPoliciesForTenantTables command. The models are discovered in directories configured in `tenancy.rls.model_directories` (+ their subdirectories). Models are considered tenant-related if they belong to tenant directly (= they're using the `BelongsToTenant` trait), or indirectly (= they're using the `BelongsToPrimaryModel`).

To scope the queries by the current tenant using policies, the tenants also need Postgres users. We create them using the CreatePostgresUserForTenant job – the job also gives all the permissions to the user, but by changing the `tenancy.rls.user_permissions` config, you can choose the exact permissions you want to grant. Postgres users have the ID of the tenants they belong to as their names by default, e.g. a tenant with ID 'acme' will have a Postgres user named 'acme', or you can use a custom username by setting the internal `db_username` property on tenants (`tenancy_db_username` by default). The policies check if the tenant foreign key of the queried records (or their parent/primary models) is the same as the name of the current Postgres user.

The DeleteTenantsPostgresRole job deletes the tenant's Postgres user along with the user's permissions (Postgres user can't get deleted if he still has related objects, e.g. permissions).

**Usage**
> Note: Your app has to use **single-database** tenancy

First, enable PostgresRLSBootstrapper. Then, uncomment [the jobs](https://github.com/archtechx/tenancy/compare/master...add-postgres-rls-support#diff-6ca30424f3bab5bd30feedc72b4d4e930244bc8f73fc6828396beb574df18fd0R33-R59) in TenancyServiceProvider. After that, specify the directories of your models in the Tenancy config (`tenancy.rls.model_directories`) and make sure the DB is migrated. Finally, run `php artisan tenants:create-rls-policies`.

> Note that `tenants:create-rls-policies` only creates policies for tables that are discovered in the configured directories and belong to tenants (directly, or through a primary model).

From now on, when tenants get created, we also create a Postgres user for them. You can use the `tenants:create-postgres-user` command to create Postgres users for the existing tenants.

To scope models using RLS, either:
- set `config('tenancy.rls.enabled')` to `true` – queries of all tenant-related models will be scoped using RLS
- make models that belong to tenants implement the RlsModel interface – queries of all models that use the BelongsToTenant **and implement the RlsModel interface** will get scoped using RLS


.github/workflows/ci.yml
```diff
@@ -47,8 +47,8 @@ jobs:
       postgres:
         image: postgres:latest
         env:
-          POSTGRES_PASSWORD: password
-          POSTGRES_USER: root
+          POSTGRES_PASSWORD: 'postgres'
+          POSTGRES_USER: 'postgres'
```
stancl: Are these changes still needed?

docker-compose.yml
```diff
@@ -52,16 +52,26 @@ services:
...
-            interval: 10s
+            test:
+                - CMD
+                - pg_isready
+                - '-q'
+                - '-d'
+                - '${DB_DATABASE}'
+                - '-U'
+                - '${DB_USERNAME}'
+            retries: 3
```
stancl: Can you provide a basic rundown of what these changes do/why they're needed

tests/Etc/Post.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Stancl\Tenancy\Tests\Etc;
+
+use Illuminate\Database\Eloquent\Model;
+use Stancl\Tenancy\Database\Concerns\BelongsToTenant;
+
+class Post extends Model
```
stancl, line 12: We already have the Post & Comment models in https://github.com/tenancy-for-laravel/v4/pull/3/files.

Are these models different in a way that's *specific to RLS*? If yes, they should be in some more specific folder than just `tests/Etc`. If not, then add a note to change this once the PR with early identification changes is merged (use the same model in tests of both of the features).

tests/Etc/ScopedComment.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Stancl\Tenancy\Tests\Etc;
+
+use Stancl\Tenancy\Database\Concerns\BelongsToPrimaryModel;
+
+class ScopedComment extends Comment
+{
+    use BelongsToPrimaryModel;
```
stancl, line 12: This class uses single-db scoping logic? If so, then probably add comments to these models so that it's clear how the setup works.

tests/Etc/Post.php
lukinovec: > We already have the Post & Comment models in https://github.com/tenancy-for-laravel/v4/pull/3/files.

We _had_ these – I moved these to the separate Post & Comment files

> Are these models different in a way that's specific to RLS? 

The models aren't different in that regard. Some are intended to be used with the single DB approach – Post uses BelongsToTenant and ScopedComment uses BelongsToPrimaryModel, so they're specific to the single DB approach. The Comment model isn't specific to any approach though. So that's a thing I'll need to make clear

> add a note to change this once the PR with early identification changes is merged (use the same model in tests of both of the features)

I'll add a note to sync the changes – in the early id PR, we have the separate Post & Comment files, but the ScopedX classes are defined only directly in the SingleDatabaseTenancyTest file. So in the early id PR, we'll need to use the separate ScopedX files.

tests/Etc/ScopedComment.php
lukinovec: Yeah (answered [here](https://github.com/archtechx/tenancy/pull/1105#discussion_r1200367668))

> add comments to these models so that it's clear how the setup works

Right

.github/workflows/ci.yml
lukinovec: https://github.com/archtechx/tenancy/actions/runs/5044637675/jobs/9047926168#step:2:357

```
Error: Database is uninitialized and superuser password is not specified.
          You must specify POSTGRES_PASSWORD to a non-empty value for the
          superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".
```

Looks like we need at least the password

docker-compose.yml
lukinovec: I completely reverted the healthcheck changes and things seem to be working fine. So in the end, these changes aren't needed

.github/workflows/ci.yml
lukinovec: We shouldn't need these changes, but it seems that in the tests, we're attempting to log in as the "postgres" user regardless of the user specified here. I'm trying to figure out why right now

lukinovec: So the problem was [here](https://github.com/archtechx/tenancy/pull/1105/files#diff-abf1fa77a19f720052b75bc802df0260157b9064d8c3c12a80794f01e9acc065R94-R95) – I hardcoded the credentials. I updated the credentials, things should work OK without this `ci.yml` change. I'm not sure if this env variable setup is correct in the end (e.g. using POSTGRES_USER x DB_USERNAME), I'll take another look at that.

tests/Etc/ScopedComment.php
lukinovec: https://github.com/archtechx/tenancy/pull/1105/commits/a46b26800f3ee7c8960dc6f23770af2d1ac8b1a9

.github/workflows/ci.yml
lukinovec: >  I'm not sure if this env variable setup is correct in the end (e.g. using POSTGRES_USER x DB_USERNAME)

I decided not to use `DB_USERNAME` as the default for `database.connections.pgsql.username` – I just used the default credentials from `docker-compose.yml`. This will work both with CI and with user's custom credentials set in .env (`POSTGRES_USER`, `POSTGRES_PASSWORD`)

lukinovec: So in the end, these changes aren't needed – I'm resolving this

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->originalCentralConnectionConfig = $config->get('database.connections.central');
+        $this->originalPostgresConfig = $config->get('database.connections.pgsql');
+    }
+
+    public function bootstrap(Tenant $tenant): void
+    {
+        /** @var TenantWithDatabase $tenant */
+        $this->database->purge('central');
+
+        $this->config->set('database.connections.pgsql.username', $tenant->getTenantKey());
```
stancl: Why `getTenantKey()` instead of `$tenant->database()->getUsername()`?

stancl: These things should be in line with the existing logic we have (eg how we do permission controlled DB managers).

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Stancl\Tenancy\Bootstrappers\Integrations;
+
+use Illuminate\Contracts\Config\Repository;
+use Illuminate\Database\DatabaseManager;
+use Stancl\Tenancy\Contracts\TenancyBootstrapper;
+use Stancl\Tenancy\Contracts\Tenant;
+use Stancl\Tenancy\Database\Contracts\TenantWithDatabase;
+
+class PostgresTenancyBootstrapper implements TenancyBootstrapper
```
stancl: Is this used instead of `DatabaseTenancyBootstrapper` or alongside it?

src/Commands/CreatePostgresUserForTenants.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Illuminate\Console\Command;
+use Stancl\Tenancy\Concerns\HasTenantOptions;
+use Stancl\Tenancy\Jobs\CreatePostgresUserForTenant;
+
+class CreatePostgresUserForTenants extends Command
+{
+    use HasTenantOptions;
+
+    protected $signature = 'tenants:create-postgres-user {--tenants=* : The tenant(s) to run the command for. Default: all}';
```
stancl, line 15: Not sure about using singular/plural for user here since if this can be/is meant to be executed for multiple tenants, using `users` might make more sense.

src/Commands/CreateRLSPoliciesForTenantTables.php
```diff
@@ -0,0 +1,73 @@
...
+                    $this->components->info("Table '$table' is not related to tenant. Make sure $modelName uses the BelongsToPrimaryModel trait.");
+                }
+            } else {
+                DB::statement("CREATE POLICY {$table}_rls_policy ON {$table} USING ({$tenantKey}::TEXT = current_user);");
+
+                $this->makeTableUseRls($table);
+
+                $this->components->info("Created RLS policy for table '$table'");
+            }
+        }
```
stancl: Probably extract parts of this to protected methods. The if statement is too long and has a lot of nesting/layers

src/Jobs/CreatePostgresUserForTenant.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $name = $this->tenant->getTenantKey();
```
stancl: Same as the other review. Any reason for using the tenant key instead of the username method that is used for permission controlled DB managers?

src/Jobs/DeleteTenantsPostgresUser.php
```diff
@@ -0,0 +1,46 @@
...
+    public function handle()
+    {
+        $name = $this->tenant->getTenantKey();
+
+        // Revoke all permissions of a Postgres user before dropping it
+        // Skip dropping permissions if the user doesn't exist
+        if (count(DB::select("SELECT usename FROM pg_user WHERE usename = '$name';")) > 0) {
+            DB::statement("DROP OWNED BY \"{$name}\";");
+            DB::statement("DROP USER \"{$name}\";");
+        }
```
stancl: Can we do things like this in a transaction? If yes, then we should probably use transactions. Same in other parts of the PR.

tests/Etc/Comment.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Stancl\Tenancy\Tests\Etc;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class Comment extends Model
```
stancl, line 11: Do models (`Comment`) that belong to models (`Post`) that belong to a tenant not need any traits?

stancl: Ah I see that this is what scoped comments do. Probably add comments explaining the difference then, since from `Post` it's not clear that it's not scoped.

There were no comments when the classes were just part of the single DB tenancy test, but they will be helpful if the classes are more spread out across the codebase

tests/TestCase.php
```diff
@@ -85,6 +85,20 @@ protected function getEnvironmentSetUp($app)
                     PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
                 ]) : [],
             ],
+            'database.connections.pgsql' => [
```
stancl: Did we not have a pgsql connection before? Or are we setting something differently here

.github/workflows/ci.yml
stancl: When you leave a lot of comments on a review you can keep it unresolved so that I see the thread and can read it.

The file is still changed, so now the question is if the new changes are needed/what are they needed for.

lukinovec: > The file is still changed

It isn't, I reverted the changes (check [the files tab](https://github.com/archtechx/tenancy/pull/1105/files))

tests/TestCase.php
lukinovec: We had a pgsql connection before, and now, I see we don't really need to set this here – the config is getting set in the PostgresTenancyBootstrapper, so I don't think there's a reason to keep this.

I deleted the config setting

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
lukinovec: I see, `$tenant->database()->getUsername()` makes more sense. Changed`$tenant->getTenantKey()` to `$tenant->database()->getUsername() ?? $tenant->getTenantKey()`. So we only default to the tenant key if the db username is null

src/Jobs/CreatePostgresUserForTenant.php
lukinovec: Responded to the other review https://github.com/archtechx/tenancy/pull/1105#discussion_r1203823368

src/Commands/CreatePostgresUserForTenants.php
lukinovec: I think `user` might be better because we intend to create one user for each of the tenants. But we'll most likely end up creating multiple different users

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
lukinovec: Instead, since we only use `DatabaseTenancyBootstrapper` with multi-db tenancy. I'll update the bootstrapper's docblock to make this clear

lukinovec: https://github.com/archtechx/tenancy/pull/1105/commits/0991f62aa5cab0acd250abe2d29ceff3f62142ab

src/Jobs/DeleteTenantsPostgresUser.php
lukinovec: I moved the DB statements into transactions

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: I extracted lines 29-46 to a separate method, and that cleaned up the code a bit – I think I can further reduce the nesting, so this is still WIP

lukinovec: The code looks like this now: https://github.com/archtechx/tenancy/pull/1105/files#diff-964269653c823a5110f745ba0a2f386575f946f6ce35152b89c741ca367859cf

tests/Etc/Comment.php
lukinovec: I made [some small changes](https://github.com/archtechx/tenancy/pull/1105/commits/1f9a344f33e158edf6aa3283d2e81c7670d30339). I think the difference is clearer now

src/Commands/CreatePostgresUserForTenants.php
stancl: > because we intend to create one user for each of the tenants

Ah, that'd make sense.

> But we'll most likely end up creating multiple different users

What do you mean by this?

tests/Etc/Comment.php
stancl: > This model is not intended to be used with the single-database tenancy approach

Rather than mentioning that something is not intended to be used with a certain thing, comments should explain what the thing *is* used for.

src/Commands/CreateRLSPoliciesForTenantTables.php
```diff
@@ -0,0 +1,87 @@
...
+            // Table is not directly related to tenant
+            if (in_array(BelongsToPrimaryModel::class, class_uses_recursive($model::class))) {
+                $this->makeSecondaryModelUseRls($model);
+            } else {
+                $modelName = $model::class;
+
+                $this->components->info("Table '$table' is not related to tenant. Make sure $modelName uses the BelongsToPrimaryModel trait.");
+            }
+        } else {
+            DB::transaction(fn () => DB::statement("CREATE POLICY {$table}_rls_policy ON {$table} USING ({$tenantKey}::TEXT = current_user);"));
```
stancl: What's the point of wrapping a single DB call in a transaction?

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
stancl: Think we should rename the class, right now it sounds like something that people could enable when they're using postgres. To be clear: this is ONLY used when using RLS, right?

If so, maybe something like `PostgresRLSBootstrapper` could work better (or anything similar to that), so that it's explicitly in the name.

```
/**
 * Purge central connection, set Postgres credentials to the tenant's credentials
 * and use Postgres as the central connection.
 *
 * This bootstrapper is intended to be used with single-database tenancy.
 */
```

Is RLS "triggered" by using the user (that it's configured/enforced for) in the DB connection?

.github/workflows/ci.yml
stancl: CleanShot 2023-05-29 at 8 33 51@2x

I was referring to the files tab.

src/Commands/CreatePostgresUserForTenants.php
lukinovec: I mean we create user "foo" for tenant "foo", user "bar" for tenant "bar" and so on, so `users` might be good too. I still like `user` more

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: I thought that if the call was wrapped in a transaction, and the call failed, the DB would be rolled back. That seemed like the correct thing to do – if the call would've broken something, the rollback would fix that

tests/Etc/Comment.php
lukinovec: Yeah, that makes sense. I'll update that

lukinovec: https://github.com/archtechx/tenancy/pull/1105/commits/28efc0ca97e41e275a9e4d49ac304e8c06d86127 updated the docblocks

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
lukinovec: > To be clear: this is ONLY used when using RLS, right?

When using the single-database approach – no matter if RLS is used

> Is RLS "triggered" by using the user (that it's configured/enforced for) in the DB connection?

RLS is enabled (or triggered, I think) for specific tables by creating the policies using CreateRLSPoliciesForTenantTables. The user in the DB connection is used for checking the RLS constraint (checking if the record's `tenant_id`/column name returned by `tenancy()->tenantKeyColumn()` matches the ID of the current user – maybe we could make the column name configurable, since we made the username more customizable using `$tenant->database()->getUsername() ?? $this->tenant->getTenantKey()`)

I thought about changing the bootstrapper's name too, but a name suggesting that RLS is the only thing this is used for doesn't seem right – the most correct name would be something like PostgresSingleDatabaseBootstrapper.

.github/workflows/ci.yml
lukinovec: For the record, I wrote about the `docker-compose.yml` file on Discord

> docker-compose.yml changes:
> added comment
> instead of hardcoding the `postgres` user in the healthcheck, we use the user specified in the `environment` section (`${POSTGRES_USER}` – that's `root`)

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: I deleted the single statement transactions, and I put some method calls that execute multiple statements into transactions

https://github.com/archtechx/tenancy/compare/8ccc27e8c3d5c2ad106dcbb77321a343862e456e..61519ecec34c41efc7b9e72f782da353204ba0cf

src/Commands/CreatePostgresUserForTenants.php
stancl: Right, but it's user per tenant, so singular makes sense.

src/Bootstrappers/Integrations/PostgresTenancyBootstrapper.php
stancl: (Conversation continued on Discord)

lukinovec: For the record, I've renamed the bootstrapper to PostgresRLSBootstrapper, updated the annotation and simplified the bootstrapper a bit

https://github.com/archtechx/tenancy/pull/1105/commits/12b36f407d9ae557bb648885b03225a5c782e9fa

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->originalPostgresConfig = $config->get('database.connections.pgsql');
+    }
+
+    public function bootstrap(Tenant $tenant): void
+    {
+        /** @var TenantWithDatabase $tenant */
+        $this->config->set('database.connections.pgsql.username', $tenant->database()->getUsername() ?? $tenant->getTenantKey());
+        $this->config->set('database.connections.pgsql.password', $tenant->database()->getPassword() ?? 'password');
+
+        $this->config->set('database.connections.central', $this->config->get('database.connections.pgsql'));
```
stancl: What's the point of setting `central` specifically?

The name of the central connection is configurable: https://github.com/archtechx/tenancy/blob/bd9bbe8b41822f7d157cdb4843c4adc0e7404804/assets/config.php#L135

Same with the `purge('central')` call. The central connection can have a custom name. It's not always `central`.

lukinovec: Right, I missed that the name's configurable

lukinovec: https://github.com/archtechx/tenancy/pull/1105/commits/5754f8bdc4226d72fdfcc181f346ae19ed3ae431

.github/workflows/ci.yml
lukinovec: All `ci.yml` and `docker-compose.yml` changes are reverted now

assets/config.php
```diff
@@ -178,6 +177,37 @@
...
+        'enabled' => false,
+
+        /**
+         * Permissions to grant to the tenant Postgres users.
+         *
+         * By default, all permissions are granted.
+         *
+         * @see Stancl\Tenancy\Jobs\CreatePostgresUserForTenant
+         */
+        'user_permissions' => ['ALL'],
```
stancl: Just to double check: Do these permissions only apply on the tenant's schema, or will the created user have these permissions across the whole postgres server?

If it's the latter, then we might want to use something more restrictive by default.

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    protected array $originalCentralConnectionConfig;
+    protected array $originalPostgresConfig;
+
+    public function __construct(
+        protected Repository $config,
+        protected DatabaseManager $database,
+    ) {
+        $this->originalCentralConnectionConfig = $config->get('database.connections.central');
+        $this->originalPostgresConfig = $config->get('database.connections.pgsql');
```
stancl: This was added here since the last review, right? Is the point of it to make sure that the `central` connection will be based on the `pgsql` "preset" even if it'd be e.g. a mysql config by default?

I'm not 100% sure if that's what we want. If we do go with this approach, we should make the connection name configurable I think since `pgsql` is just what the default Laravel skeleton uses, but it's not necessarily what it's always called I think.

But more importantly I'm not sure if this is even needed/makes sense. What'd be the use case for having a central connection that's **not** postgres if we want to use Postgres RLS? The point of the feature is to in a sense add automatic scoping to a **single-database postgres-based** multi-tenant app. So I don't see a case where the central connection would not be postgres.

stancl: This review relates to the other changes here as well, especially lines 42-45.

src/Commands/CreateRLSPoliciesForTenantTables.php
```diff
@@ -0,0 +1,81 @@
...
+            }
+        });
+
+        return Command::SUCCESS;
+    }
+
+    /**
+     * Make model use RLS if it belongs to a tenant directly, or through a parent primary model.
+     */
+    protected function useRlsOnModel(Model $model): void
```
stancl: Maybe this method could have a more accurate name, something like "apply" or some other word that implies the persistent change we're making to the DB. "Use" sounds like it could be something temporary/something within the lifecycle of a single request.

Also, a small nitpick: we should probably use uppercase `RLS` consistently across class names and methods. Unless you think there's a good reason to use different casing for class names and methods.

src/Commands/CreateRLSPoliciesForTenantTables.php
```diff
@@ -0,0 +1,81 @@
...
+            $parentTable = $model->$parentName()->make()->getTable();
+
+            DB::statement("CREATE POLICY {$table}_rls_policy ON {$table} USING (
+                {$parentKey} IN (
+                    SELECT id
+                    FROM {$parentTable}
+                    WHERE ({$tenantKey} = (
+                        SELECT {$tenantKey}
+                        FROM {$parentTable}
+                        WHERE id = {$parentKey}
```
stancl: Could we use query parameters (`?`) here instead of directly creating a SQL string in PHP and executing it in the database?

~~`$parentKey` *might be* user data I think, so this might potentially open up the application to SQL injection.~~ Actually it's not the *key*, it's the *key name*. So we should probably rename `$tenantKey` and `$parentKey` to `$tenantKeyName` and `$parentKeyName`. But still, using parameters would be much better than what we're doing here if it'd be possible.

There's a lot going on here (compared to eg the DB managers), so even if it'd be secure now, there's a risk some user data might slip in in the future without being noticed.

Ideally we'd change all of these queries to use parameters (including the ones in all DB managers, `CreatePostgresUserForTenant` and all other places) if possible.

src/Jobs/CreatePostgresUserForTenant.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $name = $this->tenant->database()->getUsername() ?? $this->tenant->getTenantKey();
+        $password = $this->tenant->database()->getPassword() ?? 'password';
```
stancl: Is `password` the default password if there's no password available in `getPassword()`?

Maybe a blank password would be more standard, that's the case with MySQL, not sure about pgsql.

tests/Etc/EarlyIdentification/Service.php
```diff
@@ -10,6 +10,6 @@ class Service
 
     public function __construct()
     {
-        $this->token = config('tenancy.token');
+        $this->token = config('tenancy.token', '');
```
stancl: What's this change for? Specifically within this PR.

src/Database/Concerns/DealsWithModels.php
```diff
@@ -0,0 +1,63 @@
...
+            return null;
+        }, iterator_to_array($modelFiles)), fn (object|null $class) => $class instanceof Model);
+    }
+
+    /**
+     * Filter all models retrieved by static::getModels() to get only the models that belong to tenants.
+     */
+    public static function getTenantModels(): array
+    {
+        return array_filter(static::getModels(), fn (Model $model) => tenancy()->modelBelongsToTenant($model) || tenancy()->modelBelongsToTenantIndirectly($model));
```
stancl:
```suggestion
        return array_filter(static::getModels(), fn (Model $model) => static::modelBelongsToTenant($model) || static::modelBelongsToTenantIndirectly($model));
```
We should be using `static::` within this file for these calls, not `tenancy()->`.

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: Yeah I thought about the casing too – "Rls" seemed more appropriate, but we can make it uppercase everywhere,

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: Okay, I'll look into that.

src/Jobs/CreatePostgresUserForTenant.php
lukinovec: I see, the default password should be `postgres` 
edit: `''`

src/Commands/CreateRLSPoliciesForTenantTables.php
stancl: Maybe keep this review on hold for now and I'll revisit it when the other things in the code are resolved.

tests/Etc/EarlyIdentification/Service.php
lukinovec: the service's `token` property had to have a string value, so the service relied on `config('tenancy.token')` to be set. 

I think this isn't needed now – the model discovery discovered all instantiable classes in the tests/Etc directory, and that logic is a bit different now.

edit: this was correct – reverted the change and things work fine.

src/Jobs/CreatePostgresUserForTenant.php
lukinovec: Changed this entirely – since we can't use empty string passwords with Postgres, I added a static property to the RLS bootstrapper

```php
/** Mustn't return an empty string */
public static Closure|string $defaultPassword = 'password';
```

This is used both for the Postgres connection password config, and the newly created user.

The bootstrapper now has a public static method so that it can be used conveniently in both the bootstrapper and the user creation job.

```php
public static function getDefaultPassword(): string
{
    return is_string(static::$defaultPassword) ? static::$defaultPassword : (static::$defaultPassword)();
}
```

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
lukinovec: > since pgsql is just what the default Laravel skeleton uses, but it's not necessarily what it's always called I think.

I think this is true. We should also probably add a static property for configuring the postgres connection name (default would be `pgsql`)

lukinovec: > What'd be the use case for having a central connection that's not postgres if we want to use Postgres RLS? The point of the feature is to in a sense add automatic scoping to a single-database postgres-based multi-tenant app. So I don't see a case where the central connection would not be postgres.

I think I did this essentially to keep the config the same except the username and password.  I'm going to look into a better way of doing this, but importantly, it was because of the credentials

lukinovec: Yeah, no need to touch or configure the postgres connection config at all (only the central one). So the static property isn't needed. I made some changes so that the bootstrapper's cleaner.

https://github.com/archtechx/tenancy/pull/1105/commits/a568aa36264ee06ad0385094a5c3a57d2d8eb843

assets/config.php
lukinovec: Looks like they apply on the tenant's schema only https://github.com/archtechx/tenancy/commit/d005e7b06dbfdd4b81192f0dfd51faabb7ce7097

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: I renamed the variables, and looked into supplying the query parameters extensively. I haven't found a way to do that with Postgres queries though.

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
lukinovec: Updated this so that the pgsql connection isn't touched
https://github.com/archtechx/tenancy/pull/1105/commits/a568aa36264ee06ad0385094a5c3a57d2d8eb843

src/Jobs/CreatePostgresUserForTenant.php
lukinovec: Improved this so that the default password is configurable https://github.com/archtechx/tenancy/pull/1105/commits/0fa52e502027743d9181d03a5031f88ac0dc6476

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: I figured out how to use the parameters in the Postgres statements:

https://github.com/archtechx/tenancy/pull/1105/commits/855ebfbd0b835ef322f95305f53174fc026d35d2
https://github.com/archtechx/tenancy/pull/1105/commits/15cdd6b2c1fe042f147cca2beae86fb076f5896d

assets/config.php
lukinovec: Also, updated the default permissions to `'UPDATE', 'DELETE', 'SELECT', 'INSERT'`, which is a better default since it only allows the basic operations

stancl: https://www.postgresql.org/docs/current/ddl-priv.html

I'm thinking that some of these privileges look like something we'd want:

CleanShot 2023-06-29 at 7 33 17@2x

However, they might actually not be needed because the user is only used *sometimes*.

For instance:

> `REFERENCES` — Allows creation of a foreign key constraint referencing a table, or specific column(s) of a table.

This is something that'd be done in migrations **which are executed by the main user**, not the tenant user.

`TRUNCATE` is also something that'd probably be done by the central user. There *might* be use cases in the tenant app (e.g. to delete the contents of some table and also reset indexes, for data that doesn't need to be permanent e.g. some types of "cache" or custom sessions etc).

As for `EXECUTE`, `USAGE`, and `TEMPORARY`, those might have uses, I'm not 100% sure. Requires a bit more research.

What's the reason for this one not being needed?

> `CONNECT` — Allows the grantee to connect to the database. This privilege is checked at connection startup

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * This bootstrapper is intended to be used with Postgres RLS (single-database tenancy).
+ *
+ * @see \Stancl\Tenancy\Commands\CreatePostgresUserForTenants
+ * @see \Stancl\Tenancy\Commands\CreateRLSPoliciesForTenantTables
+ */
+class PostgresRLSBootstrapper implements TenancyBootstrapper
+{
+    protected array $originalCentralConnectionConfig;
+    protected array $originalPostgresConfig;
```
stancl: This property isn't used now

assets/config.php
lukinovec: > What's the reason for this one not being needed?

Because I couldn't figure this one out myself, I asked ChatGPT (the prompt was "A Postgres DB user I created doesn't have the 'CONNECT' privilege, though it can still connect to the Postgres database. Why exactly does this happen?")

The reply:

> In PostgreSQL, the ability to connect to a database is controlled by two main factors: the server configuration and the database user privileges.
> 
> By default, PostgreSQL allows all users to connect to a database, even if they don't have explicit privileges granted. This behavior is determined by the server configuration parameter `pg_hba.conf` (Host-Based Authentication Configuration). This file defines the authentication rules that the PostgreSQL server uses to determine whether a client can connect.
> 
> When a client attempts to connect, the server consults the `pg_hba.conf` file to check if the client's IP address and authentication method match any of the rules. If a match is found, the client is allowed to connect, regardless of the specific user privileges within the database.
> 
> Once the client has successfully connected to the database, the user's privileges within the database itself determine what actions they can perform. The privileges control access to tables, views, functions, and other database objects.
> 
> Therefore, even if a database user doesn't have the 'CONNECT' privilege explicitly granted, they can still connect to the database as long as the server's authentication rules in `pg_hba.conf` allow it. However, the user's actions within the database will be restricted based on their assigned privileges.

lukinovec: So in conclusion, by default, the CONNECT privilege is granted to new Postgres users by default depending on the user's Postgres config (and other commonly granted privileges; for convenience).

We can either strip the newly created Postgres user of all its privileges and give it only the privileges configured in our package's config, or let things as they are now. The first option is definitely the more secure one, since the user would have complete control over the permissions of the tenant Postgres users.

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
lukinovec: Right, deleted.

lukinovec: https://github.com/archtechx/tenancy/pull/1105/commits/90550344511fe57363c3c4d8535cfe448bb2ed0f

src/Jobs/CreatePostgresUserForTenant.php
```diff
@@ -0,0 +1,72 @@
...
+            foreach ($tenantModels as $model) {
+                $table = $model->getTable();
+
+                foreach (config('tenancy.rls.user_permissions') as $permission) {
+                    $formattedStatement = $databaseManager->database()->select("SELECT format('GRANT %s ON %I TO %I', '$permission', '$table', '$userName')")[0]->format;
+                    $databaseManager->database()->statement($formattedStatement);
+                }
+
+                $formattedStatement = $databaseManager->database()->select("SELECT format('GRANT USAGE ON ALL SEQUENCES IN SCHEMA public TO %I', '$userName')")[0]->format;
+                $databaseManager->database()->statement($formattedStatement);
```
arinc0: Hello,

This feature is fantastic and I'm eagerly looking forward to seeing it merged and released.

While reviewing the code, I noticed that the public schema is hard-coded in this statement.
This could potentially cause issues in projects that use a different schema.
 In Laravel, it's possible to specify the schema for PostgreSQL using the search_path in the database configuration.

To make the code more flexible and reusable across different environments, I propose dynamically retrieving the schema from the configuration, or allowing it to be set in a configuration file like tenancy.php.

The code could be modified as follows:

```php
$schema = config('database.connections.pgsql.search_path');
$formattedStatement = $databaseManager->database()->select("SELECT format('GRANT USAGE ON ALL SEQUENCES IN SCHEMA %I TO %I', '$schema', '$userName')")[0]->format;
$databaseManager->database()->statement($formattedStatement);
```

Please note that this change should be thoroughly tested to ensure that it doesn't break existing functionality or affect other parts of the code.

I look forward to your feedback on this proposal.

assets/config.php
lukinovec: > the CONNECT privilege is granted to new Postgres users by default depending on the user's Postgres config

Revisting this now, and this doesn't seem to be true. The ability to connect to the db itself depends on the config – no privilege is granted

src/Jobs/CreatePostgresUserForTenant.php
lukinovec: Added this, thank you for the suggestion!

assets/config.php
lukinovec: With the default Postgres configuration, the Postgres users don't need the `CONNECT` privilege to connect to the db. The privilege is only needed if you're trying to connect to the db from a non-trusted IP, and by default, the `pg_hba.conf` allows all loopback connections.

So the current default privileges are probably good enough – if someone uses a different config, he should revise the privileges, and customize them if needed.

src/Commands/CreateRLSPoliciesForTenantTables.php
lukinovec: Used "apply" instead of "use" in the method name and changed all "Rls" instances to "RLS"

src/Jobs/CreatePostgresUserForTenant.php
stancl: Mention this in the pr desc, so that people make sure to either have users that have passwords specified, or use a secure default password.

Alternatively, since this feature isn't really about separating users for security reasons, but rather for *convenience* (automatic scoping), wouldn't it make sense to use the DB password that's being used by the central DB?

lukinovec: > since this feature isn't really about separating users for security reasons, but rather for convenience (automatic scoping), wouldn't it make sense to use the DB password that's being used by the central DB

I think this makes sense. I'll update the code (updated https://github.com/archtechx/tenancy/pull/1105/commits/5a1beb1f3b993000748c4567152da4b4ce277c42)

stancl: > Mention this in the pr desc, so that people make sure to either have users that have passwords specified, or use a secure default password.

Is this done as well? Just mention that no password defaults to the one used for the central connection.

src/Bootstrappers/Integrations/PostgresRLSBootstrapper.php
```diff
@@ -0,0 +1,54 @@
...
+    public function bootstrap(Tenant $tenant): void
+    {
+        $centralConnection = $this->config->get('tenancy.database.central_connection');
+        $centralConnectionPassword = $this->config->get("database.connections.$centralConnection.password");
+
+        $this->database->purge($centralConnection);
+
+        /** @var TenantWithDatabase $tenant */
+        $this->config->set([
+            'database.connections.' . $centralConnection . '.username' => $tenant->database()->getUsername() ?? $tenant->getTenantKey(),
```
stancl, line 41: What happens when `?? $tenant->getTenantKey()` is used?

src/Database/Concerns/BelongsToTenant.php
```diff
@@ -21,7 +21,11 @@ public function tenant(): BelongsTo
 
     public static function bootBelongsToTenant(): void
     {
-        static::addGlobalScope(new TenantScope);
+        // If 'tenancy.rls.enabled' is true or this model implements RLSModel
+        // Scope queries using Postgres RLS instead of TenantScope
```
stancl:
```suggestion
        // If 'tenancy.rls.enabled' is true or this model implements RLSModel
        // Postgres RLS is used for scoping, so we don't enable the scope used with single-database tenancy.
```
Also just to confirm — this means that if `tenancy.rls.enabled` is used, individual models don't need to implement `RLSModel`, and RLS is used as the default "strategy" for scoping models that use this trait, right?

src/Database/Concerns/DealsWithModels.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public static function getModels(): array
+    {
+        if (static::$modelDiscoveryOverride) {
+            return (static::$modelDiscoveryOverride)();
+        }
+
+        $modelFiles = Finder::create()->files()->name('*.php')->in(config('tenancy.rls.model_directories'));
+
+        // todo1 Add array property for excluding specific models
```
stancl: Resolve priority todos in this PR

src/Database/Concerns/DealsWithModels.php
```diff
@@ -0,0 +1,66 @@
...
+        }, iterator_to_array($modelFiles)), fn (object|null $object) => $object instanceof Model && ! in_array($object::class, static::$modelsExcludedFromDiscovery));
+    }
+
+    /**
+     * Filter all models retrieved by static::getModels() to get only the models that belong to tenants.
+     */
+    public static function getTenantModels(): array
+    {
+        return array_filter(static::getModels(), fn (Model $model) => static::modelBelongsToTenant($model) || static::modelBelongsToTenantIndirectly($model));
+    }
```
stancl: Is this used only with RLS? If so, should it also check for whether models implement the interface RLS interface/whether RLS is used globally in config?

src/Database/Concerns/DealsWithModels.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public static function modelBelongsToTenant(Model $model): bool
+    {
+        return in_array(BelongsToTenant::class, class_uses_recursive($model::class));
+    }
+
+    public static function modelBelongsToTenantIndirectly(Model $model): bool
+    {
+        return in_array(BelongsToPrimaryModel::class, class_uses_recursive($model::class));
+    }
```
stancl: If these are only used within this class, they can be protected.

Also, if this trait is only used on `Tenancy`, all of the methods can be instance methods, no need for them to be static.

stancl: Ah looks like they are used in other places. But the comment about static still applies.

src/Jobs/CreatePostgresUserForTenant.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $centralConnection = config('tenancy.database.central_connection');
+        $centralConnectionPassword = config("database.connections.$centralConnection.password");
+
+        $name = $this->tenant->database()->getUsername() ?? $this->tenant->getTenantKey();
+        $password = $this->tenant->database()->getPassword() ?? $centralConnectionPassword;
+
+        // Create the user only if it doesn't already exist
+        if (! count(DB::select('SELECT usename FROM pg_user WHERE usename = $1', [$name])) > 0) {
+            $formattedStatement = DB::select("SELECT format('CREATE USER %I LOGIN PASSWORD %L', '$name', '$password');")[0]->format;
```
stancl, line 45: Could we use a parameterized query instead of using variables in a string?

Username & password *should* be safe, but there are edge cases where they could come from user input. In which case the user could bypass the quoting I think.

src/Database/Concerns/BelongsToTenant.php
lukinovec: Yeah, that's correct

src/Database/Concerns/DealsWithModels.php
lukinovec: Deleted the comment (the `$modelsExcludedFromDiscovery` property was already there, so this comment was a leftover). Seems like this was the only priority to-do added in this PR, so I think we can resolve this.

src/Database/Concerns/DealsWithModels.php
lukinovec: Made the methods non-static https://github.com/archtechx/tenancy/pull/1105/commits/692860eb6f1728e766710f09487b9fafe0c1f0b6

src/Jobs/CreatePostgresUserForTenant.php
lukinovec: I was trying to use parameterized query initially, but that didn't work. I've been looking for a way of using parameterized query here, but I couldn't find anything that worked, and it looks like Postgres doesn't support parameterized placeholders for queries like "CREATE USER x LOGIN PASSWORD y".

I think the current format() approach might be the most we can do, though it's sort of a security risk.

src/Database/Concerns/DealsWithModels.php
lukinovec: It is used only with RLS (so far), but I don't think we have to make it specifically an RLS thing. The method just retrieves models that should belong to tenants, so I'd leave it as is

src/Jobs/CreatePostgresUserForTenant.php
stancl: Searched a bit and seems like `%I` and `%L` properly escape everything.

stancl:
Resubmitted here https://github.com/tenancy-for-laravel/v4/pull/33

You can ignore the remaining reviews here, I'll resubmit all relevant ones in the new PR

PR #1116, opened 2023-05-30
Fixes custom disk file url for non-root_override disks

SurajDadral:
Storage Url of custom disk configured in tenancy.filesystem.disks but without root_override was giving wrong url when generated using `Storage::disk('branding')->url('logo.png')`

Modified this behaviour same as for `root` option, which appends `tenant<id>` to disk root for non-root_override disks.

Before:
`Storage::disk('branding')->url('logo.png')` returns `'storage/logo.png'`

After:
`Storage::disk('branding')->url('logo.png')` returns `'storage/tenant1/logo.png'`

### DIFF
--- a/src/Bootstrappers/FilesystemTenancyBootstrapper.php
+++ b/src/Bootstrappers/FilesystemTenancyBootstrapper.php
@@ -72,6 +72,11 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
             }
 
             $this->app['config']["filesystems.disks.{$disk}.root"] = $finalPrefix;
+
+            if (!isset($this->app['config']["tenancy.filesystem.root_override.{$disk}"])) {
+                $originalUrl = $this->app['config']["filesystems.disks.{$disk}.url"];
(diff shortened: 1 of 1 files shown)


stancl:
> but without root_override

Why not simply use root override here?

SurajDadral:
> > but without root_override
> 
> Why not simply use root override here?

To serve files like branding images directly from web server using storage:link option of laravel, we need to store files in `storage/app/public/` directory.
And to have files separated for each tenant in different directory like `storage/app/public/tenant1/`, `storage/app/public/tenant2/`,etc. , we can make disk tenancy aware by defining in config `tenancy.filesystem.disks`

If we use `root_override` option here, then files will be stored in `storage/tenant1/app/public/`, and to serve them directly using web server, we need to link public directory of each tenant in `public/storage/tenant{n}`

So, instead of that, we can fix url generation of non-root_override disks and then serve them directly from `storage/app/public/`

Let me know if I am missing something here.

Thanks,

stancl:
Think this should resolve that in v4 https://github.com/archtechx/tenancy/pull/909

PR #1121, opened 2023-06-11
Fix tenant identification in assets when using path-identification

craigrileyuk:
By default, calling `tenant_asset()` results in an asset URL like '/tenancy/assets/{path}`. This works fine for everything except path-identification since the tenant_id is required to be in the path itself.

These fixes add a second parameter to `tenant_asset(string $path, bool $usePathIdentification = false)` which uses a different route to enable proper URLs can be generated.

This restored functionality required a few other tweaks and some extra protection against path traversal attacks.

### DIFF
--- a/assets/routes.php
+++ b/assets/routes.php
@@ -8,6 +8,6 @@ Route::get('/tenancy/assets/{path?}', 'Stancl\Tenancy\Controllers\TenantAssetsCo
     ->where('path', '(.*)')
     ->name('stancl.tenancy.asset');
 
-Route::get('/{tenant}/assets/{path?}', 'Stancl\Tenancy\Controllers\TenantAssetsController@asset')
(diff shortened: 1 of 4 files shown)


zscalersptest:
comment

stancl:
Hi, thanks for the PR. We're dealing with improving path identification and the asset helper in v4 now, so I went over the code here in depth and I think we'll be handling things differently.

Regarding path traversal: this doesn't seem to be an issue on modern webservers, but it might be worth adding that check just in case. I'll commit this in a modified way to the 3.x branch.

Regarding the duplicated route and `forgetParameter()`, we have a different way of doing that in v4, so we'll address it there.

stancl:
https://github.com/archtechx/tenancy/commit/4af70d302ffcf19306cc16009bd94162f3743ca7

PR #1126, opened 2023-06-21
Fixed integrity constraint violation

icaliman:
Fixed this error:

`PDOException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1' for key 'PRIMARY'`


stancl:
Hey, thanks for the PR. This was fixed in v4 in a bit more detailed way (letting people specify the synced attributes in detail, with default values etc). We don't want to change the logic in v3 since it'd come with breaking changes.

This is technically a bug but some people's apps might work by assuming this behavior, so we'd rather not touch it in v3 and just release v4 with the many improvements to synced resources that we've made.

PR #1128, opened 2023-06-23
Close tenant connections after jobs

### LINKED ISSUE #1022: tenants:run with jobs combined with queue:work keeps multiple database connections open
### Bug description

I am facing a problem while running some jobs with queues combined with `tenants:run`.

I am executing a company command with a heavy job inside (usually i had a supervisor making this task but has been disabled for debugging), all jobs are sent to Redis before will be processed.

The problemas arrives when command is executed for all tenants, in that case, every time I try to run the command, a new connection is open for each tenant keeping previous one open, and finally provoking a collapse in MySQL throwing "Too many connections" when running via supervisor.

Currently problem is solved using queue:listen instead but the performance is better with queue:work command.

Example: 3 tenants in my system, 2 `tenants:run` iterations executed and you can see duplicates connections.

capture

### Steps to reproduce

- Execute php artisan queue:work myCustomQueue --tries=3 --sleep=3 (queue is ready to accept jobs)

- php artisan tenants:run company:myCommand

### Expected behavior

Keep connections open by tenant and use it instead of create new one.

### Laravel version

8.83.18

### stancl/tenancy version

3.5.7

lukinovec:
This PR changes QueryTenancyBootstrapper so that after a job gets processed, the context gets reverted to the previous one. This prevents having extra tenant connections that wouldn't get closed (should fix #1022).

### DIFF
--- a/tests/QueueTest.php
+++ b/tests/QueueTest.php
@@ -86,7 +86,7 @@ test('tenant id is not passed to central queues', function () {
     });
 });
 
-test('tenancy is initialized inside queues', function (bool $shouldEndTenancy) {
+test('tenancy is initialized inside queues', function () {
     withTenantDatabases();
     withFailedJobs();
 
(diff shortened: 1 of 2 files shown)


tests/QueueTest.php
```diff
@@ -182,6 +193,22 @@
     expect(pest()->valuestore->get('tenant_id'))->toBe('The current tenant id is: acme');
 });
 
+test('tenant connections do not persist after tenant jobs get processed', function() {
```
stancl, line 196: If you removed the `static::revertToPreviousState()` call here https://github.com/archtechx/tenancy/pull/1128/files#diff-d2557c1097c40e15ec7fc42979d9d8133351c52d2c0e5ba6fb89d2e78caffbdaR72, would this test fail?

lukinovec: Yeah, the tenant connection remains open, so this test fails

PR #1151, opened 2023-10-02
Make universal routes work for controller middleware

chillbram:
This PR allows support for controller middleware. `$route->computedMiddleware` is equal to ```Router::uniqueMiddleware(array_merge($route->middleware(), $route->controllerMiddleware()));```. So basically the same result as before, but with the controllerMiddleware added. I have added a test, which should cover this use case.

### DIFF
--- a/src/Features/UniversalRoutes.php
+++ b/src/Features/UniversalRoutes.php
@@ -35,7 +35,7 @@ class UniversalRoutes implements Feature
 
     public static function routeHasMiddleware(Route $route, $middleware): bool
     {
-        if (in_array($middleware, $route->middleware(), true)) {
+        if (in_array($middleware, $route->computedMiddleware, true)) {
             return true;
         }
 
--- a/src/Features/UniversalRoutes.php
+++ b/src/Features/UniversalRoutes.php
@@ -35,7 +35,7 @@ class UniversalRoutes implements Feature
 
     public static function routeHasMiddleware(Route $route, $middleware): bool
     {
-        if (in_array($middleware, $route->computedMiddleware, true)) {
+        if (in_array($middleware, $route->computedMiddleware ?? $route->middleware(), true)) {
             return true;
         }
 
--- a/tests/UniversalRouteTest.php
+++ b/tests/UniversalRouteTest.php
@@ -63,4 +63,46 @@ class UniversalRouteTest extends TestCase
             ->assertSuccessful()
             ->assertSee('acme');
     }
+
+    /** @test */
+    public function universal_route_works_when_middleware_is_inserted_via_controller_middleware()
+    {
+        Route::middlewareGroup('universal', []);
+        config(['tenancy.features' => [UniversalRoutes::class]]);
+
+        Route::get('/foo', [UniversalRouteController::class, 'show']);
+
+        $this->get('http://localhost/foo')
+            ->assertSuccessful()
+            ->assertSee('Tenancy is not initialized.');
+
+        $tenant = Tenant::create([
+            'id' => 'acme',
+        ]);
+        $tenant->domains()->create([
+            'domain' => 'acme.localhost',
+        ]);
+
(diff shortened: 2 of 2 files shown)


chillbram:
@stancl Could you take a look at this PR. For my projects this would be the last puzzle piece to make Tenancy v3 work with Livewire v3

stancl:
Curious how you're using controller middleware with Livewire. Are you specifying it in the component in some way?

And is there a reason you can't use route middleware here?

chillbram:
Livewire v3 itself now uses controller middleware for the File Upload feature by using the [getMiddleware() function](https://github.com/livewire/livewire/blob/main/src/Features/SupportFileUploads/FileUploadController.php). Now that [this Livewire PR](https://github.com/livewire/livewire/pull/6660) is merged, we can define `['universal', InitializeTenancyByDomain::class]` inside the Livewire config and apply the correct tenancy middleware for the file upload routes. 

With the way this feature is coded now, there is no other way to apply middleware to this route. A PR was opened to allow custom Route definitions, but that was closed due to issues with the implementation.

stancl:
@chillbram Does your PR handle: this route as well? https://github.com/livewire/livewire/blob/1ec54cdf57c6f1bf734bb14f9732904473b91c3f/src/Features/SupportFileUploads/SupportFileUploads.php#L38

And wouldn't your `array_map()` lead to route definition like this:
```php
[
    [
      "middleware" => "tenancy",
      "options" => [],
    ],
    [
      "middleware" => "universal",
      "options" => [],
    ],
]
```
Which I *think* should be:
```php
[
    [
      "middleware" => ["tenancy", "universal"],
      "options" => [],
    ],
]
```

chillbram:
No it does not handle the preview file. I don't know why it doesn't implement middleware in the same way, but I've never used that feature and have no idea how it works.

You're right that the array_map forms the array in that way, but that's how Laravel wants it. I guess they want to allow options to be set for each individual middleware. I tested passing an array to the middleware array key, but Laravel does not support that.

stancl:
I see, so the usage of `options` would be (in a generalized example) something like this?
```php
Route::group([
    'prefix' => 'foo',
    'middleware' => [
        ['middleware' => FooMiddleware::class, 'options' => ['options for foo middleware']],
        ['middleware' => BarMiddleware::class, 'options' => ['options for bar middleware']],
    ],
], function () {
    //
});
```

chillbram:
I don't think your code example works when used like that on the `Route::` facade. But if you'd move that middleware array to a `getMiddleware()` function in a controller, that would have the result you'd expect. Controller middleware isn't really documented I don't think, so we kind of have to live with their quirks.

chillbram:
@stancl Do you agree with this implementation and its test coverage?

stancl:
We're looking into this at the moment. The solution used here is causing some issues with how we do things in v4, so we're looking at other ways to approach the problem.

stancl:
@chillbram could you try if changing the Livewire code locally [like this](https://github.com/livewire/livewire/compare/main...lukinovec:livewire:file-controller-middleware) fixes the issue for you?

chillbram:
It works for files uploaded on a tenant, but not on the central domain. Universal routes don't seem to work with this implementation yet

stancl:
I submitted a PR to Livewire with a similar implementation to the one I linked.

Once it's merged, if you could check whether the changes submitted here work with the new version of Livewire, I can merge this.

Our priority is Livewire 3 support in v4, but since this is essentially a one line PR and doesn't appear to break any tests, I can merge this into v3.

chillbram:
Just checked and this change still works with Livewire v3.4.1, so I think this is safe to merge into Tenancy v3

stancl:
I added a fallback to `middleware()` in case this code was ever executed before `computedMiddleware` got filled (since the value of that property depends on the point in the lifecycle where the route is referenced, i.e. `gatherMiddleware()` needs to be called first).

PR #1153, opened 2023-10-07
fix: don't create database if its exists

mmkhak:


stancl:
Throwing an exception is the correct behavior. If you want to _explicitly_ skip database creation, you can use `tenancy_create_database => false`.

https://github.com/archtechx/tenancy/blob/85c7465aca2fb017545a5304dea0ee0006bee5f0/src/Jobs/CreateDatabase.php#L35

mmkhak:
I won't create the database if it exists

PR #1161, opened 2023-11-05
Execute the bootstrappers in reverse order when reverting back to central

edrep:
Bootstrappers order may be relevant in specific cases.
For example, when a custom bootstrapper queries the tenant database, it will be placed after the Database bootstrapper in the bootstrappers list and it must be reverted before the database connection is reverted back to central.


stancl:
> For example, when a custom bootstrapper queries the tenant database

Curious what use case you have for this

PR #1166, opened 2023-11-15
Fix connection config template array_merge parameter order

sergiopa2000:
I think this `array_merge($templateConnection, $this->tenantConfig())` is not merging the config correctly, it should merge it in reverse so it preserves the template config.


stancl:
Don't see the point of this change. `$this->tenantConfig()` should take precedence over `$templateConnection`.

sergiopa2000:
Yeah you are right, tought "tenantConfig()" was giving another thing, sorry. Im closing the PR

PR #1176, opened 2024-01-04
fix check isSubdomain

bangnokia:
I think append the dot `.` before the central domain could avoid the case central domain is considered as sub domain


stancl:
Not needed here, the code is used with the PreventAccess middleware to handle central requests correctly.

With this change, if you look at where `isSubdomain()` is used, the only difference would be that the central domain would be treated as a tenant domain instead of a tenant subdomain.

PR #1181, opened 2024-01-30
Allow suffix for prefixed Redis connections

miagg:
This will allow suffixing Redis tenant connections with a configured suffix or default to `_` for better readability.

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -149,6 +149,7 @@ return [
      */
     'redis' => [
         'prefix_base' => 'tenant', // Each key in Redis will be prepended by this prefix_base, followed by the tenant id.
+        'suffix' => '',
         'prefixed_connections' => [ // Redis connections whose keys are prefixed, to separate one tenant's keys from another.
             // 'default',
         ],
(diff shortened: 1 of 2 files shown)


stancl:
Could consider merging this, though some things would need to be resolved first:
- the `'_'` default is a breaking change
- all new functionality needs to be tested
- not sure about the `suffix` term. Can't really think of anything better though this makes it sound like a suffix that could go *after* the entire key.

Maybe I could change this to be
```
'prefix' => 'tenant%tenant%_',
```
in v4, where `%tenant%` gets replaced by the current tenant key.

miagg:
> Could consider merging this, though some things would need to be resolved first:
> 
> * the `'_'` default is a breaking change
> * all new functionality needs to be tested
> * not sure about the `suffix` term. Can't really think of anything better though this makes it sound like a suffix that could go _after_ the entire key.
> 
> Maybe I could change this to be
> 
> ```
> 'prefix' => 'tenant%tenant%_',
> ```
> 
> in v4, where `%tenant%` gets replaced by the current tenant key.

I agree on using a template string. Not sure how close v4 is though. Would it be merged in v3 if I remove the `_` default?

colinmackinlay:
@miagg @stancl Might this also be relevant to Laravel 11 since the : is no longer appended to prefixes?
https://laravel.com/docs/11.x/upgrade#cache-key-prefixes
I'm currently trying to get my upgrade to L11 finished and wrestling with some caching errors which might be related

stancl:
I will not be merging this PR into v3, but yes, L11 had some cache prefix changes. I don't think there's anything needed to change on our end, just flush all cache. Laravel basically just stopped inconsistently adding `:`. In v4 we just updated some low level test asssertions.

stancl:
Implemented in v4 as described above.

PR #1199, opened 2024-03-24
Added ability for custom resolving column in PathTenantResolver

tijanidevit:


stancl:
I appreciate this PR but unfortunately I won't be able to merge it due to missing cache (invalidation) logic + tests, and the fact that in v3 this is a bit broken in general so I'd need to refactor how resolvers work in the first place.

I ported this into version 4 and added a proper config key for the used column, as well as support for binding route fields.

PR #1215, opened 2024-04-26
Fixes UniversalRoute overrides custom middleware onFail

Noah-Wilderom:


stancl:
This change fixes universal routes overriding custom `$onFail`, but replaces it with custom `$onFail` breaking universal routes.

In v4 the universal routes logic is completely changed so I don't recall if we had fixed this there in some form earlier, but I think trying to combine the two approaches in the way the package currently works would be fragile since the order in which these are set can be context-dependent.

For what it's worth, the docs page for universal routes does mention this behavior including how to address it in your application so it's not considered a bug.

PR #1217, opened 2024-04-29
Enhance Cache Methods to Handle Tag Support with Different Drivers

rtraselbd:
Enhance Cache Methods to Handle Tag Support with Different Drivers

This commit modifies the cache methods in Laravel to handle tag support differently based on the driver being used. When using a driver that supports tags, such as the array driver, the methods like Cache::put(), Cache::get(), etc., will utilize tagging functionality as expected. However, for drivers like the default database driver, which do not support tags, the methods will now fallback to direct cache operations without tagging. 

This enhancement ensures consistent behavior across different cache drivers, optimizing functionality for both tag-supported and non-tag-supported drivers.

### DIFF
--- a/src/CacheManager.php
+++ b/src/CacheManager.php
@@ -32,6 +32,10 @@ class CacheManager extends BaseCacheManager
             return $this->store()->tags(array_merge($tags, $names));
         }
 
-        return $this->store()->tags($tags)->$method(...$parameters);
+        if ($this->store()->supportsTags()) {
+            return $this->store()->tags($tags)->$method(...$parameters);
+        }
+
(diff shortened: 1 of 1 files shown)


stancl:
From what I understand, this would just lead the bootstrapper silently failing when cache can't be scoped?

It's better to get an exception, notice that you're not using a driver that supports tagging, and reconsider from there how you want to handle caching, than to falsely believe your data is getting scoped.

> ensures consistent behavior

It's making it inconsistent.

If you're trying to use a driver that doesn't support tagging, you can simply comment out the bootstrapper in the tenancy config file. Your cache won't be scoped though.

PR #1228, opened 2024-05-24
Fix tenant_route without absolute

hecktor-devsquad:
This PR aims to resolve the issue when trying to pass `Absolute: false` to the `tenant_route` helper

When we try to generate a route with the `tenant_route` helper, passing absolute as false, we receive the following error.
`tenant_route(): argument #1 ($domain) must be of type string, null provided, called on line 1.`

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -61,7 +61,7 @@ if (! function_exists('tenant_route')) {
     {
         // replace first occurance of hostname fragment with $domain
         $url = route($route, $parameters, $absolute);
-        $hostname = parse_url($url, PHP_URL_HOST);
+        $hostname = parse_url($url, PHP_URL_HOST) ?? '';
         $position = strpos($url, $hostname);
 
         return substr_replace($url, $domain, $position, strlen($hostname));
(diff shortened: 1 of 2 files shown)


stancl:
What use case do you have for calling the `tenant_route()` helper like this? It's a `route()` helper that generates the URL **on a specific hostname**, so how would relative URLs work?

hecktor-devsquad:
I need to send a protocolless URL, and the easiest way to get it using this helper is to set `absolute` to false.

More like `parse_url` in these cases returns `null` and we treat strings later, it ends up giving the error.

willypriv:
I'm having the same problem here, I need to get the url without the 'https', but if I pass the `absolute` as `false` I get the same error.

stancl:
That to me sounds like abusing the implementation details of `tenant_route()` tbh. The helper is meant to generate absolute URLs with the specified hostname. Passing `absolute: false` would be a hack, since **you still want an absolute URL**, just without `https://`. Just do string manipulation on the result or write a custom helper.

hecktor-devsquad:
@stancl  Ok, so it doesn't make any sense for this helper to have the `absolute` parameter as it already is!

stancl:
I think I went with that function signature to keep it identical to `route()`, but yeah, `$absolute` will be removed since it doesn't make sense to use here. Either way, **it has nothing to do with whether the protocol is included**, that's not what absolute/relative routes refers to.

PR #1245, opened 2024-08-22
Fix Duplicate Primary Key Error During Tenant to Central Database Synchronization

dipaksarkar:
Resolved an issue where creating records in the central database from a tenant context could result in duplicate primary key errors due to conflicting `id` values. Introduced `getCentralModelFillable` in the `Syncable` interface and `getTenantModelFillable` in the `SyncMaster` interface to ensure that only fillable attributes are synced during resource creation. This fix prevents `id` conflicts by excluding it from the synced attributes, allowing for seamless synchronization without primary key collisions.

### DIFF
--- a/src/Contracts/BaseSyncable.php
+++ b/src/Contracts/BaseSyncable.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Contracts;
+
+interface BaseSyncable
+{
+    public function getGlobalIdentifierKeyName(): string;
+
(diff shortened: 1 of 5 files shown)


stancl:
We've decided we won't be making any large changes like these to resource syncing in v3, since nearly every bug fix would break some use case or expected behavior. All of these issues have been fixed in v4 already. I don't think the abstractions used in this PR make much sense anyway.

ii-jj:
protected $guarded = ['id'];

Can solve this problem

dipaksarkar:
I tried it doesn't worked. For this reason i have created this PR.

PR #1246, opened 2024-08-22
add early exit for 'runForMultiple' method

dvarilek:
This PR adds the ability to stop processing further tenants in the 'runForMultiple' method by returning false from the passed callback function. This behaviour is similar to that of Laravel's EnumerableValues trait's 'each' method early exit.

This feature could be particularly useful when a developer needs to check whether a value is unique across tenant databases, allowing them to exit early once a match is found.

### DIFF
--- a/src/Tenancy.php
+++ b/src/Tenancy.php
@@ -158,7 +158,7 @@ class Tenancy
 
             $this->initialize($tenant);
 
-            if (! $callback($tenant)) {
+            if ($callback($tenant) === false) {
                 break;
             };
         }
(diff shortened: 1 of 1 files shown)


stancl:
This is a major breaking change.

dvarilek:
I see, sorry, there should be a strict false comparison instead of the falsey check since this would basically break out of the for loop between the first and second iteration. I'll submit a fix

stancl:
Even that's a risky change. I'd suggest just implementing your own `runForMultiple()` that checks for this as needed.

PR #1251, opened 2024-09-10
Always revert queue worker to the central context — fix #1229

### LINKED ISSUE #1229: Using RedisTenancyBoostrapper causes problems with queue worker
### Bug description

When calling `queue:restart`, it will cause all future `queue:work` commands to exit after only one job, if the job was dispatched in tenant context.

### Steps to reproduce

See this demo repository: https://github.com/pr4xx/tenancy-problem-demo

### Expected behavior

`queue:work` does not stop itself

### Laravel version

11.9

### stancl/tenancy version

3.8

stancl:
This PR changes the QueueTenancyBootstrapper to **always** revert back to the previous context. In practice, this means reverting back to the central context *unless* `dispatchNow()` was used.

This addresses the issue described in #1229 where if the queue worker stayed in the tenant context, it would not respond to queue:restart signals.

The `$forceRefresh` static property is now unused and deprecated.

v4 will simplify the bootstrapper further since there are fewer backwards compatibility concerns.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,6 @@ jobs:
     - name: Start docker containers
       run: PHP_VERSION=${{ matrix.php }} docker-compose up -d
     - name: Install dependencies
-      run: docker-compose exec -T test composer require --no-interaction "laravel/framework:^${{ matrix.laravel }}.0"
+      run: docker compose exec -T test composer require --no-interaction "laravel/framework:^${{ matrix.laravel }}.0"
     - name: Run tests
       run: ./test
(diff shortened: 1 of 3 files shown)


stancl:
@pr4xx Sorry about the delay on this, could you test the queue bootstrapper version from this PR?

https://raw.githubusercontent.com/archtechx/tenancy/e6ab0ee9af48771def77febc37933910cae52529/src/Bootstrappers/QueueTenancyBootstrapper.php

niconico291:
@stancl Thanks for the update. I deployed your version of the QueueTenancyBootstrapper to the testing environment and after a few short tests, it seems to work fine. I will report back after some more days of testing.

pr4xx:
@stancl Some time has passed and had no issue with the updated code in production.

stancl:
Awesome, thanks for all the help with this.

PR #1253, opened 2024-09-12
feat: Added maintenance mode events

chinmaypurav:
I would like to listen to the Maintenance mode events so I can plan out certain actions on that tenant.


stancl:
We already have this implemented in v4 in a bit cleaner way.

chinmaypurav:
Okay. Waiting for v4!

PR #1257, opened 2024-09-23
Fix ModelNotSyncMasterException message

coreyhn:
Fix the ModelNotSyncMasterException message by replacing "an" with "a" and adding a period at the end of the sentence to match the style of the other exceptions.


stancl:
Thanks!

PR #1268, opened 2024-11-25
feat: write cache in separate method

jaulz:
I have a case when a lot of concurrent requests all try to connect at the same time to the central database to refresh the cache. This is not optimal because in my case it takes all the available connections (from RDS proxy) and no connections for the actual target database are free anymore. Hence, I would like to control myself when the cache is written and use a job instead. I could copy the logic to my custom job but it would be nicer if I could simply call the new `writeCache` method.

### DIFF
--- a/src/Resolvers/Contracts/CachedTenantResolver.php
+++ b/src/Resolvers/Contracts/CachedTenantResolver.php
@@ -40,17 +40,17 @@ abstract class CachedTenantResolver implements TenantResolver
             return $tenant;
         }
 
-        $this->writeCache(...$args);
-
-        return $tenant;
+        return $this->writeCache(...$args);
     }
 
-    public function writeCache(...$args): void
+    public function writeCache(...$args): Tenant
     {
         $key = $this->getCacheKey(...$args);
         $tenant = $this->resolveWithoutCache(...$args);
 
         $this->cache->put($key, $tenant, static::$cacheTTL);
+
+        return $tenant;
     }
 
     public function invalidateCache(Tenant $tenant): void
--- a/src/Resolvers/Contracts/CachedTenantResolver.php
+++ b/src/Resolvers/Contracts/CachedTenantResolver.php
@@ -34,18 +34,23 @@ abstract class CachedTenantResolver implements TenantResolver
             return $this->resolveWithoutCache(...$args);
         }
 
-        $key = $this->getCacheKey(...$args);
-
-        if ($tenant = $this->cache->get($key)) {
+        if ($tenant = $this->cache->get($this->getCacheKey(...$args))) {
             $this->resolved($tenant, ...$args);
 
             return $tenant;
         }
 
-        $tenant = $this->resolveWithoutCache(...$args);
-        $this->cache->put($key, $tenant, static::$cacheTTL);
+        $this->writeCache(...$args);
 
         return $tenant;
+    }
+
+    public function writeCache(...$args): void
+    {
+        $key = $this->getCacheKey(...$args);
+        $tenant = $this->resolveWithoutCache(...$args);
+
+        $this->cache->put($key, $tenant, static::$cacheTTL);
     }
 
     public function invalidateCache(Tenant $tenant): void


stancl:
I'm not sure what this PR is solving. If you deferred the cache write to a queued job, the request would *still* be trying to identify the tenant directly from the DB?

Are you trying to optimize database connections or cache connections here?

src/Resolvers/Contracts/CachedTenantResolver.php
```diff
@@ -34,20 +34,25 @@ public function resolve(...$args): Tenant
...
-        if ($tenant = $this->cache->get($key)) {
+        if ($tenant = $this->cache->get($this->getCacheKey(...$args))) {
             $this->resolved($tenant, ...$args);
 
             return $tenant;
         }
 
-        $tenant = $this->resolveWithoutCache(...$args);
-        $this->cache->put($key, $tenant, static::$cacheTTL);
+        $this->writeCache(...$args);
```
stancl: How would the `return $tenant` work now?

jaulz: You are right 😊 Now `writeCache` returns the tenant. Maybe we could even rename it to `resolveAndCache` to make it clearer?

jaulz:
If I cache the tenants every 12 hours (for example)  and the TTL is 24 hours, no request should anytime hit the database, shouldn't it?

stancl:
Again, not sure what you're trying to do.

jaulz:
Sorry, for the bad explanation. I would like to have a job that runs every 12 hours that caches all tenants and their domains:
```php
foreach ($domains as $domain) {
  app(DomainTenantResolver::class)->writeCache($domain);
}
```

Once I have this job, no request would every hit the central database and I don't have to worry about too many requests trying to cache the current domain/tenant association.

stancl:
Too many connections occurs _because_ you're trying to cache all domains at once. The feature is meant to update cache as it's used. If you want to maximize the cache, you can simply increase the TTL.

jaulz:
If my domain is hit by 100 requests at once and there is nothing in the cache, it will try to read the domain configuration from the database for every single request. Every request will connect to central database first and will then try to change to the tenant database. The problem is now that RDS proxy reserves the 100 connections for the central database for X minutes and not a single request can switch to tenant database because the maximum number of connections was reached.

Now, if I have the job like described above I can cache the tenant/domain configuration in advance and no request will every try to hit the central database.

I hope this makes my case clearer. Sorry for the confusion 😊

stancl:
> If my domain is hit by 100 requests at once and there is nothing in the cache, it will try to read the domain configuration from the database for every single request

If the same tenant domain is hit by 100 requests at once, at most n (= php-fpm worker count) requests will try to connect to the central database before requests start reading from the cache.

I think you're just trying to hack around how the cache TTL works. Why not just increase it? And again, I don't think the actual problem is being described well here. There will not be 100 concurrent DB requests if one tenant gets 100 requests at once. Additionally, requests will be naturally spread across different tenants which will update each tenant as it's used.

It just sounds like there's something misconfigured on your end and this convo is an xyproblem misdiagnosing it. I don't think you should be seeing any more central connections than you have fpm workers in the first place. You can try things like persistent connections, increasing the cache TTL here, and looking into what's actually happening more carefully.

jaulz:
I should have told you that I'm using Bref (i.e. serverless) and the number of requests is not limited but the worker count. If I increase the TTL it will just shift the problem to a later point in time. Thus, I don't think it's a hack at all 🤔

stancl:
If you want to implement it that way, I don't think a new method needs to be added. All of the relevant methods are public and you can just write logic that sets the cache key as desired. It is a hack to be using resolveWithoutCache() for extending existing entries. You can just update the records in your cache to have infinite TTL.

But I want to point out that updates to tenants clear this cache, so you'd also need to instantly re-cache updated tenants. And frankly it seems like an architectural flaw if the app grinds to a halt when the cache is not populated. It's only meant to marginally improve response times and lower database load.

If there's a problem with 100 requests opening 100 connections to the central database (just until the cache gets populated, which will take around the time of 1 request), I don't see how there wouldn't be a problem with 100 connections being opened for a tenant database concurrently, nor how this would scale as the number of those requests grows.

jaulz:
Okay, thanks for your comments anyway. Though, I still think you didn't really understand the issue and my changes. My suggested change was just to add a `writeCache` method which can be publicly used and without having to think about the internals. I didn't mention using `resolveWithoutCache` at all. Also it's not about having infinite TTL because my change would only extend entries and non-existent entries would still expire. 

The difference between central database and tenant database is that connections to the central database are used only once and afterwards they are not needed anymore.

stancl:
From the description I was given here I don't see a legitimate reason for making a new public "convenience method" that I haven't been convinced has a good use case. The existing logic is public and you can achieve what you're trying to do with it.

> I didn't mention using resolveWithoutCache at all

```php
public function writeCache(...$args): Tenant
{
    $key = $this->getCacheKey(...$args);
    $tenant = $this->resolveWithoutCache(...$args);

    $this->cache->put($key, $tenant, static::$cacheTTL);

    return $tenant;
}
```

> used only once and afterwards they are not needed anymore

Doesn't address the discrepancy in your scaling concerns. If this were truly a concern, you'd be in a position where you cannot add any more requests/sec or tenants to your application. If it's not a true concern, you should use the cache logic as intended.

PR #1278, opened 2024-12-29
[4.x] General code cleanup

xHeaven:
This PR closes https://github.com/archtechx/tenancy/pull/1271 as well.


src/Concerns/ParallelCommand.php
```diff
@@ -48,15 +48,17 @@ protected function forkProcess(mixed ...$args): int
...
-            // Child
-            DB::reconnect();
-
-            exit($this->childHandle(...$args) ? 0 : 1);
         }
+
+        // Child
+        DB::reconnect();
+
+        exit($this->childHandle(...$args) ? 0 : 1);
```
stancl: When forking I prefer to handle all cases in a single if-elseif-else block rather than basically early returns of varying indentation, think it reads a bit better to have all cases formatted the same.

src/Concerns/ParallelCommand.php
```diff
@@ -70,7 +72,9 @@ protected function sysctlGetLogicalCoreCount(bool $darwin): int
         // perflevel0 refers to P-cores on M-series, and the entire CPU on Intel Macs
         if ($darwin && $ffi->sysctlbyname('hw.perflevel0.logicalcpu', FFI::addr($cores), FFI::addr($size), null, 0) === 0) {
             return $cores->cdata;
-        } else if ($darwin) {
+        }
+
+        if ($darwin) {
```
stancl: Think there's no real difference between this version of the code and the previous, but bit like in the previous review, I prefer to have this joined as a single if-elseif block.

src/Listeners/QueueableListener.php
```diff
@@ -17,12 +17,12 @@ public function shouldQueue(object $event): bool
...
-            $this->handle($event);
-
-            return false;
         }
+
+        // The listener is not queued so we manually
+        // pass the event to the handle() method.
+        $this->handle($event);
+
+        return false;
```
stancl: Same as the previous reviews, just reads better without the early return.

xHeaven:
@stancl Is there anything you like in these commits? [commit1](https://github.com/archtechx/tenancy/pull/1278/commits/ba5583523ffeacd36f7844777a8d10429b8922f5) [commit2](https://github.com/archtechx/tenancy/pull/1278/commits/374c704607f72bd49e7b8cad153f36b6e151e600) If no, I can just simply revert them. You are right, there is no real benefit in changing the structure, I just thought they look better/cleaner like this, but of course, it boils down to personal preference and what works best for you and your team.

stancl:
Basically everything I didn't submit a review for I'd keep. So first commit keep in full and second commit only revert what I commented on

src/Listeners/QueueableListener.php
xHeaven: [`fabc09a` (#1278)](https://github.com/archtechx/tenancy/pull/1278/commits/fabc09a16bf32fca9a42bb04df042a516de5d520)

src/Concerns/ParallelCommand.php
xHeaven: [`fabc09a` (#1278)](https://github.com/archtechx/tenancy/pull/1278/commits/fabc09a16bf32fca9a42bb04df042a516de5d520)

src/Concerns/ParallelCommand.php
xHeaven: [`fabc09a` (#1278)](https://github.com/archtechx/tenancy/pull/1278/commits/fabc09a16bf32fca9a42bb04df042a516de5d520)

xHeaven:
Reverted the unwanted changes.

Edit: PHPStan is currently failing due to this change: [`d2aaf5b` (#1278)](https://github.com/archtechx/tenancy/pull/1278/commits/d2aaf5b11b2f392c359daad001d7fc3be1545faf)
How do we want this resolved?

stancl:
Thanks a lot for the work on this PR!

PR #1280, opened 2024-12-30
[4.x] Fix: no primary key on RLS views

alec-superfile:
When running `php artisan tenants:rls` with a Postgres view an exception is thrown:

```
Attempt to read property "column_name" on null

at vendor/stancl/tenancy/src/Database/TenantDatabaseManagers/PermissionControlledPostgreSQLSchemaManager.php:44
```

This PR attempts to fix this by granting RLS permissions to just BASE TABLEs and ignoring views.

To test, add a Postgres view, my use case is below but any view will do

```sql
CREATE VIEW friends AS
SELECT
    f1.follower_id AS user_id,
    f1.following_id AS friend_id
FROM followers f1
INNER JOIN followers f2
    ON f1.follower_id = f2.following_id
    AND f1.following_id = f2.follower_id
WHERE f1.accepted_at IS NOT NULL
AND f2.accepted_at IS NOT NULL
```

then run `php artisan tenants:rls`

### DIFF
--- a/src/Database/TenantDatabaseManagers/PermissionControlledPostgreSQLSchemaManager.php
+++ b/src/Database/TenantDatabaseManagers/PermissionControlledPostgreSQLSchemaManager.php
@@ -27,7 +27,7 @@ class PermissionControlledPostgreSQLSchemaManager extends PostgreSQLSchemaManage
         $this->connection()->statement("GRANT USAGE, CREATE ON SCHEMA \"{$schema}\" TO \"{$username}\"");
         $this->connection()->statement("GRANT USAGE ON ALL SEQUENCES IN SCHEMA \"{$schema}\" TO \"{$username}\"");
 
-        $tables = $this->connection()->select("SELECT table_name FROM information_schema.tables WHERE table_schema = '{$schema}'");
+        $tables = $this->connection()->select("SELECT table_name FROM information_schema.tables WHERE table_schema = '{$schema}' AND table_type = 'BASE TABLE'");
 
         // Grant permissions to any existing tables. This is used with RLS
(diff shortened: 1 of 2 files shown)


stancl:
Thanks for the PR! Do you think you'd be able to add a regression test? It's probably enough to take some simple existing test, add a view to it, and expect an exception on `tenants:rls`.

alec-superfile:
Sure! One issue though, since my code fixes the issue I'm not sure how to get it to throw an exception.

In tests/RLS/PolicyTest.php

```php
test('rls command fails when a view is in the database', function (string $manager) {
    DB::statement("
        CREATE VIEW post_comments AS
        SELECT
            comments.id AS comment_id,
            posts.id AS post_id
        FROM comments
        INNER JOIN posts
            ON posts.id = comments.post_id
    ");

    DB::statement("ALTER VIEW post_comments SET (security_invoker = on)");

    config(['tenancy.rls.manager' => $manager]);

    try {
        pest()->artisan('tenants:rls');

        pest()->assertTrue(false);
    } catch (\Exception $e) {
        pest()->assertTrue(true);
    }
})->with([
    TableRLSManager::class,
    TraitRLSManager::class,
]);
```

This test will pass without my change and will fail with, let me know what I should do to get it to fail with my fix in place.

stancl:
Yeah with regression tests they're meant to pass now (after the fix) and fail without the fix. No real automated way of checking that besides undoing the change by hand.

alec-superfile:
Got it! Flipped the assertions :)

stancl:
Awesome. I've confirmed that the regression test fails without the patch and slightly refactored it. Thanks a lot for the PR!

PR #1285, opened 2025-01-07
[4.x] Test that global scopes on syncable models can break resource syncing, and that $scopeGetModelQuery can be used as a workaround for that 

lukinovec:
Having global scopes on syncable models breaks resource syncing in some cases (in response to https://discord.com/channels/976506366502006874/1043694788358189167/1325068028567883786).

When using global scopes on a model, you can take advantage of the `UpdateOrCreateSyncedResource::$scopeGetModelQuery` property to exclude the global scopes that make the syncable model records excluded from resource syncing queries. Note that this only affects resource syncing. For non-resource syncing queries, the global scope will still be preserved.

For example, in TSP's `boot()`, do:
```php
UpdateOrCreateSyncedResource::$scopeGetModelQuery = function (Builder $query) {
    if ($query->getModel()->hasGlobalScope(FooScope::class)) {
        $query->withoutGlobalScope(FooScope::class);
    }
};
```


tests/ResourceSyncingTest.php
```diff
@@ -108,6 +111,28 @@
     UpdateOrCreateSyncedResource::$scopeGetModelQuery = null;
 });
 
+test('multiple tenants can have users synced to a central resource', function() {
```
stancl: The name doesn't make it very clear what exactly this test asserts.
1. What's the reason for it being added here - what was not being tested before?
2. Can you make the name more descriptive

Is it just trying to assert that one user can "belong to multiple tenants"? Do we not test that elsewhere or is this just a basic test that establishes that basic thing, while other tests test this as part of larger assertions?

tests/ResourceSyncingTest.php
```diff
@@ -1173,6 +1198,53 @@
     expect($centralUser->foo)->toBe('bar');
 });
 
+test('resource syncing works correctly when using a global scope on a tenant model', function(bool $scopeGetModelQuery) {
```
stancl: Nitpick but this name implies it just works out of the box, and claims the opposite of the PR title. Also the "tenant model" is ambiguous - `Tenant` model or model in the tenant context.

Better name would be something _along the lines of_ "resource syncing can work with global scopes when scopeGetModelQuery is used" or "scopeGetModelQuery can make resource syncing work with global scopes" (maybe better).

tests/ResourceSyncingTest.php
```diff
@@ -1173,6 +1198,53 @@
...
+    if ($scopeGetModelQuery) {
+        UpdateOrCreateSyncedResource::$scopeGetModelQuery = function (Builder $query) {
+            if ($query->getModel()->hasGlobalScope(TestingScope::class)) {
+                $query->withoutGlobalScope(TestingScope::class);
+            }
+        };
+    }
+
+    $centralUser = CentralUser::firstOrCreate(
```
stancl: No idea why we're using ambiguous code like `firstOrCreate` when we have full control over the context in this test. If the user already exists, use it. If it doesn't, create it. **We know which one it is** in tests. In this case the latter, there's no user created in `beforeEach()`.

stancl: Resolved

tests/ResourceSyncingTest.php
```diff
@@ -1173,6 +1198,56 @@
...
+        pest()->expectExceptionMessage('Duplicate entry');
+    }
+
+    TenantUserWithScope::create([
+        'global_id' => $centralUser->global_id,
+        'name' => $centralUser->name,
+        'email' => $centralUser->email,
+        'password' => $centralUser->password,
+        'role' => 'admin'
+    ]);
```
stancl: Reading this test it's not completely clear why the exception would be thrown. Some comments would help, there's a lot of missing context. Alternatively rewrite this part of the test for clarity.

Right now it looks like:
1. TenantUserWithScope is scoped with `whereNull('name')`
2. TenantUserWithScope is created in tenant1
3. Then we try to create it in tenant2
4. If the global scope is bypassed, that works fine. If it's not, we get an exception.

But why? It's not clear why the user is created in tenant2. We never attach it to tenant2, so why do we get an exception? It's coming from the tenant DB (and the tenant user model, not the central one, is the one that's scoped) but I don't see how tenant2's DB would interact with the user created in tenant1 besides through the central model - but that one is not scoped and was already created.

It may be that the `create()` with the same `global_id` implicitly does an attach but that seems like unnecessarily depending on the implementation detail and still the test looks *wrong* — if we're just creating the user there *directly* I do not see why we would get this duplicate entry issue.

tests/ResourceSyncingTest.php
lukinovec: The test should be something like "creating the same user across multiple tenants does not create multiple central resources" (used that for now, pushed).

> Is it just trying to assert that one user can "belong to multiple tenants"?

It's more specific than that. The test checks that when tenant users with the same global id get created, only a single central user gets created, and the tenant resources are synced to that central user.

> Do we not test that elsewhere or is this just a basic test that establishes that basic thing, while other tests test this as part of larger assertions?

Other tests cover that one user can belong to multiple tenants through explicit attaching.

tests/ResourceSyncingTest.php
lukinovec: Right. Changed the name to "scopeGetModelQuery can make resource syncing work with global scopes", seems accurate enough

tests/ResourceSyncingTest.php
lukinovec: I see that this is pretty confusing. Added comments to make this clear

> It may be that the create() with the same global_id implicitly does an attach

You're right. When we attempt to create a tenant user with a name, and $scopeGetModelQuery is true, the central user won't be found because it has a name, and then, it will attempt to create a new central user with **the same global id**. This is the reason for the exception -- the global id is unique.

> if we're just creating the user there directly I do not see why we would get this duplicate entry issue

When creating a tenant resource, and a corresponding central resource is not found, we create it (and in this case, the central resource **does exist**, it just can't be found because of the scoping), makes sense to me

So in this test, we're checking if the $scopeGetModelQuery works, but at the same time, it tests an edge case. Maybe the test could be simpler, I don't even think this edge case needs to be covered, it's just weird.

lukinovec: I think testing the edge case is still worth. Made changes that should make the test's intentions clearer

tests/ResourceSyncingTest.php
```diff
@@ -1173,6 +1198,53 @@
     expect($centralUser->foo)->toBe('bar');
 });
 
+test('resource syncing works correctly when using a global scope on a tenant model', function(bool $scopeGetModelQuery) {
+    if ($scopeGetModelQuery) {
+        UpdateOrCreateSyncedResource::$scopeGetModelQuery = function (Builder $query) {
```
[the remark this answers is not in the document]
lukinovec: I don't think we want a helper function for this, the current code seems fine to me

stancl (changes requested):
The test should clearly assert what this PR says:
1. scopes can break resource syncing
2. scopeGetModelQuery can be used to fix that

Use the custom scope — change it if needed — instead of soft deletes. We already have tons of tests for soft deletes, including for bypassing the scope using `scopeGetModelQuery` + `withTrashed()`.

tests/ResourceSyncingTest.php
lukinovec: Updated the test and the comments

tests/ResourceSyncingTest.php
lukinovec: Simplified the test a lot, properly commented it and now, it's much clearer.

tests/ResourceSyncingTest.php
```diff
@@ -108,6 +110,29 @@
     UpdateOrCreateSyncedResource::$scopeGetModelQuery = null;
 });
 
+test('creating multiple tenant resources with the same global_id only creates corresponding central resource if it does not exist yet', function () {
```
lukinovec: Revert changes since this https://github.com/archtechx/tenancy/commit/e199b9d50c0cf7b5cc0b4f443660a260439e56fc#diff-0c32d140e95604d5ab8852beb5eba2255180184351adfaa0ff7b2895012d4d7dR114-R120

stancl (approved):

PR #1290, opened 2025-01-13
[4.x] Support `database` cache store tenancy

### LINKED ISSUE #852: Introduce database driver support for caching.
#### Description
The project should support the database driver as a caching option. The CacheManager forces the use of tags, making the database driver untenable. Instead, as an option, the CacheManager should use the tenant-initialized DB connections instead of relying on tags to separate cached data.

#### Why this should be added
Given the utility of the database driver for clean session scoping, the caching implementation for this project seems unusually database-averse. I realize it isn't the optimal driver for many caching processes, but it seems in line with the overall intention of the project. 

In a database implementation, the use of tags would be unnecessary (since the cache table could be created in each tenant). Access to the data would be separated by tenant. Resources consumed for cache maintenance would be separated by tenant.

In struggling to get other caching implementations to work for my project, I assumed (incorrectly) I could default to the database driver.

lukinovec:
This PR adds the DatabaseCacheBootstrapper. The bootstrapper can be used instead of CacheTenancyBootstrapper for the tenant cache to be saved in the tenant databases instead of keeping it in the central DB.

On `bootstrap()`, DatabaseCacheBootstrapper sets the `database` cache store's connection to 'tenant' and deletes the `database` cache store from the CacheManager's resolved stores -- this forces the manager to resolve a new instance of the `database` store created with the 'tenant' DB connection on the next cache operation.

On `revert()`, the bootstrapper reverts the `database` cache store's connection back to the originally used connection (again, by altering the database cache store's configured connection and deleting the `database` store from the manager's resolved stores).

Solves #852


src/Bootstrappers/DatabaseCacheBootstrapper.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * On revert, the cache store's connection is reverted to the originally used one (usually 'central'), and again,
+ * the database cache store is purged from the CacheManager's resolved stores so that the originally used one is resolved on the next cache operation.
+ */
+class DatabaseCacheBootstrapper implements TenancyBootstrapper
+{
+    public function __construct(
+        protected Repository $config,
+        protected CacheManager $cache,
+        protected string|null $originalConnection = null,
```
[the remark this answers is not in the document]
lukinovec: For the record, even though `$originalConnection` is initialized with `null`, we still set the property to `$this->config->get('cache.stores.database.connection')` in `bootstrap()`. So I don't think this comment/suggestion is relevant, as we're just initializing the property in the constructor. Though maybe it'd make sense to initialize it outside of the constructor, but I don't think that would make any difference in the end

src/Bootstrappers/DatabaseCacheBootstrapper.php
```diff
@@ -0,0 +1,52 @@
...
+        protected string|null $originalLockConnection = null
+    ) {}
+
+    public function bootstrap(Tenant $tenant): void
+    {
+        $this->originalConnection = $this->config->get('cache.stores.database.connection');
+        $this->originalLockConnection = $this->config->get('cache.stores.database.lock_connection');
+
+        $this->config->set('cache.stores.database.connection', 'tenant');
+        $this->config->set('cache.stores.database.lock_connection', 'tenant');
```
stancl: The *name* of the cache store is irrelevant. I may have multiple database-based stores. I may rename the database store to something else. All that makes it a database-based cache store is that `driver` is set to `database`.

Add:
```php
public static array|null $stores = null;
```

If set to null, loop over all cache stores and filter those that have a `driver` of `database`.

If set to an array, only look at the listed store names (perhaps still do the `driver` check though, so just a different way to create the array but the same rest of the logic).

That way this bootstrapper will work out of the box without any config, while still being configurable.

lukinovec: Makes sense. Updated the bootstrapper accordingly, also added tests here https://github.com/archtechx/tenancy/pull/1290/commits/872bf9df72d9fb7639e6f530329e68e9e058b05d

stancl:
Merged manually

PR #1293, opened 2025-01-14
[4.x] Make forcing RLS configurable

lukinovec:
This PR adds the `$forceRls` static property to the `tenants:rls` command.

If `$forceRls` is `true` (default), the table owners won't be able to query their own tables (unless the owners have the BYPASSRLS privilege). If the property is set to `false`, table owners will bypass RLS, allowing them to query the owned tables.

(follow up on #1288)

---

Additional context:
- By default we expect the central user to have the `BYPASSRLS` perm
- If that's not the case, this setting can be set to false and everything should still work normally — table owners bypass RLS by virtue of being table owners
- When forceRls is true, table owners do *not* automatically bypass RLS policies unless they have the `BYPASSRLS` perm
- In custom setups where additional users may be involved, or the central user isn't the user creating tables, additional care should be taken when customizing this
- The setting is true by default for more safety (even if in the most standard cases **there's no change in behavior regardless of how this setting is configured**)


tests/RLS/TableManagerTest.php
```diff
@@ -21,6 +21,7 @@
 use Stancl\Tenancy\Database\Exceptions\RecursiveRelationshipException;
 
 beforeEach(function () {
+    CreateUserWithRLSPolicies::$forceRls = true;
```
stancl, line 25: Also reset this in afterEach

stancl: Same thing in the other test files

tests/RLS/TraitManagerTest.php
```diff
@@ -262,10 +264,10 @@
...
     expect(fn () => DB::statement("INSERT INTO comments (text, post_id) VALUES ('third comment', {$post1->id})"))
         ->toThrow(QueryException::class);
-})->with([
-    true,
-    false
-]);
+})->with(
+    [true, false],
+    [true, false],
+);
```
stancl:
```suggestion
})->with(
    [true, false],
    [false, true],
);
```
or
```suggestion
})->with([true, false])->with([true, false]);
```

stancl: Hmm or is the syntax that it builds a matrix from those arrays? I see in another dataset it's used that way.

Probably change these to use two `with()` calls to make that more clear.

lukinovec: > Hmm or is the syntax that it builds a matrix from those arrays?

Yeah

tests/RLS/TableManagerTest.php
lukinovec: https://github.com/archtechx/tenancy/pull/1293/commits/a5c79d49c38d85ee26d3223fd89876309b8e27d3

tests/RLS/TraitManagerTest.php
lukinovec: Used the two `with()` calls

https://github.com/archtechx/tenancy/pull/1293/commits/1f41a8a863317879a53ea6fd939bca34ce970288

TheDeadCode:
Any news on this? <3

tests/RLS/TableManagerTest.php
```diff
@@ -534,6 +541,99 @@
...
+
+    pest()->artisan('tenants:rls');
+
+    [$order1, $order2] = [
+        Order::create(['name' => 'order1', 'tenant_id' => $tenant1->getTenantKey()]),
+        Order::create(['name' => 'order2', 'tenant_id' => $tenant2->getTenantKey()]),
+    ];
+
+    // The table owner should see all the records
+    expect(Order::all())->toHaveCount(2);
```
stancl: In this test, add a dataset for `$forceRls` so the difference is actually observed. Assuming `createPostgresUser()` creates a user without BYPASSRLS, the change should be simple.

lukinovec: Added the dataset here https://github.com/archtechx/tenancy/pull/1293/commits/b7dda966a731531de4c671fab2cd9291253021af (also fixed the failing test in the commit)

stancl (approved):

PR #1296, opened 2025-01-16
Allow overriding root URL in tests by default

lukinovec:
This PR allows overriding root URL in tests by default by changing the default value of `RootUrlBootstrapper::$rootUrlOverrideInTests` to `true`.


stancl:
Also add a todo@revisit so we don't forget to remove this if it becomes unnecessary

PR #1299, opened 2025-01-16
chore: mysql version in docker compose updated to v8

chinmaypurav:
#### Description
When contributing to the package, I faced several challenges on my MacBook Pro M3 chip due to ARM architecture limitations.

#### Why this should be added
MySQL 5.7 has been EOL already since Oct 2023 and has no supported ARM tags. We should upgrade the MySQL version to 8.

MySQL 8 does have ARM support.

I was able to run the tests successfully on the said docker image.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,7 @@ services:
         stdin_open: true
         tty: true
     mysql:
-        image: mysql:5.7
+        image: mysql:8
         environment:
             MYSQL_ROOT_PASSWORD: password
             MYSQL_DATABASE: main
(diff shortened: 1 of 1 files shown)


stancl:
> When contributing to the package, I faced several challenges on my MacBook Pro M3 chip due to ARM architecture limitations.

See CONTRIBUTING.md. There's a docker compose override for M-series chips.

chinmaypurav:
> > When contributing to the package, I faced several challenges on my MacBook Pro M3 chip due to ARM architecture limitations.
> 
> See CONTRIBUTING.md. There's a docker compose override for M-series chips.

Sorry, my bad; I missed that part as I went in the other direction to solve that problem. That also works fine.

stancl:
https://github.com/archtechx/tenancy/commit/7ce76298645e35a88a70e09bb5cfc4c78c2e8b66

PR #1300, opened 2025-01-17
[4.x] Improve id generators

stancl:
- [x] Add bigint generator
- [x] ~~Add an option to store hex bytes as raw bytes? Figure out how casts and queries would work here. UUIDs have a `uuid()` string type (in migration syntax) but are used as strings, I'm not sure if we could do the same with an arbitrary binary column~~
- [x] ~~Consider a snowflake generator~~

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -33,6 +33,7 @@ return [
          *
          * @see \Stancl\Tenancy\UniqueIdentifierGenerators\UUIDGenerator
          * @see \Stancl\Tenancy\UniqueIdentifierGenerators\RandomHexGenerator
+         * @see \Stancl\Tenancy\UniqueIdentifierGenerators\RandomIntGenerator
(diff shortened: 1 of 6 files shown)


stancl:
Looks like storing random raw bytes would be inconvenient because even if we add some Eloquent accessors and setters, `where()` clauses still need to be able to format the data and there are many cases where Eloquent may be bypassed. UUIDs seem unique in that DB engines can transparently convert between the string representation and the byte storage. So we won't be making those changes.

stancl:
Snowflake generator is not being added since the implementation is too app-specific and apps that do use snowflakes would likely want to use a consistent implementation for all models rather than Tenancy rolling its own.

stancl:
Not sure why pcov is not picking up the coverage here. The method is being executed in the test correctly.

PR #1308, opened 2025-02-04
fix: not using template/tenant hostname while creating database users

### LINKED ISSUE #1309: hostname isn't set in created db user while using PermissionControlledMySQLDatabaseManager
## The issue
database user created while using PermissionControlledMySQLDatabaseManager is given access to hostname '%' and it's not customizable and creates issues on some hosts which prevents the created database user from accessing the database.

## The details
this package allows mysql multi database tenancy with database user permissions using the PermissionControlledMySQLDatabaseManager::class database manager which creates a user with permissions on created tenant database, one of the best features of this package is how configurable is it. one of the configuration options is the connection details customizatiton using tenancy_db_ prefix while creating the tenant or using template_tenant_connection that's cool but the issue I'm facing happens because the key tenancy_db_host is never used while creating the database user, I created a PR fixing such issue but the author @stancl required creating an issue before submitting a PR so here we are. the PR that solves the issue is #1308 

### Steps to reproduce

comment out default mysql database manager in config tenancy.database.managers.mysql and use the PermissionControlledMySQLDatabaseManager one, then create a tenant like this Tenant::create(['tenancy_db_host' => 'localhost']);

### Expected behavior

the database user should be created with permissions on hostname 'localhost' not the default '%'

### Laravel version

^11.0

### stancl/tenancy version

^3.8

shaheenfawzy:
the `$hostname` variable was defined but never used while creating new database user in the PermissionControlledMySQLDatabaseManager and a default hostname of % was used which caused permission issues like database user can't access mysql using password on localhost 
closes #1309

### DIFF
--- a/src/TenantDatabaseManagers/PermissionControlledMySQLDatabaseManager.php
+++ b/src/TenantDatabaseManagers/PermissionControlledMySQLDatabaseManager.php
@@ -25,14 +25,14 @@ class PermissionControlledMySQLDatabaseManager extends MySQLDatabaseManager impl
         $hostname = $databaseConfig->connection()['host'];
         $password = $databaseConfig->getPassword();
 
-        $this->database()->statement("CREATE USER `{$username}`@`%` IDENTIFIED BY '{$password}'");
+        $this->database()->statement("CREATE USER `{$username}`@`{$hostname}` IDENTIFIED BY '{$password}'");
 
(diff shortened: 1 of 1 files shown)


stancl:
Before submitting a PR, please open an issue with clear steps for reproducing the issue you're trying to solve here.

PR #1321, opened 2025-02-25
[4.x] Laravel 12 support

stancl:
This PR adds Laravel 12 support and drops support for all other Laravel versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         include:
-        - laravel: "^11.0"
+        - laravel: "^12.0"
           php: "8.4"
 
     steps:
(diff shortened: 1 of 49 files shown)


Jacobtims:
@stancl Why did you drop support for Laravel 11?

stancl:
It is easier to update the average app to Laravel 12 than for us to maintain compatibility with both L11 and L12.

Jacobtims:
There are still a lot of other packages that don't support L12, Now I can't update my tenancy projects.

stancl:
You can lock the tenancy version to an older commit that supports Laravel 11 and only switch back to dev-master when you can update your entire app. We'll soon be pushing changes that will likely depend on changes made to Laravel 12 — it would make no sense to support Laravel 11.

Jacobtims:
When I lock it to an older commit, composer will still look at the latest `composer.json`. That doesn't work unfortunately...

PR #1324, opened 2025-02-27
fix: add explicit nullable type declarations for UserImpersonation parameters

mu-hanz:
This PR addresses deprecation warnings in UserImpersonation.php by adding explicit nullable type declarations.

Changes:
- Added explicit `?string` type for `$authGuard` parameter in bootstrap() method
- Added explicit `?int` type for `$ttl` parameter in makeResponse() method
- Updated PHPDoc to reflect the nullable type for $ttl

Fixes deprecation warnings:
- Implicitly marking parameter $authGuard as nullable is deprecated
- Implicitly marking parameter $ttl as nullable is deprecated

### DIFF
--- a/src/Features/UserImpersonation.php
+++ b/src/Features/UserImpersonation.php
@@ -18,7 +18,7 @@ class UserImpersonation implements Feature
 
     public function bootstrap(Tenancy $tenancy): void
     {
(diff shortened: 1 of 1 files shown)


stancl:
Thanks!

PR #1325, opened 2025-03-03
[4.x] Add Laravel 11 and PHP 8.3 support

SuperDJ:
Added Laravel 11 and PHP 8.3 support. Laravel 11 is still supported to [March 12th, 2026](https://laravel.com/docs/12.x/releases) and PHP 8.3 is still supported to [31 Dec 2027](https://www.php.net/supported-versions.php). I also think that going from php 8.0 to php 8.4 as a minimum is too big of a leap.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,12 +16,14 @@ jobs:
     strategy:
       matrix:
         include:
+        - laravel: "^11.0"
+          php: "8.3"
         - laravel: "^12.0"
           php: "8.4"
 
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


stancl:
In the failing tests you can see why Laravel 11 support was dropped. PHP 8.3 was dropped because we want to use PHP 8.4 features and there aren't any significant changes complicating upgrading to PHP 8.4.

SuperDJ:
@stancl 
>and there aren't any significant changes complicating upgrading to PHP 8.4.

there are if you don't have influence on the PHP version run on the server.

PR #1326, opened 2025-03-11
[3.x] Add comment to config file explaining --force with seeding in production

laundy:
cf. https://github.com/stancl/tenancy-docs/issues/292


stancl:
You can omit the change to `tests/DatabasePreparationTest.php`, the flags are passed through to the underlying seeder logic so that's essentially Laravel territory. I can merge the comment addition.

PR #1327, opened 2025-03-12
Invalidate the resolver cache when a tenant or related model is deleted

laundy:
Currently, whenever `$shouldCache` in a `CachedTenantResolver` is set to `true`, the cache is only invalidated on the `saved` event of the model. As a result, whenever a `Domain` or a `Tenant` is deleted, the cache is not invalidated and the domains are still accessible (i. e. not falling back to the `$onFail` logic or throwing the `TenantCouldNotBeIdentifiedException`) which may lead to unhandled errors (no database permissions etc.).

This PR makes sure the invalidation is also performed on a `delete` event of `Tenant` or the related models like `Domain`.

### DIFF
--- a/src/Database/Concerns/InvalidatesResolverCache.php
+++ b/src/Database/Concerns/InvalidatesResolverCache.php
@@ -19,7 +19,7 @@ trait InvalidatesResolverCache
     public static function bootInvalidatesResolverCache()
     {
         static::saved(fn(Tenant $tenant) => static::invalidateTenantCache($tenant));
-        static::deleted(fn(Tenant $tenant) => static::invalidateTenantCache($tenant));
+        static::deleting(fn(Tenant $tenant) => static::invalidateTenantCache($tenant));
     }
 
     private static function invalidateTenantCache(Tenant $tenant): void
--- a/src/Database/Concerns/InvalidatesResolverCache.php
+++ b/src/Database/Concerns/InvalidatesResolverCache.php
@@ -18,13 +18,17 @@ trait InvalidatesResolverCache
 
     public static function bootInvalidatesResolverCache()
     {
-        static::saved(function (Tenant $tenant) {
-            foreach (static::$resolvers as $resolver) {
-                /** @var CachedTenantResolver $resolver */
-                $resolver = app($resolver);
+        static::saved(fn(Tenant $tenant) => static::invalidateTenantCache($tenant));
+        static::deleted(fn(Tenant $tenant) => static::invalidateTenantCache($tenant));
+    }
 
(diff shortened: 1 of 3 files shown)


stancl:
Did you check if the tests actually test what they should? I commented out both `deleted` event listeners and everything is still passing for me.

laundy:
You are right, I'm not able to get it working yet. This is strange as the `deleted` event should be fired when running `$tenant->delete()` or `$tenant()->domains()->first()->delete()`. I will have a further look at it.
In general, there is a chance that the `deleted` events (and also all other events) are not fired when using Eloquent queries (cf.  https://laravel.com/docs/10.x/eloquent#events and https://laravel.com/docs/10.x/eloquent#deleting-models-using-queries). In my understanding, that means that the cache won't be invalidated properly e.g. when mass-editing or -deleting (but this is a bigger topic).

stancl:
> that means that the cache won't be invalidated properly e.g. when mass-editing or -deleting

Yes, events only fire when individual models are used.

laundy:
Found the issue (and it makes total sense): once the model is `deleted`, there is no chance that the `getArgsForTenant` method called in `invalidateCache` of `CachedTenantResolver` is able to get the cache key to invalidate the cache (as the `foreach` loop won't be iterated):
https://github.com/archtechx/tenancy/blob/19bc316cf67479d5902553aca825347184547f1e/src/Resolvers/Contracts/CachedTenantResolver.php#L57-L59

I've switched to the `deleting` event and now the tests also test what they should.

stancl:
The tests don't make much sense to me. You've inverted some assertions and expect a very general exception — so what **is** the expected behavior? The tests should just assert that clearly. Specifically that the *bug you're reporting here is fixed*, not necessarily implementation details. The tests should work as general regression tests where anything that fixes the reported bug makes them pass, ideally.

laundy:
I don't get your feedback tbh.

The test checks that the resolver cache is invalidated/cleared whenever a tenant or tenant domain is deleted. After a tenant was deleted, resolving the tenant will throw a `TenantCouldNotBeIdentifiedException` (if no custom `$onFail` logic is set which is default). This is the expected behavior and also what the test is asserting. In addition, the test checks if the resolver tries to use the DB (as this indicates that the cache has no hit) after deleting the tenant. This follows the same pattern as the `cache_is_invalidated_when_a_tenants_domain_is_changed`.

With the current implementation, the cache won't be invalidated and there would be no `TenantCouldNotBeIdentifiedException`thrown. This is why I've added the `deleting` listeners.

stancl:
Closing this in favor of #1328 and #1329, thanks for reporting this bug and the proposed fix!

PR #1330, opened 2025-03-13
nitpick: make the default tenant database name a bit more readable

CoolGoose:
Right now databases on a fresh install are setup like 

```tenanta0f50601-87fb-4f8e-bb18-3fb0b39a69bc```

added `_` just to make it a bit tidier 😅


stancl:
I'll leave this as optional config since this would break backwards compatibility and is arguably not what you want if you use different ID generators, like numeric IDs.

PR #1332, opened 2025-03-14
[4.x] Add ULIDGenerator

CoolGoose:
I didn't change the UUID implementation to just use the Str::uuid() , but that's also an option to remove the dependency on ramsey (and have it as an indirect via the laravel/support)


stancl:
illuminate/support only *suggests* ramsey/uuid I think

PR #1341, opened 2025-04-01
Check if the database exists.

solomon-ochepa:
This update introduces a method to verify whether a tenant's database exists.  

#### Usage:  
```php
$tenant->database()->exists();
```
Returns `true` if the tenant's database exists, otherwise `false`.  

#### Why is this necessary?  
In some cases, a tenant's database may be missing due to system migrations or failures.  
Attempting to access `$tenant->database` when the database doesn't exist results in an error:  

> `Modules\Tenancy\App\Models\Tenant::database must return a relationship instance.`  

This makes it difficult to perform conditional checks or debug issues effectively.

### DIFF
--- a/src/DatabaseConfig.php
+++ b/src/DatabaseConfig.php
@@ -163,4 +163,14 @@ class DatabaseConfig
 
         return $databaseManager;
     }
+
+    /**
+     * Check if the database exists.
+     *
+     * @return bool Returns true if the database exists, otherwise false.
+     */
+    public function exists(): bool
+    {
+        return $this->manager()->databaseExists($this->getName());
+    }
 }


stancl:
The linked issues don't seem related to this at all. Please don't unnecessarily ping people in other threads.

The PR doesn't describe why this change would be needed. I'm fairly sure the check can be achieved with the current public API easily.

solomon-ochepa:
> Allows developers to check if a database exists.

Sorry about the pinging, but a brief description of what it does was given. Please this function is important and useful.

solomon-ochepa:
> I'm fairly sure the check can be achieved with the current public API easily.

Please, can you specify the `public API`?

stancl:
**All** of the code you're calling here is public. No method needs to be added.

solomon-ochepa:
> **All** of the code you're calling here is public. No method needs to be added.

I've updated the description. Could you please review the PR again and consider merging it?  

This function is essential. I've already implemented it locally in our Tenant model, and I believe it will benefit the community.  

The implementation is simple and does not affect any existing functionality:  
```php
public function exists(): bool
{
    return $this->manager()->databaseExists($this->getName());
}
```  

This method significantly improves readability and usability.

Instead of:  
```php
$tenant->database()->manager()->databaseExists($tenant->database()->getName());
```  
Developers can simply use:  
```php
$tenant->database()->exists();
```  
which is much cleaner and easier to understand.

---
Please, I need you to rethink and understand that clean and readable code makes development enjoyable and your package more useful.

PR #1350, opened 2025-05-12
Feature: Base TestCase for Tenant-Aware Testing

lokeshrangani:
# Summary
This PR introduces a `TenantAwareTestCase` base class under the `Stancl\Tenancy\Testing` namespace to help developers write cleaner, tenant-aware tests with minimal boilerplate.

This PR provides an abstract base class that:

- Automatically initializes tenancy before each test (via TEST_TENANT)
- Allows easy switching to the central context using runInCentralContext()
- It is fully optional and does not affect existing behavior

# Usage
```
use Stancl\Tenancy\Testing\TenantAwareTestCase;

class OrdersTest extends TenantAwareTestCase
{
    public function test_orders_are_isolated_by_tenant()
    {
        $this->get('/api/orders')->assertStatus(200);
    }

    public function test_central_logic()
    {
        $this->runInCentralContext(function () {
            $this->assertDatabaseHas('users', ['email' => 'admin@example.com']);
        });
    }
}
```

# Features
1. Auto tenancy setup using .env.testing
5. $tenancy = false toggle for opt-out
6. Central DB testing support
7. 100% backward compatible
8. Fully optional — users can ignore it if not needed

Happy to rename, restructure, or adapt the PR based on your feedback or project goals. 😀

### DIFF
--- a/assets/config.php
+++ b/assets/config.php
@@ -196,4 +196,6 @@ return [
         '--class' => 'DatabaseSeeder', // root seeder class
         // '--force' => true, // This needs to be true to seed tenant databases in production
     ],
+
+    'test_tenant' => env('TEST_TENANT', 1),
 ];
--- a/src/Testing/TenantAwareTestCase.php
+++ b/src/Testing/TenantAwareTestCase.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace Stancl\Tenancy\Testing;
+
+use App\Models\Tenant;
+use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
+
+abstract class TenantAwareTestCase extends BaseTestCase
+{
+    use CreatesApplication;
+
+    /**
+     * Set this to true in child classes to enable tenancy.
+     *
+     * @var bool
+     */
+    protected bool $tenancy = true;
+
+    /**
+     * Tenant instance used in the test.
+     *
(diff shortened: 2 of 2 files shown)


stancl:
Thanks for the PR. The reason we haven't shipped a base TestCase in the package is that they're *very* app-specific and it's better to simply provide guidance on how to write one in our documentation.

As an example, in the new SaaS boilerplates that will launch with Tenancy v4, we use completely different logic in the test cases to support parallel testing, in-memory SQLite DBs, DB transactions, etc. In other versions of the boilerplate we use much simpler logic due to a difference in the DB setup.

There's no good one-size-fits-all here even as a base class to be extended imo. The only thing I'm planning on changing is improving the documentation here.

lokeshrangani:
Thanks for the thoughtful response — that makes a lot of sense.

You're right: test environments can vary wildly across projects (e.g., parallel tests, SQLite, custom tenant setups), and trying to enforce a base TestCase might limit flexibility or introduce assumptions.

I'm glad to hear documentation improvements are planned — if helpful, feel free to reuse any of the structure or helpers from this PR as examples for the docs. That way, developers who want a starting point can still benefit without the constraints of a package-level base class.

Appreciate the clarity, and excited to see what’s coming in v4 and the new boilerplates!

stancl:
Appreciate the effort here!

PR #1354, opened 2025-05-21
[4.x] TableRLSManager refactor, comment constraints

lukinovec:
## TableRLSManager refactor
The structure of TableRLSManager got updated. Now, the only public methods are
- `shortestPaths()` (generates the shortest, valid, preferably non-nullable path from table X to the tenants table, used in tests)
- `generateQueries()` (used by the `tenants:rls` command to generate RLS policies for tables, uses `shortestPaths()`)

### Performance
The path generation logic now utilizes dynamic programming (recursion + caching), so generating the paths from tables to the tenants table (and so generating the RLS policies) is now significantly faster, especially in apps with larger/complex DB schemas.

Shortest paths from table X to the tenants table are now cached during path generation. This prevents repeatedly generating already generated paths.

Paths that don't lead to the tenants table ("dead ends") are also cached so that we don't unnecessarily try to build paths that don't lead to the tenants table.

### Other improvements
Recursive relationships now behave as described in the docs. The RecursiveRelationshipException is only thrown if a recursive relationship was detected on the table AND there's no valid path from that table to the tenants table.

Comment logic (the "rls"/"no-rls" explicit/implicit scoping behavior) got corrected, the behavior is also clearer from the code now.
- "no-rls" comment on a column = the column won't be considered while generating a path from table X to the tenants table
- While `$scopeByDefault` is `false`, foreign key columns have to have an "rls" comment for paths leading through them to the tenants table to be considered during path generation (previously, the comment just had to be non-null)

## Comment constraints
Before, the TableRLSManager would completely ignore columns without constraints from policy generation. This PR makes it possible to add "fake" constraints to columns using comments. The columns now don't need to have real constraints for them to be included in RLS policy generation.

### Usage
The comment constraint can be defined like this:
`$table->string('tablename_id')->comment('rls tablename.id')`

The comment has to be formatted like `rls <foreign_table>.<foreign_id>`. If you format the comment incorrectly, you'll receive an exception while attempting to generate the paths (in a real app, while running `tenants:rls`).

Note that in relationship paths, you can combine real and comment constraints in any way.

### Real-world example
A comment constraint is a "fake" foreign key constraint that only serves for generating the RLS policies. This is useful if some of your migrations don't use real constraints, either for performance reasons, or for handling constraint edge cases in a simpler way.

The Jetstream's teams table for example (which appears to be the latter case):

```php
// The teams table
Schema::create('teams', function (Blueprint $table) {
    $table->id();
    $table->foreignId('user_id')->index(); // bigInteger column without a real constraint
    // Other columns..
});

// The users table
Schema::create('users', function (Blueprint $table) {
    $table->id();
    $table->foreignId('current_team_id')->nullable(); // bigInteger column without a real constraint
    $table->string('tenant_id');
    $table->foreign('tenant_id')->references('id')->on('tenants')
        ->onUpdate('cascade')->onDelete('cascade'); // Column with a real constraint
 });
```

This example intentionally doesn't use real constraints (because of the "chicken and egg" problem described below), therefore, the teams table **will be skipped** while generating the RLS policies using `php artisan tenants:rls`, and the data won't be properly separated per-tenant using the RLS policies.

The `user_id` and `current_team_id` columns can't have a normal constraint. If they had normal constraints (= the foreignId columns would have the `->constrained()` call):
- the users table would have to be created before the teams table because the teams table references the users table
- we wouldn't be able to create the users table because it also references the teams table

So we'd need to modify the migrations in a complex way for the tables to have real constraints. That could also alter the default behavior (not inherently incorrect, but assume it is for the sake of this example).

The ideal way to deal with this and keep the default behavior intact would be to give the `user_id`and `current_team_id` columns info about the constraint (= foreign table and foreign id) using comments for the data separation to work using RLS policies without defining actual constraints:

```php
// The teams table
Schema::create('teams', function (Blueprint $table) {
    $table->id();
    $table->foreignId('user_id')->index()->comment('rls users.id'); // Column with a comment constraint
    // Other columns..
});
```

Now, running `tenants:rls` will generate a RLS policy for the teams table (sidenote: also the `team_invitations` table, which has a constraint to the teams table on the `team_id` column), ensuring proper data separation.

### DIFF
--- a/src/Exceptions/RLSCommentConstraintException.php
+++ b/src/Exceptions/RLSCommentConstraintException.php
@@ -10,6 +10,6 @@ class RLSCommentConstraintException extends Exception
 {
     public function __construct(string|null $message = null)
     {
-        parent::__construct($message ?? "Invalid comment constraint.");
+        parent::__construct($message ?? 'Invalid comment constraint.');
     }
 }
--- a/src/Exceptions/RLSCommentConstraintException.php
+++ b/src/Exceptions/RLSCommentConstraintException.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Exceptions;
+
+use Exception;
+
+class RLSCommentConstraintException extends Exception
+{
+    public function __construct(string|null $message = null)
+    {
+        parent::__construct($message ?? "Invalid comment constraint.");
+    }
+}
--- a/tests/RLS/TableManagerTest.php
+++ b/tests/RLS/TableManagerTest.php
@@ -329,96 +329,6 @@ test('table rls manager generates relationship trees with tables related to the 
     /** @var TableRLSManager $manager */
     $manager = app(TableRLSManager::class);
 
-    $expectedTrees = [
-        'authors' => [
-            // Directly related to tenants
-            'tenant_id' => [
-                [
-                    [
-                        'foreignKey' => 'tenant_id',
-                        'foreignTable' => 'tenants',
-                        'foreignId' => 'id',
-                        'nullable' => false,
-                    ]
-                ],
-            ],
-        ],
-        'comments' => [
-            // Tree starting from the post_id foreign key
-            'post_id' => [
-                [
-                    [
-                        'foreignKey' => 'post_id',
-                        'foreignTable' => 'posts',
-                        'foreignId' => 'id',
-                        'nullable' => false,
-                    ],
-                    [
-                        'foreignKey' => 'author_id',
-                        'foreignTable' => 'authors',
-                        'foreignId' => 'id',
-                        'nullable' => false,
-                    ],
-                    [
-                        'foreignKey' => 'tenant_id',
-                        'foreignTable' => 'tenants',
-                        'foreignId' => 'id',
-                        'nullable' => false,
-                    ],
-                ],
-                [
-                    [
-                        'foreignKey' => 'post_id',
-                        'foreignTable' => 'posts',
-                        'foreignId' => 'id',
-                        'nullable' => false,
-                    ],
-                    [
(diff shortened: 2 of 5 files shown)


tests/RLS/TableManagerTest.php
```diff
@@ -766,13 +710,91 @@
...
+    $shortestPaths = app(TableRLSManager::class)->shortestPaths();
+
+    expect(array_keys($shortestPaths))
+        ->toContain('posts', 'comments')
+        // Shortest paths do not include the recursive_posts table
+        // because it has a 'no-rls' comment on its only foreign key
+        ->not()->toContain('recursive_posts');
 });
 
+test('table manager can generate paths leading through comment constraint columns', function() {
```
lukinovec, line 762:
```suggestion
test('table manager can generate paths leading through columns with comment constraint', function() {
```

Imagine we have `$table->string('tenant_id')->comment('rls tenants.id')` in a migration. `tenant_id` is a column with a comment constraint. "column with comment constraint" sounds verbose, "comment constraint column" sounds weird/vague (and still pretty verbose?)

I'd probably go with the former for clarity here.. What would you prefer @stancl?

src/Exceptions/RLSCommentConstraintException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Exceptions;
```
stancl: Move to `src/RLS/Exceptions`

lukinovec: Moved https://github.com/archtechx/tenancy/pull/1354/commits/8934114f92789265559372f3e9b3bae60982ecfc

PR #1360, opened 2025-06-02
[4.x] Test that route model binding works correctly with path identification

lukinovec:
This PR adds a test for checking if route model binding works with routes using path identification.

For example: `Route::get('/{tenant}/{user}', fn (User $user) => $user->name)->middleware([InitializeTenancyByPath::class, 'web']);`

In this route, tenancy will be initialized first, and then, the user will be resolved in the tenant context.

### DIFF
--- a/tests/PathIdentificationTest.php
+++ b/tests/PathIdentificationTest.php
@@ -258,7 +258,7 @@ test('any extra model column needs to be whitelisted', function () {
     pest()->get('/acme/foo')->assertSee($tenant->getTenantKey());
 });
 
-test('route model binding works with closure-based routes', function() {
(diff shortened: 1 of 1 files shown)


tests/PathIdentificationTest.php
```diff
@@ -246,3 +257,21 @@
     config(['tenancy.identification.resolvers.' . PathTenantResolver::class . '.allowed_extra_model_columns' => ['slug']]);
     pest()->get('/acme/foo')->assertSee($tenant->getTenantKey());
 });
+
+test('route model binding works with closure-based routes', function() {
```
lukinovec:
```suggestion
test('route model binding works with path identification', function() {
```

PR #1369, opened 2025-06-27
migrate-fresh can show output with verbose

lordofthebrain:
I would be very happy if migrate:fresh don't run silent by default. 
Silent is still possible with the silent option

### DIFF
--- a/src/Commands/MigrateFresh.php
+++ b/src/Commands/MigrateFresh.php
@@ -8,6 +8,7 @@ use Illuminate\Console\Command;
 use Stancl\Tenancy\Concerns\DealsWithMigrations;
 use Stancl\Tenancy\Concerns\HasATenantsOption;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\NullOutput;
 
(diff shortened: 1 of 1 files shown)


schmidti2002 (approved):

stancl:
Would need more info here

lordofthebrain:
With `callSilent` I have this output:
callSilent

With `call` I can see every single migration and the duration: (I have blacked out names):
call

Note: `php artisan tenants:migrate` don't run silent by default.

stancl:
The output is simplified when using `migrate-fresh` because it'd be too much output (compared to running regular migrations) and that output would be the exact same for all tenants. I'm probably fine with using `call()` instead of `callSilent()` when a higher verbosity level is used.

lordofthebrain:
Good point with the output:
I adjusted the code.
Now `php artisan tenants:migrate-fresh`  runs silent as before  
`php artisan tenants:migrate-fresh -v` shows all output

stancl:
Closing in favor of #1464. The new PR includes you as a co-author to keep the contribution credit. Thank you!

PR #1377, opened 2025-07-15
Update MailConfigBootstrapper.php

mepsd:
while running in queue:work previously it was not working made this changes now it is working fine


stancl:
Not enough information, nor an explanation of *why* this is needed or why it's implemented in this way

PR #1378, opened 2025-07-17
[4.x] Make SyncMaster's tenants relationship customizable

lukinovec:
Resource syncing only worked properly when the SyncMaster model had the `tenants()` relationship. Now, the SyncMaster model can use any name for the relationship, but it needs to be returned by the SyncMaster's `getTenantsRelationshipName()` method

In most cases, this method would return `tenants`, so now, a SyncMaster model would have these two extra methods :

```php
public function getTenantsRelationshipName(): string
{
        return 'tenants';
]


{
        return $this->morphToMany(config('tenancy.models.tenant'), 'tenant_resources', 'tenant_resources', 'resource_global_id', 'tenant_id', $this->getGlobalIdentifierKeyName())
            ->using(TenantMorphPivot::class);
}
```

Before, these methods weren't needed to be added manually: it was assumed that the tenants relationship method was 'tenants', and the method was added to the SyncMaster class by the ResourceSyncing trait.

### DIFF
--- a/src/ResourceSyncing/Listeners/DeleteResourcesInTenants.php
+++ b/src/ResourceSyncing/Listeners/DeleteResourcesInTenants.php
@@ -19,13 +19,15 @@ class DeleteResourcesInTenants extends QueueableListener
         $centralResource = $event->centralResource;
         $forceDelete = $event->forceDelete;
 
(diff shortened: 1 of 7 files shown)


stancl:
Since this makes the (already complex) code even more complex, I am going to close this. This PR being in history will be a helpful reference if we decide to implement this for v5, if enough people request this.

PR #1387, opened 2025-08-06
[4.x] Clean up expired impersonation tokens instead of just aborting, add command for cleaning up expired tokens

### LINKED ISSUE #1348: Some ImpersonationTokens are not cleaned up
### Bug description

If aborts happen in `UserImpersonation::makeResponse` the token records are never cleaned up.

### Steps to reproduce

Create an error state, or cause a timeout on redirects and trigger make response.

### Expected behavior

A method or job that cleans up old token.

### Laravel version

12.9.2

### stancl/tenancy version

3.9.1

### Ref.

https://github.com/archtechx/tenancy/blob/d98a170fbd2e114604bfec3bc6267a3d6e02dec1/src/Features/UserImpersonation.php#L43

lukinovec:
This PR makes the expired/invalid tenant impersonation tokens get deleted instead of just aborting with 403.

The PR also adds a command (ClearExpiredImpersonationTokens) used like `php artisan tenants:purge-impersonation-tokens`. As the name suggests, it clears all expired impersonation tokens (= tokens older than `UserImpersonation::$ttl`).

Resolves #1348


src/Commands/ClearExpiredImpersonationTokens.php
```diff
@@ -0,0 +1,41 @@
...
+namespace Stancl\Tenancy\Commands;
+
+use Illuminate\Console\Command;
+use Stancl\Tenancy\Features\UserImpersonation;
+
+/**
+ * This command clears expired impersonation tokens.
+ * By default, all tokens older than UserImpersonation::$ttl (60 seconds by default)
+ * are deleted. To override this, you can use the --ttl option, for example
+ * --ttl=120, all tokens older than 120 seconds will be deleted, ignoring the default.
```
stancl: Is there a use case for wanting to keep expired tokens?

lukinovec: I don't think so

lukinovec: Maybe the annotation should suggest that the command can not only clear _expired_ tokens, but also clear tokens older than the passed TTL.

lukinovec: Ended up removing the `--ttl` option

PR #1389, opened 2025-08-07
[4.x] Fix ViteBundler not affecting Vite static calls

### LINKED ISSUE #1388: ViteBundler doesn't affect Vite:: static calls when asset_helper_tenancy is true
### Bug description

When `asset_helper_override` is enabled, Laravel's asset() helper becomes tenant-aware, which breaks Vite asset resolution in both tenant and central contexts. Stancl's ViteBundler tries to solve this by binding a custom Vite class that overrides assetPath(), but this only affects container-based usage (like @vite('resources/images/logo.png') ), not static calls like Vite::asset() 
or Vite::anyMacroDefinedFunction() which are commonly used in Blade templates.

As a result, Vite may generate incorrect tenant-scoped asset URLs, even in the central context.

### Steps to reproduce

Instead of using the standard Blade directive:
### 🐛 Steps to Reproduce

1. Instead of using the standard Blade directive:

```blade
   @vite('resources/images/logo/logo.png')
```

2. Use one of the following alternatives:

A. 
```blade
Vite::asset('resources/images/logo/logo.png')
```

B. using a custom Vite::image() macro
```blade
Vite::image('logo/logo.png')
```
Where the macro is defined as: 
```blade
Vite::macro('image', fn (string $asset) => $this->asset("resources/images/{$asset}"));
```

3. Try this inside both tenant and central contexts.

### Expected behavior

To make the ViteBundler works for both container and static usage.

### My purposed fix is to:-
Update ViteBundler to replace the singleton binding with a call to createAssetPathsUsing() method on the Vite facade. This ensures the asset resolver is applied to both container and static usage.

```php
// Updated  ViteBundler bootstrap method
    public function bootstrap(Tenancy $tenancy): void
    {
        \Illuminate\Support\Facades\Vite::createAssetPathsUsing(function ($path, $secure = null) {
            return global_asset($path);
        });
    }
```

### Laravel version

12.20.0

### stancl/tenancy version

dev-master

farishrf:
Replace custom Vite class override with Vite::createAssetPathsUsing() to ensure ViteBundler works for both container and static usage when asset_helper_override is enabled.

Fixes #1388


tests/Features/ViteBundlerTest.php
```diff
@@ -3,27 +3,134 @@
...
+    $viteAssetUrl = app(Vite::class)->asset($this->assetPath);
+    $expectedGlobalUrl = global_asset('build/assets/foo-AbC123.js');
+
+    expect($viteAssetUrl)
+        ->toBe($expectedGlobalUrl)
+        ->and($viteAssetUrl)
+        ->not->toBe(route('stancl.tenancy.asset', ['path' => 'build/assets/foo-AbC123.js']));
+});
+
+test('vite uses tenant assets when asset_helper_override is enabled without ViteBundler', function () {
```
stancl: Is this ever desired behavior or does this just establish that this undesired behavior occurs with a certain configuration (= without ViteBundler)

tests/Features/ViteBundlerTest.php
```diff
@@ -3,27 +3,134 @@
...
+
+    $viteAssetUrl = app(Vite::class)->asset($this->assetPath);
+
+    expect($viteAssetUrl)
+        ->toBe(route('stancl.tenancy.asset', ['path' => 'build/assets/foo-AbC123.js']))
+        ->and($viteAssetUrl)
+        ->not->toBe(global_asset('build/assets/foo-AbC123.js'));
+});
+
+test('vite asset helper works correctly with path identification', function (bool $kernelIdentification) {
```
stancl: Does path identification affect the `global_asset()` / Vite logic?

tests/Features/ViteBundlerTest.php
farishrf: Yes this is the undesired behavior that occurs **without** `ViteBundler`.

When `asset_helper_override` is enabled but `ViteBundler` is not used, in a **tenant** context, any Vite asset URL will be generated as follows:

- **With `ASSET_URL` set:**
```
https://project.test/tenant7a63c1a7-a9e2-4832-a630-fab302853fad/build/assets/favicon-CkiKZZzN.svg
```

- **Without `ASSET_URL` set:**
```
http://XXX.project.test/tenancy/assets/build/assets/favicon-CkiKZZzN.svg
```

If you enable `ViteBundler` **and** set `asset_helper_override` to `true`, Vite correctly generates global asset URLs:
```
http://XXX.project.test/build/assets/favicon-CkiKZZzN.svg
```
(This pulls directly from the global assets, which is exactly the purpose of `ViteBundler` to generate correct asset paths)

If you **never** enable `asset_helper_override`, Vite works fine without the need for `ViteBundler`.

tests/Features/ViteBundlerTest.php
farishrf: No , it doesn't affect the global_asset() / Vite Logic

This test just verifies that ViteBundler works consistently regardless of when tenancy gets initialized ( route -level , kernel)
it's not critical to our main PR - we could delete it if you prefer, it just might provides some regression protection just in case,  since we have lots of interaction between the components.

tests/Features/ViteBundlerTest.php
stancl: Removed the test, we don't need to assert incorrect behavior

tests/Features/ViteBundlerTest.php
stancl: Also removed this test since there's no need to test this

stancl:
Removed unnecessary tests and simplified the test file overall. Some main points:
- We don't use properties like `$this->tenant`
- There's no need to actually create the file - when writing tests no logic should be added unless it's needed
- Same thing for a lot of the static property config - it just isn't needed, therefore it shouldn't be added
- Simplified some assertions in the one test that I kept

Will merge once tests pass, thanks!

stancl:
Seems the file actually does need to exist (it's created persistently so you don't notice this locally), re-added that logic

PR #1390, opened 2025-08-07
[4.x] Make TenancyUrlGenerator inherit the original UrlGenerator's scheme (http or https)

lukinovec:
Before, when using UrlGeneratorBootstrapper, and your app had a `https://` url, in tenant context, the url would have the `http://` scheme.

Now, the bootstrapper makes sure that the TenancyUrlGenerator inherits the original UrlGenerator's scheme. So if your app has e.g. url "https://some-url.test", `route('home')` in tenant context will return "http**s**://some-url.test/home" (originally, you'd get "http://some-url.test/home" - the original scheme - https - wouldn't be respected in the tenant context).

This PR addresses the issue reported on Discord (https://discord.com/channels/976506366502006874/976506736120823909/1399012794514411621).


src/Bootstrappers/UrlGeneratorBootstrapper.php
```diff
@@ -78,6 +79,10 @@ protected function useTenancyUrlGenerator(Tenant $tenant): void
             }
         }
 
+        // Inherit scheme (http/https) from the original generator
+        $originalScheme = Str::before($this->originalUrlGenerator->formatScheme(), '://');
+        $newGenerator->forceScheme($originalScheme);
```
stancl, line 84: What does `formatScheme()` return if no scheme was forced? Is it just `http://`?

Wondering if we can detect if a scheme was forced so we do not force a scheme if a scheme was not forced on the original generator.

stancl: Tests should also probably first start with a case where the scheme was not forced.

lukinovec: > What does formatScheme() return if no scheme was forced? Is it just http://?

Yeah, `app('url')->formatScheme()` returns "http://"

```php
public function formatScheme($secure = null)
{
        if (! is_null($secure)) {
            return $secure ? 'https://' : 'http://';
        }

        if (is_null($this->cachedScheme)) {
            $this->cachedScheme = $this->forceScheme ?: $this->request->getScheme().'://';
        }

        return $this->cachedScheme;
}
```

PR #1393, opened 2025-08-13
[4.x] Add tenant parameter BEFORE existing prefixes by default

lukinovec:
Originally, cloning a route would add the tenant parameter **after** all existing route prefixes. Now, the clone action has the `$tenantParameterBeforePrefix` property, which is `true` by default, meaning the tenant parameter gets added **before** the existing prefixes. This can be customized by calling `$cloneAction->tenantParameterBeforePrefix(false)` -- that way, the original behavior would be preserved.

For example, this solves the problem with Filament, where the cloned routes would look like `/filament/{tenant}/something`, which isn't ideal for many reasons (main reason is that the routes would be completely broken and the admin panel routes wouldn't work correctly. The other reason is that for consistency, tenant routes should have the `{tenant}` parameter **before** the existing prefixes (but as I mentioned, this is configurable).


stancl:
Related history https://github.com/tenancy-for-laravel/v4/pull/49

stancl:
And slightly less related https://github.com/archtechx/tenancy/commit/0f7cd2e8680a64bcf1f4ac9fd14087ae242ecdb0

stancl (approved):

PR #1402, opened 2025-10-01
[4.x] Filesystem bootstrapper: scoped disk support

### LINKED ISSUE #1401: Edge case for scoped filesystem driver for a tenant bootstrapped disk
### Bug description

When using scoped disks/driver from a disk that has been defined in the `tenancy.filesystem.disks` config. The resolved instances of those disks are not forgotten when initializing or reverting the `FilesystemTenantBootstrapper` and therefore persisting the previous tenant path prefix withing the application lifecycle.

### Steps to reproduce

1. Define a tenant disk in `config/tenancy.php`
```php
return [
    // ...
    'filesystem' => [
        'disks' => ['local'],
    ],
    // ...
];
```

2. Define a scoped disk in `config/filesystem.php`
```php
return [
    // ...
    'disks' => [
        'tenant-public' => [
            'driver' => 'scoped',
            'disk' => 'local',
            'prefix' => 'app/public',
            'visibility' => 'public',
        ],
    ],
    // ...
];
```

3. Create two tenants
4. Initialize one tenant and request the path using the Storage facade
5. Initialize the second tenant and request the path using the Storage facade

### Expected behavior

Disks scoped to a tenant bootstrapped disk should also be forgotten on tenant initialization

### Laravel version

11.46.0

### stancl/tenancy version

3.9.1

viicslen:
Fixes #1401


stancl:
If I'm reading the proposed change correctly, it basically looks at `filesystem.disks` and also scopes any disk *whose `['disk']` matches a disk in tenancy.filesystem.disks*?

In other words, instead of this you could simply add those scoped disks to `tenancy.filesystem.disks`, and this change only makes it so that you don't have to list each scoped variant of a disk by hand, instead it detects scoped versions automatically?

viicslen:
Sorry for the confusion I used the term `scope` and `bootstrap` without really providing any context
- Scope: Scoped Disk driver from Laravel, not related to the package (https://laravel.com/docs/12.x/filesystem#scoped-and-read-only-filesystems)
- Bootstrapped: Disks defined in `tenancy.filesystem.disks` in which the tenant is prefixed to the root path

> If I'm reading the proposed change correctly, it basically looks at filesystem.disks and also scopes any disk whose ['disk'] matches a disk in tenancy.filesystem.disks?

It doesn't really bootstrap the scoped disk, it just forgets the already resolved disk instance so that the prefix can be recalculated again from the parent disk.

> In other words, instead of this you could simply add those scoped disks to tenancy.filesystem.disks, and this change only makes it so that you don't have to list each scoped variant of a disk by hand, instead it detects scoped versions automatically?

I thought it would add the prefix twice but after doing a quick test it does solve the issue as well. Although it does go through the process of assigning a root to the scoped disk which is not really used by the driver, we just need to forget the resolved instance

PR #1411, opened 2025-11-04
[4.x] Improve resource syncing (refactor + mapping cleanup + morph maps)

lukinovec:
This PR adds listeners for handling cleanup of pivot records/mappings
- When any synced resource got deleted - on `SyncedResourceDeleted` (fired when any synced resource is deleted), `DeleteResourceMapping`. Previously, the cleanup was only happening on `SyncMaster`/central resource deletion. Now, `triggerDeleteEvent` is called while deleting/force deleting **any** synced resource, making the cleanup possible even for tenant resources.
- when a tenant gets deleted - on `TenantDeleted`, `DeleteAllTenantMappings` -- commented out by default (until now, the pivot records were only getting deleted when the tenant key column had a constraint, and the constraint had `onDelete('cascade')`-- db-level)

This PR also improves some resource syncing implementation details, for example, the `ResourceSyncing` trait now uses `static::deleted()` instead of `static::deleting()`, which is safer. The `SyncedResourceDeleted` event with the `DeleteResourceMapping` listener also lets us separate the resource x pivot record deletion concerns, and improve the performance (e.g. the mass delete of pivot records when `SyncMaster`'s deleted instead of a for loop where we were deleting the pivot records one by one).

The `PivotWithRelation` interface got simplified and renamed to `PivotWithCentralResource`. The interface now only requires the `getCentralResourceClass(): string` method that should return the central resource's class name (instead of a model instance).

We also move global ID generation to a separate method to make it easily overridable, in case a developer wants to use different UniqueIdentifierGenerator than the one used for tenant IDs.

We also add support for morph maps.


stancl:
We should also support using a `UniqueIdentifierGenerator` distinct from the one used for tenant key generation.

src/ResourceSyncing/Listeners/DeleteResourceMapping.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        // Delete pivot records if the central resource doesn't use soft deletes
+        // or the central resource was deleted using forceDelete()
+        if ($event->forceDelete || ! in_array(SoftDeletes::class, class_uses_recursive($centralResource::class), true)) {
+            Pivot::withoutEvents(function () use ($centralResource, $event) {
+                // $event->tenant is null when the deleted resource is a SyncMaster - all mappings are deleted in that case
+                // When $event->tenant is not null (= a Syncable was deleted), only delete the mapping for that tenant
+                $centralResource->tenants()->detach($event->tenant);
+            });
```
stancl, line 40: It is not entirely clear how this exact code achieves the stated:
```
 * If a SyncMaster (central resource) is deleted, all pivot records for that resource are deleted.
 * If a Syncable (tenant resource) is deleted, only delete the pivot record for that tenant.
```

Is it just a matter of how `detach()` behaves if no argument is passed (i.e. if `$event->tenant`) is null? If so, should the comment say something like:
> If detach() is called with null -- if $event->tenant is null -- this means a central resource was deleted and detaches all tenants.
> If detach() is called with a specific tenant, it means the resource was deleted in that tenant, and we only delete that single mapping.

That seems to be what the existing comments say, but I'd specifically frame this in the context of how `detach()` behaves if an argument is passed vs no argument (or null) is passed.

lukinovec: > Is it just a matter of how detach() behaves if no argument is passed (i.e. if $event->tenant) is null?

Yeah, exactly

src/ResourceSyncing/ResourceSyncing.php
```diff
@@ -19,36 +20,33 @@ trait ResourceSyncing
...
+            static::forceDeleting(static function (Syncable&Model $model) {
+                if ($model->shouldSync()) {
                     $model->triggerDeleteEvent(true);
                 }
             });
 
-            static::restoring(function (Syncable&Model $model) {
-                if ($model->shouldSync() && $model instanceof SyncMaster) {
+            static::restoring(static function (Syncable&Model $model) {
+                if ($model->shouldSync()) {
```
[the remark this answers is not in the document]
lukinovec: Resolved in https://github.com/archtechx/tenancy/pull/1411/commits/2b45a01018b10df09281a07b4b1c8f90a44b8359 (moved triggerRestoredEvent() to the Syncable interface)

src/ResourceSyncing/Listeners/DeleteResourceMapping.php
lukinovec: Resolved in https://github.com/archtechx/tenancy/pull/1411/commits/2b45a01018b10df09281a07b4b1c8f90a44b8359 (updated the comments above the `detach()` call as you suggested)

PR #1413, opened 2025-11-07
[4.x] Fix dropRLSPolicies()

imhayatunnabi:
`?` parameters are not supported in these statements, so we have to use string interpolation like in other related code.

### DIFF
--- a/src/Concerns/ManagesRLSPolicies.php
+++ b/src/Concerns/ManagesRLSPolicies.php
@@ -17,18 +17,28 @@ trait ManagesRLSPolicies
     {
         return array_map(
             fn (stdClass $policy) => $policy->policyname,
-            DB::select("SELECT policyname FROM pg_policies WHERE tablename = '{$table}' AND policyname LIKE '%_rls_policy%'")
+            DB::select(
+                "SELECT policyname FROM pg_policies WHERE tablename = ? AND policyname LIKE ?",
+                [$table, '%_rls_policy%']
+            )
         );
     }
 
     public static function dropRLSPolicies(string $table): int
     {
         $policies = static::getRLSPolicies($table);
+        $quotedTable = static::quoteIdentifier($table);
 
         foreach ($policies as $policy) {
-            DB::statement('DROP POLICY ? ON ?', [$policy, $table]);
+            $quotedPolicy = static::quoteIdentifier($policy);
+            DB::statement("DROP POLICY {$quotedPolicy} ON {$quotedTable}");
         }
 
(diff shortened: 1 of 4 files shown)


stancl:
Please provide more context for what this is actually fixing. Need a clear case that was broken previously.

imhayatunnabi:
Sure @stancl  ! The issues I found were, Basically two issues.
  1. When the system creates a new tenant, it also creates a tenant-specific database like multi database system per tenant. I found that if a tenant’s ID contained unusual characters like quotes or backticks or any other special characters supported with sql, the database creation could break or—worse which will let someone sneak in a harmful command. And now the proposed thing is to clean and safely wrap every tenant ID before using it in those database commands, so it can’t be misinterpreted or misused.
  2. There’s also a tool that’s supposed to remove old security policies inside PostgreSQL when they’re no longer needed. The way it was written, those policies were never actually deleted it silently failed every time. I have updated the commands so PostgreSQL accepts them, and policies now get removed as intended.
  
  The thing is to filter and not pass the value directly to the sql. pass them as sql as sql standard format which will automatically by sql behaviour the will be checked

stancl:
1. This is the case across our database managers. We make it clear tenant IDs should never come from user input as they get directly embedded into these SQL statements.
2. Would need to see a failing case for that — this should be covered by tests.

stancl:
If there is a genuine bug in 2), you can start by submitting a failing test. The other changes should be reverted.

imhayatunnabi:
@stancl , I added a quick test to show the exact bug we’re fixing. Right now, when you ask Tenancy to remove PostgreSQL RLS policies, nothing actually happens because Postgres doesn’t accept statements like DROP POLICY ? ON ?. The test creates a dummy policy (comments_rls_policy_manual), confirms it exists, calls Tenancy::dropRLSPolicies('comments'), and then checks if the policy is gone. Before this fix the test stays red because the policy never disappears; after quoting the identifiers properly the test goes green, proving the cleanup finally works.

stancl:
Thank you. I see what you're suggesting now — the `dropRLSPolicies` is not actually covered by tests and indeed may be incorrect as it uses `?` parameters which tend to not work in these contexts. Will take a look.

imhayatunnabi:
Great @stancl , Thanks for taking another look! I’m happy to adjust anything else you spot.

stancl:
For some additional context, at least in tests, the `dropRLSPolicies()` method wasn't failing silently but throwing a SQL syntax exception.

Issue #1414, opened 2025-11-08
[BUG] Cannot inject `Domain` contract for single-domain tenants


imhayatunnabi:
### Bug description

Controllers, listeners, or jobs can normally inject `Stancl\Tenancy\Contracts\Domain` to learn which hostname triggered the request. That works for multi-domain tenants (where domains live in a separate table), but it always returns `null` when the app uses `SingleDomainTenant`. The resolver simply never sets `DomainTenantResolver::$currentDomain` in that mode, yet the service provider still binds the `Domain` contract to that value. The end result: any code that relies on `Domain` suddenly loses access to the hostname the moment the project opts into single-domain tenants—even though the interface is advertised as available.

### Steps to reproduce

1. Make your tenant model implement `SingleDomainTenant` (the documented approach when each tenant has exactly one domain stored on the tenants table).  
2. Inject `Domain $domain` into a controller or job.  
3. Hit that controller under a tenant domain.

### Expected behavior

The created `$domain` should contain at least the hostname that was resolved (similar to how it behaves when domains are stored via the `Domain` model), so templates and services can show “You are on acme.example.com”.

But here the actual behavious I am seeing is, the container resolves `null` because `DomainTenantResolver::setCurrentDomain()` is skipped for `SingleDomainTenant`. See `src/TenancyServiceProvider.php:34-44` (binding) and `src/Resolvers/DomainTenantResolver.php:69-74` (early return).

### Laravel version

12.0.3

### stancl/tenancy version

4.x

stancl:
This should be on 4.x, not 3.9.1, right?

imhayatunnabi:
In my project it is currently at 3.9.1. But targeted fix for release version is 4.x. @stancl

stancl:
I'm saying you're reporting that this bug occurs in 3.9.1 in your issue but 3.x doesn't have `SingleDomainTenant`.

imhayatunnabi:
The issue occurs from master branch install. so it will be 4.x my bad.

stancl:
See my comment in the related PR.

PR #1415, opened 2025-11-08
Fix Domain injection for single-domain tenants

imhayatunnabi:
Issue: https://github.com/archtechx/tenancy/issues/1414

Description and the necessary for the single domain tenants in the domain injection:
- Ensure `DomainTenantResolver` always populates `DomainTenantResolver::$currentDomain`, even when the tenant implements `SingleDomainTenant`, by synthesizing a domain model that carries the resolved hostname and tenant key.
- Add a regression test (`tests/SingleDomainTenantTest.php`) that registers a controller injecting `Domain` and proves it now receives the hostname for single-domain tenants.

The reason it is necessary because:
- Apps that use `SingleDomainTenant` (domain stored directly on the tenants table) lost access to the `Domain` contract entirely. Any controller/job that relied on `Domain $domain` suddenly received `null`, breaking features as simple as “You’re messaging us from **acme.example.com**”.
- The service provider advertises the contract binding, and multi-domain setups work perfectly; having it silently fail in the single-domain configuration was confusing and forced teams to re-implement domain parsing themselves.

### DIFF
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -69,12 +69,21 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
     protected static function setCurrentDomain(Tenant $tenant, string $domain): void
     {
         /** @var Tenant&Model $tenant */
+        $domainModelClass = config('tenancy.models.domain');
+
         if ($tenant instanceof SingleDomainTenant) {
-            $domainModelClass = config('tenancy.models.domain');
-            /** @var Domain&Model $domainModel */
+            /**
+             * Single-domain tenants keep their hostname on the tenant row, so there's no Domain record
+             * we could look up. Build an in-memory instance that carries enough data for the Domain
+             * contract binding (hostname + tenant relation).
+             *
+             * @var Domain&Model $domainModel
+             */
             $domainModel = new $domainModelClass;
-            $domainModel->setAttribute('domain', $domain);
-            $domainModel->setAttribute(Tenancy::tenantKeyColumn(), $tenant->getTenantKey());
+            $domainModel->forceFill([
+                'domain' => $domain,
+                Tenancy::tenantKeyColumn() => $tenant->getTenantKey(),
+            ]);
             $domainModel->setRelation('tenant', $tenant);
 
             static::$currentDomain = $domainModel;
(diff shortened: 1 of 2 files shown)


src/Resolvers/DomainTenantResolver.php
```diff
@@ -69,9 +69,20 @@ public function resolved(Tenant $tenant, mixed ...$args): void
...
         /** @var Tenant&Model $tenant */
-        if (! $tenant instanceof SingleDomainTenant) {
-            static::$currentDomain = $tenant->domains->where('domain', $domain)->first();
+        if ($tenant instanceof SingleDomainTenant) {
+            $domainModelClass = config('tenancy.models.domain');
+            /** @var Domain&Model $domainModel */
+            $domainModel = new $domainModelClass;
+            $domainModel->setAttribute('domain', $domain);
+            $domainModel->setAttribute(Tenancy::tenantKeyColumn(), $tenant->getTenantKey());
+            $domainModel->setRelation('tenant', $tenant);
```
stancl, line 87: Using a fake `Domain` instance that only has `tenant_id` set, while querying the proper model instance below, doesn't make much sense.

src/Resolvers/DomainTenantResolver.php
```diff
@@ -69,9 +69,20 @@ public function resolved(Tenant $tenant, mixed ...$args): void
...
+
+            static::$currentDomain = $domainModel;
+
+            return;
         }
+
+        static::$currentDomain = $tenant->domains->where('domain', $domain)->first();
```
stancl: A `SingleDomainTenant` wouldn't necessarily have a hasMany "domains" relation.

imhayatunnabi:
Single-domain tenants now synthesize a Domain instance using forceFill() plus the tenant relation so the **Domain contract** binding has the hostname + tenant context even though no domain row exists.
Multidomain tenants load the domains relation on demand and fall back to a direct domain query if the collection didn’t contain the requested hostname, so we always persist real Domain models when they exist

stancl:
Looking at the code again, should the Domain instance *even exist*?
```php
/**
 * @property string|null $domain
 */
interface SingleDomainTenant extends Tenant {}
```

The `$domain` property should be a string — not a domain instance.

`SingleDomainTenantTest` also doesn't use Domain instances.

stancl:
The new diff adds quite a bit of complexity. I think at this time I'd just say using `Domain` bindings with single domain tenancy doesn't make much sense so it is not supported. Single domain tenants are fundamentally different architecturally, so it's not meant to be a switch you can turn on or off with existing code that injects `Domain` contracts.

Code that uses single-domain tenants should just use `tenant()->domain` to get the *string* domain (or just use `request()->getHost()`-based logic) — there are **no Domain instances involved.** If you do need an abstraction layer that always provides a `Domain`, you can write a simple custom helper that returns `app(Domain::class) ?? ... your fake model instance`.

imhayatunnabi:
Working on the solution. Will get back to you @stancl

Issue #1398, opened 2025-09-14
Problem with InitializeTenancyByDomainOrSubdomain when domain is similar to central domain


jakobjp:
### Bug description

Case using InitializeTenancyByDomainOrSubdomain

Central domain: **tms.test**
Tenant 1 Subdomain: **test1.tms.test**
Tenant 1 Domain: **test1-tms.test** (Does not work)
Alternative Tenant 1 Domain: **test1-tenant.test** (Does work)

Conclusion: The tenant 1 domain **test1-tms.test** is very similar to the central domain **tms.test**. Could this be the reason why the **test1-tms.test** is not identified?

<img width="451" height="100" alt="Image" src="https://github.com/user-attachments/assets/0e594010-779c-4625-8567-b6b328377844" />
<img width="1288" height="109" alt="Image" src="https://github.com/user-attachments/assets/6c6fe0ae-254d-48c1-818c-fd69e83a6574" />
<img width="1289" height="109" alt="Image" src="https://github.com/user-attachments/assets/9e570b0e-7f34-4902-9b85-8276a6a6bd9f" />
<img width="1283" height="435" alt="Image" src="https://github.com/user-attachments/assets/f8318946-ff7a-4928-a909-a4293b5067f4" />

### Steps to reproduce

Fresh Laravel installation with Tenancy. Followed the Quick Start Guide (using Sail). And setup one tenant with domains like described above.

### Expected behavior

I would expect any subdomain or domain to be identified. Also if it is similar to the central domain.

### Laravel version

12

### stancl/tenancy version

3

kostamilorava:
I have tried the exact same setup and `http://test1-tms.test/` is identified as it should be. Can you share the code?

jakobjp:
> I have tried the exact same setup and `http://test1-tms.test/` is identified as it should be. Can you share the code?

Okay, I have added it in this repo: https://github.com/jakobjp/laravel-tenancy-domains-test

stancl:
#1425

PR #1423, opened 2025-12-11
Fix incorrect subdomain identification

kostamilorava:
After reviewing issue #1398 the code for subdomain initialization missed few cases.

For example, `test1-tms.test` is not a subdomain, but original `Str::endsWith` will always return true and identify it as subdomain.

If this solution is ok, I will add test cases.

### DIFF
--- a/src/Resolvers/DomainTenantResolver.php
+++ b/src/Resolvers/DomainTenantResolver.php
@@ -56,9 +56,19 @@ class DomainTenantResolver extends Contracts\CachedTenantResolver
         return $tenant;
     }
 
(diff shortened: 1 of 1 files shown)


stancl:
I think when we tried to reproduce that issue, we could reproduce it in 3.x but not 4.x. Would need a regression test either way to merge anything.

kostamilorava:
It's relatively easy to reproduce. Just attach central domain `gov.test`, which is acceptable domain for central. If you will ever need to attach `tenantgov.test` to any tenant, it will incorrectly identify as a subdomain (as it ends with `gov.test`).

PR #1427, opened 2025-12-29
[4x] Remove redundant universal route check from PreventAccess MW

### LINKED ISSUE #1418: is PreventAccessFromUnwantedDomains useful when InitializeTenancyByRequestData is used?
Hi i'm starting to use your project v4 and I stumbled upon a couple of inconsistencies, or maybe I'm just misunderstanding something.. Let me know if you'd welcome a PR to address my uncertainties.

Why would we want to high-prio init `PreventAccessFromUnwantedDomains` if we're only using `InitializeTenancyByRequestData`?

https://github.com/archtechx/tenancy/blob/0cd0bc44b1c0bdb41d26b7e902755d2daaa4b795/assets/TenancyServiceProvider.stub.php#L252

It also seems `UsableWithEarlyIdentification->shouldBeSkipped()` is checking for `Tenancy::routeIsUniversal($route)`, so here the second condition is not necessary?
https://github.com/archtechx/tenancy/blob/0cd0bc44b1c0bdb41d26b7e902755d2daaa4b795/src/Middleware/PreventAccessFromUnwantedDomains.php#L34

lukinovec:
The PreventAcessFromUnwantedDomains MW had the `tenancy()->routeIsUniversal($route)` check either for returning early, or it was a leftover from some older implementation, so I removed it.

The middleware aborts if the `$this->accessingTenantRouteFromCentralDomain($request, $route) || $this->accessingCentralRouteFromTenantDomain($request, $route)` check passes. Meaning, **for the middleware to abort, the route has to be either in central or tenant mode**. When the route is in universal mode, the middleware will never reach `return $abortRequest()`. `return $next($request)` will always get reached, even when the `|| tenancy()->routeIsUniversal($route)` check is deleted from the previous condition, so that check was basically useless.

Since the docblock for the class does mention the behavior for universal routes explicitly, we've instead added a comment documenting that things work this way. That's probably the most reasonable way to have this explicit behavior for universal routes easily understandable in this fairly complex logic without redundant code.

Resolves #1418

## Summary by CodeRabbit

* **Bug Fixes**
  * Domain access checks are now applied consistently, including routes that were previously skipped, ensuring tenant vs. central domain restrictions are enforced and preventing unauthorized access that could bypass protections.

[Review Change Stack](https://app.coderabbit.ai/change-stack/archtechx/tenancy/pull/1427)

### DIFF
--- a/src/Middleware/PreventAccessFromUnwantedDomains.php
+++ b/src/Middleware/PreventAccessFromUnwantedDomains.php
@@ -31,7 +31,7 @@ class PreventAccessFromUnwantedDomains
     {
         $route = tenancy()->getRoute($request);
 
-        if ($this->shouldBeSkipped($route) || tenancy()->routeIsUniversal($route)) {
+        if ($this->shouldBeSkipped($route)) {
             return $next($request);
         }
 
--- a/src/Middleware/PreventAccessFromUnwantedDomains.php
+++ b/src/Middleware/PreventAccessFromUnwantedDomains.php
@@ -35,6 +35,8 @@ class PreventAccessFromUnwantedDomains
             return $next($request);
         }
 
+        // If the route is universal, neither of these checks will pass and the logic will
+        // fall through to the $next($request) call at the end.
         if ($this->accessingTenantRouteFromCentralDomain($request, $route) || $this->accessingCentralRouteFromTenantDomain($request, $route)) {
             $abortRequest = static::$abortRequest ?? function () {
                 abort(404);


src/Middleware/PreventAccessFromUnwantedDomains.php
lukinovec, line 15: Keeping the docblock like this may be a bit confusing because now, skipping this MW for universal routes is not as explicit. Instead of the explicit condition, this is implied (`if ($this->accessingTenantRouteFromCentralDomain || $this->accessingCentralRouteFromTenantDomain) $abortRequest($request, $next)` -- the abort only happens for central or tenant routes, never universal).

So maybe I'd rephrase this a bit

lukinovec: Looking at this now, the current docblock makes sense, probably no reason to change it.

stancl: This also seemed strange to me, so I'm marking this review unresolved. The docblock explicitly says that if the route is universal, the middleware will be skipped. Previously that was also implemented explicitly.

Now with this change it'd be implemented implicitly, making the docblock seem weird (you read it and think "*where* is this logic for skipping if it's `universal`?") and possibly too reliant on implementation details that aren't super explicit.

Unlike this PR's description, the original issue said the `routeIsUniversal()` check is redundant because `shouldBeSkipped()` is checking for the route being universal already (though that doesn't seem correct to me because the relevant code is actually more nuanced)
```php
if (Tenancy::routeIsUniversal($route) && $this instanceof IdentificationMiddleware) {
    return $this->determineUniversalRouteContextFromRequest(request()) === Context::CENTRAL;
}
```
On the other hand this PR's desc talks about the universal case being "default fallthrough" by virtue of not passing the `$this->accessingTenantRouteFromCentralDomain($request, $route) || $this->accessingCentralRouteFromTenantDomain($request, $route)` condition.

Since we use route modes in a few different contexts, it wasn't immediately clear to me if the code used in those methods (`tenancy()->getRouteMode($route)`) even can be universal, or if it just determines the actual runtime context of the request. But I can see the docblock for that method says it has explicit high-priority logic for the `universal` case.

But if I change the code like this:
```diff
if ($this->accessingTenantRouteFromCentralDomain($request, $route) || $this->accessingCentralRouteFromTenantDomain($request, $route)) {
    $abortRequest = static::$abortRequest ?? function () {
        abort(404);
    };

    return $abortRequest($request, $next);
+} else {
+    dump(tenancy()->getRouteMode($route)->name);
}
```

~~I only ever see `CENTRAL` and `TENANT` printed, never `UNIVERSAL`, so I'm not sure if the explanation in this PR's description is accurate.~~

That's when running PreventAccess tests, Universal tests do show UNIVERSAL here. So we just don't have tests for this case in the PreventAccess test file. Could be improved but not important.

Also confirmed that that branch is only ever reached if the patch from this PR is applied.

PR #1434, opened 2026-02-17
Change tenants.id column type from string to uuid

IlyasMohetna:
This PR updates the tenants table migration by changing the primary key column type from string to uuid.

Before:
$table->string('id')->primary();

After:
$table->uuid('id')->primary();

This ensures proper UUID typing at the database level and aligns with Stancl Tenancy's recommended configuration.

https://laravel.com/docs/12.x/migrations#column-method-uuid


stancl:
This is an optimization that can be made optionally in individual applications.

We use string by default because we support other ID generators that produce strings. UUID isn't required and switching ID generators should be simple by default.

PR #1447, opened 2026-03-28
Enhance tenant migration command to support parallel execution

9khaledSayed:
- Added a `--parallel` option to the Migrate command for concurrent tenant migrations.
- Introduced a `parallel-batch-size` option to control the number of concurrent migrations.
- Implemented a new method `runParallel` to handle parallel migrations.
- Updated the `handle` method to initiate parallel migrations when the option is set.
- Added tests to verify the functionality of the parallel migration option.

### DIFF
--- a/src/Commands/Migrate.php
+++ b/src/Commands/Migrate.php
@@ -10,12 +10,18 @@ use Illuminate\Database\Migrations\Migrator;
 use Stancl\Tenancy\Concerns\DealsWithMigrations;
 use Stancl\Tenancy\Concerns\ExtendsLaravelCommand;
 use Stancl\Tenancy\Concerns\HasATenantsOption;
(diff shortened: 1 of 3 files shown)


stancl:
Hi, thanks for the PR.

We already have parallel migrations in v4 (master branch) as well as parallel execution for all commands that deal with multiple tenants.

I'm not sure if we want to maintain two separate implementations for this.

If you'd like to make some improvements to the parallel logic in v4, I'm happy to take a look.

9khaledSayed:
Thanks for the context — I wasn’t aware v4 already covers parallel migrations and parallel multi-tenant commands. I’ll drop this change against 3.x and, if I have improvements, I’ll open a follow-up against master that builds on the existing v4 implementation.

PR #1448, opened 2026-03-31
[4.x] BroadcastingConfigBootstrapper: fix central instances persisting in tenant context

### LINKED ISSUE #1454: Modifications to `BroadcastingConfigBootstrapper::$credentialsMap` are overwritten with values from `$mapPresets`
### Bug description

The v4 docs mention that when using `BroadcastingConfigBootstrapper` you can map your own custom tenant values in `TenancyServiceProvider::boot`, like so:

```
BroadcastingConfigBootstrapper::$credentialsMap[
    'broadcasting.connections.pusher.key'
] = 'pusher_key';
```

However this doesn't work as `BroadcastingConfigBootstrapper` merges `$credentialsMap` and `$mapPresets` together with **`$mapPresets` as the second param to `array_merge`**, causing customisations to keys present in `$mapPresets` to be prioritized.

https://github.com/archtechx/tenancy/blob/e31249dd097b8b3b526808c378fbc5db1bc941fc/src/Bootstrappers/BroadcastingConfigBootstrapper.php#L57

### Steps to reproduce

Set some custom values in `TenancyServiceProvider::boot`:

```
BroadcastingConfigBootstrapper::$credentialsMap = [
    'broadcasting.connections.reverb.key' => 'slug',
    'broadcasting.connections.reverb.app_id' => 'slug',
    'reverb.apps.0.reverb.key' => 'slug',
    'reverb.apps.0.reverb.app_id' => 'slug',
];
```

Add a `dd` at the end of `BroadcastingConfigBootstrapper::__construct`

```
public function __construct(
        protected Repository $config,
        protected Application $app
    ) {
        static::$broadcaster ??= $config->get('broadcasting.default');
        static::$credentialsMap = array_merge(static::$credentialsMap, static::$mapPresets[static::$broadcaster] ?? []);

        dd(static::$credentialsMap);  // <--- HERE
    }
```

Observe that the customisations have been reset back to the presets:

```
array:6 [▼ // vendor/stancl/tenancy/src/Bootstrappers/BroadcastingConfigBootstrapper.php:59
  "broadcasting.connections.reverb.key" => "reverb_key"
  "broadcasting.connections.reverb.app_id" => "reverb_app_id"
  "reverb.apps.0.reverb.key" => "slug"
  "reverb.apps.0.reverb.app_id" => "slug"
  "broadcasting.connections.reverb.secret" => "reverb_secret"
  "broadcasting.connections.reverb.options.cluster" => "reverb_cluster"
]
```

### Expected behavior

Values in `$credentialsMap` do not get reverted to the preset values.

Note: In my tests swapping the params to `array_merge` indeed fixes the issue. For now I am just overriding `$mapPresets` from `boot` instead, but it feels hacky compared to the documented method.

### Laravel version

13

### stancl/tenancy version

4 (dev-master)

lukinovec:
### Broadcasting auth route problem (`Broadcast` facade always uses central `BroadcastManager`)

> Note: Tested with Pusher and Reverb. Also, this is the primary problem that this PR solves.

When using `BroadcastingConfigBootstrapper` for broadcasting on private channels with multiple broadcasting apps (= each tenant has its own Pusher/Reverb/... app and credentials), the auth requests sent to the broadcasting server use the central broadcasting credentials. This is because the `Broadcast` facade (used in `BroadcastController::authenticate` like `Broadcast::auth($request)` to authorize the channels **and then retrieve the auth key that's then sent to the broadcasting server**) keeps using the `BroadcastManager` instance resolved in the central context instead of the tenant `BroadcastManager`. Calling `withBroadcasting()` in `bootstrap/app.php` results in calling `Broadcast::routes()` in the central context, which resolves and stores the central `BroadcastManager` (which is never cleared, and is used in tenant context).

Clearing the facade's resolved instance (`Illuminate\Contracts\Broadcasting\Factory`) in `BroadcastingConfigBootstrapper::bootstrap()` forces the facade to re-resolve `Factory` (= `BroadcastManager`) on the next use, so the next `Broadcast::auth()` call uses the tenant `BroadcastManager` in the tenant context (and clearing the resolved instance in `revert()` makes the `Broadcast` facade use the central `BroadcastManager` again), and the credentials from the current config will be used for authentication.

### Central `BroadcastManager` doesn't pass custom driver creators to the tenant manager

When registering a custom driver creator (`app(BroadcastManager::class)->extend('custom-driver', fn($app, $config) => new CustomBroadcaster(...))`) in the central context (e.g. in `BroadcastServiceProvider`/`AppServiceProvider`, or in our case, in the tests), the creator won't be available in the tenant context. Calling e.g. `app(BroadcastManager::class)->driver('custom-driver')` in the tenant context (if the creator was registered only in the central context) would throw `InvalidArgumentException` ("Driver [custom-driver] is not supported.").

To fix that, register the original (central) `BroadcastManager`'s custom creators in the new `BroadcastManager` in `BroadcastingConfigBootstrapper::bootstrap()`.

> Note: This was always a problem, the tests just never caught it. The original test where we used a custom driver creator actually registered the same creator on each context switch (see the `$registerTestingBroadcaster()` calls in the removed `BroadcastingTest` file) -- if the creator was registered just once (which is what we do now, after the fix), the test would fail. The test registered the driver creator repeatedly which worked around custom creators not persisting after switching between contexts.

### Central `Broadcaster` instance bound in tenant context

> Note: This section describes the problem with resolving/injecting the `Broadcaster` contract directly (via DI or `app(Broadcaster::class)`). The `extend()` call that fixes it also plays a second, bigger role -- see the "How `BroadcastingConfigBootstrapper::bootstrap()` works" section below.

The problem was that resolving `Illuminate\Contracts\Broadcasting\Broadcaster::class` in tenant context returned the broadcaster instance from the central context (Laravel binds the contract as a singleton, resolved from the default driver's config as it was at resolution time -- before tenancy was initialized). Fixed by making `Broadcaster::class` resolve to the current `BroadcastManager`'s default broadcaster (= the tenant broadcaster in tenant context) via `app->extend()`.

Since the manager caches its broadcasters (see "Broadcasters are resolved once per tenancy initialization" below), `app(Broadcaster::class)` and `Broadcast::driver()` return the same instance in every context -- same as Laravel's default behavior.

### How `BroadcastingConfigBootstrapper::bootstrap()` works (the `extend()` calls)

Both `extend()` calls run immediately during `bootstrap()`, not lazily on some later resolution -- `extend()` executes the closure right away when the extended singleton is already resolved, and both singletons are resolved at the top of `bootstrap()` (where we save the central instances for `revert()`). 

Since the singletons are already resolved, the container also doesn't store the closures as extenders -- they run once during `bootstrap()` and that's it. Once `revert()` restores the original instances, resolving either singleton again (even after a `forgetInstance()`) goes purely through the original bindings. So when tenancy initializes:

1. `setConfig()` maps the tenant's properties to the broadcasting config.
2. The `BroadcastManager` extend swaps the bound manager for a fresh one with an empty driver cache and passes the central manager's custom driver creators to it.
3. The `Broadcaster` contract extend calls `connection()` on the tenant manager, which resolves the default broadcaster using the updated (tenant) config and caches it as the manager's default driver. The central broadcaster's auth properties -- the channel auth closures, their options, and the authenticated user callback -- are then copied onto this tenant broadcaster.
4. `Broadcast::clearResolvedInstance()` makes the facade re-resolve on the next call, so it returns the tenant manager instead of the stale central one.

When `/broadcasting/auth` gets hit later, `Broadcast::auth()` goes through the tenant manager's `driver()`, which returns the broadcaster cached in step 3 (built with the tenant credentials, using the copied channel auth closures). This means the `Broadcaster` contract extend isn't just for code that resolves or injects the contract directly -- it's also what makes channel auth work in tenant context. Without it, `app(Broadcaster::class)` would keep returning the stale central broadcaster, and the tenant broadcaster's `$channels` would stay empty, so `Broadcast::auth()` would throw a 403 for every channel registered in the central context. The same goes for user authentication -- `resolveAuthenticatedUser()` has no fallback, so without the copied authenticated user callback (registered via `Broadcast::resolveAuthenticatedUserUsing()`), `/broadcasting/user-auth` would throw a 403 in tenant context.

`revert()` restores the original central manager and broadcaster instances via `instance()` (they're saved at the top of `bootstrap()` and nothing touches them while tenancy is initialized) and clears the facade's resolved instance again.

### Broadcasters are resolved once per tenancy initialization, `TenancyBroadcastManager` is removed

Originally, `TenancyBroadcastManager` re-resolved the broadcasters listed in its `$tenantBroadcasters` static property on every retrieval. Earlier, we considered that good since it meant direct config changes in tenant context were picked up immediately. [As discussed in the review](https://github.com/archtechx/tenancy/pull/1448#discussion_r3555941319), there's probably no real use case for that, and the re-resolving actually caused a subtle bug: channels registered via `Broadcast::channel()` in tenant context got silently lost on the next retrieval (e.g. during a `/broadcasting/auth` request), because each re-resolution started over from a copy of the central channels -- making the auth request fail with a 403 as if the channel was never registered.

Now, `TenancyBroadcastManager` is removed entirely. `BroadcastingConfigBootstrapper` binds a fresh, base `BroadcastManager` with no cached broadcasters, so the broadcasters get resolved using the tenant's credentials on first use and stay cached (like in the parent manager) for the duration of the tenant's context. The central broadcaster's auth properties (the channel auth closures, their options, and the authenticated user callback) are copied to the tenant broadcaster directly in the `Broadcaster` contract's `extend()` closure -- since Laravel only ever uses the *default* broadcaster's channel auth closures for broadcasting auth (both `Broadcast::channel()` and `Broadcast::auth()` go through the default broadcaster), the properties only have to be copied to the default broadcaster.

The channel auth closures are always *only* on the default broadcaster, even in plain Laravel. Every `Broadcast::channel()` call adds an entry to the same `$channels` array on the default broadcaster -- a channel with its own authorization rules is just one of those entries, so it gets copied along with the rest. The only way to register a closure on a non-default broadcaster is calling `Broadcast::driver('foobar')->channel(...)` explicitly, and nothing in Laravel's auth flow ever reads those (`/broadcasting/auth` always authenticates using the default broadcaster).

This all means that:
- `TenancyBroadcastManager` is removed. Its `$tenantBroadcasters` property's purpose was listing the broadcasters to re-resolve (and pass the central channel auth closures to) -- custom drivers now work without the need to configure anything, and the closure copying is a few lines in the bootstrapper instead of a manager override.
- Channels registered via `Broadcast::channel()` in tenant context persist for the duration of that context. They don't leak into other tenants' contexts or into the central context.
- For broadcasters to use updated credentials, tenancy has to be reinitialized. Direct broadcasting config changes made in tenant context aren't picked up by the broadcasters, and tenant property changes are only mapped to config in `bootstrap()`. An already-injected (= stale) `Broadcaster` instance additionally needs to be obtained again after reinitialization -- reinitializing swaps the bound instance, so a previously injected instance keeps using the old credentials.

### NOTE: Already-connected clients stop receiving broadcasts after a credential update

Updating a tenant's credentials doesn't disconnect already-connected clients (tested with Reverb) -- they stay connected using the old key, but they stop receiving broadcasts, since broadcasts sent after the update are sent with the new credentials.

The frontend has to reconnect using the new key (e.g. by refreshing the page). Meaning, clients can't be notified about the credential change through websockets themselves -- a broadcast sent after the update already uses the new credentials, so it won't reach clients connected with the old key. So for things like notifying clients in response to the credential changes, a different mechanism is needed.

### Credential map: overriding presets (BroadcastingConfigBootstrapper)

Previously, credential mappings from `$mapPresets` overrode mappings defined in `$credentialsMap`. If someone used e.g. Pusher and wanted to override some of that preset's mappings, e.g. use 'pusher_app_key' instead of 'pusher_key' by specifying 'pusher_app_key' in `$credentialsMap`, the preset's mapping ('pusher_key') would still be used.

Fixed that by reversing the `array_merge()` order in `BroadcastingConfigBootstrapper::__construct()`.

### **MINOR BC:** `BroadcastingConfigBootstrapper::$broadcaster` property removed

`BroadcastingConfigBootstrapper::$broadcaster` determined which `$mapPresets` preset to apply instead of just using `broadcasting.default`. Its only effect was applying a preset for a connection other than the default -- a connection nothing resolves, so it did nothing useful. Set `$credentialsMap` directly if you need a non-default mapping.

Removing it also let us drop the static mutation from the constructor, which used to overwrite the user-configured `$credentialsMap`. The preset merge now happens locally in `setConfig()`.

### Tests

Deleted the BroadcastingTest file, moved the tests to appropriate bootstrapper test files.

Added tests
- for mapping tenant properties to broadcaster credentials (including keeping the central config values when a tenant doesn't have a mapped property, and reverting to the central credentials after a tenant with credential overrides)
- for the rest of the changes mentioned above (including a regression test for the tenant-context channel registration bug)
- for copying the channel options and the authenticated user callback along with the channel auth closures
- for the bound `Broadcaster` and the manager's default broadcaster being the same instance in central and tenant contexts
- for broadcasters that only implement the `Broadcaster` contract (instead of extending the abstract class), and for configuring which map preset is used via the `$broadcaster` property

Also improved the existing tests.

## Summary

* **New Features**
  * Added tenant-aware broadcasting configuration that rebinds broadcasting/auth state per tenancy and restores the central broadcasting context on tenancy end.
* **Bug Fixes**
  * Improved tenant credential/auth mapping so tenant channel registrations and authorization settings propagate correctly.
  * Fixed channel helper behavior for tenant/global channel registration and prefixing.
* **Refactor**
  * Replaced the prior tenancy broadcasting override with configuration bootstrapping–driven rebinding.
* **Tests**
  * Added end-to-end coverage for broadcasting config and channel helpers; removed the older broadcasting test suite.
* **Chores**
  * Updated the testing broadcaster to accept config; adjusted static analysis ignore rules.


src/Overrides/TenancyBroadcastManager.php
```diff
@@ -7,26 +7,37 @@
...
 use Illuminate\Broadcasting\BroadcastManager;
 use Illuminate\Contracts\Broadcasting\Broadcaster as BroadcasterContract;
-use Illuminate\Contracts\Foundation\Application;
 
+/**
+ * BroadcastManager override that always re-resolves the broadcasters in static::$tenantBroadcasters
+ * when attempting to retrieve them and passes the channels of the original (central) broadcaster
+ * to the newly resolved (tenant) broadcasters.
+ *
+ * Affects calls that use app(BroadcastManager::class)->get().
```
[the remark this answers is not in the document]
lukinovec: I think "calls that use app(BroadcastManager::class)->get()" is enough to make it clear that methods that use the `get()` method are affected

tests/Bootstrappers/BroadcastingConfigBootstrapperTest.php
```diff
@@ -10,18 +10,20 @@
...
 
     BroadcastingConfigBootstrapper::$credentialsMap = [];
-    TenancyBroadcastManager::$tenantBroadcasters = ['pusher', 'ably'];
+    TenancyBroadcastManager::$tenantBroadcasters = ['pusher', 'ably', 'reverb'];
 });
 
 afterEach(function () {
     BroadcastingConfigBootstrapper::$credentialsMap = [];
-    TenancyBroadcastManager::$tenantBroadcasters = ['pusher', 'ably'];
+    TenancyBroadcastManager::$tenantBroadcasters = ['pusher', 'ably', 'reverb'];
```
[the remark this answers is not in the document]
lukinovec: Do we handle static property resetting like you suggested in any test file?

lukinovec:
@coderabbitai review

tests/Bootstrappers/BroadcastingConfigBootstrapperTest.php
```diff
@@ -38,68 +40,174 @@
...
+    tenancy()->initialize($tenant1);
+
+    // When updating tenant properties without reinitializing, the tenant property update doesn't update the config,
+    // so the config has to be modified manually. Only methods that use TenancyBroadcastManager::get()
+    // will use the updated credentials without needing to reinitialize tenancy (e.g. the bound
+    // BroadcasterContract instance will still use the original credentials, even after config gets updated directly).
+    config(["broadcasting.connections.{$driver}.key" => 'new_tenant1_key']);
+
+    expect(app(BroadcastManager::class)->driver()->config['key'])->toBe('new_tenant1_key');
+    expect(Broadcast::driver()->config['key'])->toBe('new_tenant1_key');
```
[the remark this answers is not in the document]
lukinovec: I'm not sure if asserting wrong behavior like this is good. Since this is a caveat, not the desired behavior. Would it still be more correct to add the assertion?

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -64,18 +74,49 @@ public function bootstrap(Tenant $tenant): void
...
         });
+
+        // Swap currently bound Broadcaster instance for one that's resolved through the tenant BroadcastManager.
+        // Note that updating broadcasting config (credentials) in tenant context doesn't update the credentials
+        // used by the bound Broadcaster instance. If you need to e.g. send a notification in response to
+        // updating tenant's broadcasting credentials in tenant context, it's recommended to
+        // reinitialize tenancy after updating the credentials.
+        $this->app->extend(Broadcaster::class, function () {
+            return $this->app->make(BroadcastManager::class)->connection();
+        });
```
stancl: This says that it's swapping the instance in the container, *doesn't say why*, and instead says that this isn't sufficient in some cases. If we're adding a long comment here, I'd instead just explain what this is for.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -64,18 +74,49 @@ public function bootstrap(Tenant $tenant): void
...
+
+        // Clear the resolved Broadcast facade's Illuminate\Contracts\Broadcasting\Factory instance
+        // so that it gets re-resolved as TenancyBroadcastManager instead of the central BroadcastManager
+        // when used. E.g. the Broadcast::auth() call in BroadcastController::authenticate (/broadcasting/auth).
+        Broadcast::clearResolvedInstance(BroadcastingFactory::class);
```
stancl: Is the argument needed here?

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -64,18 +74,49 @@ public function bootstrap(Tenant $tenant): void
...
     }
 
     public function revert(): void
     {
-        // Change the BroadcastManager and Broadcaster singletons back to what they were before initializing tenancy
-        $this->app->singleton(BroadcastManager::class, fn (Application $app) => $this->originalBroadcastManager);
+        // Revert the bound BroadcastManager and Broadcaster singletons back to their original state
+        $this->app->singleton(BroadcastManager::class, fn (Application $app): ?BroadcastManager => $this->originalBroadcastManager);
```
stancl: What's the reason for the added typehint here? We don't use it with the other singleton reset and it wasn't here before

src/Overrides/TenancyBroadcastManager.php
```diff
@@ -7,26 +7,38 @@
...
+     * - should always be recreated using $this->resolve(), even when they're cached and available
+     *   in $this->drivers so that when you update broadcasting config in the tenant context,
+     *   the updated config/credentials will be used for broadcasting immediately.
+     *   Note that in cases like this, only direct config changes are reflected right away.
+     *   For the broadcasters to reflect tenant property changes made in tenant context,
+     *   you still have to reinitialize tenancy after updating the tenant properties intended
+     *   to be mapped to broadcasting config, since the properties are only mapped to config
+     *   on BroadcastingConfigBootstrapper::bootstrap().
+     * - should inherit the original broadcaster's channels (= the channels registered in
+     *   the central context, e.g. in routes/channels.php, before this manager overrides the bound BroadcastManager).
```
stancl: I'm thinking if there is a reason for breaking the cache like this. When would someone update the config? Only when tenancy is being initialized - the config is loaded from the tenant and then stays the same. There aren't many use cases for someone updating broadcasting config long *after* tenancy is initialized.

If they would make any change, it'd be through changes on the tenant itself which aren't reflected in the config anyway, so a reinitialization would be needed.

The only benefit of doing this that I can see is that it makes our tests a little easier maybe? Fewer reinit calls needed.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: Maybe the whole comment could be:
```
// Swap the currently bound Broadcaster singleton (resolved earlier with the central credentials)
// for one resolved through the tenant BroadcastManager, so that anything resolving the Broadcaster
// contract gets a broadcaster that uses the tenant's credentials instead of the stale central one.
// Unlike broadcasters resolved through the manager (re-resolved on each call), this instance is
// resolved once, so config changes made later in tenant context are only picked up on reinitialization.
```

I think changing it to this could make things clear enough (also didn't include the reinitialization example on purpose here since it's not tested and likely invalid)

lukinovec: > didn't include the reinitialization example on purpose here since it's not tested and likely invalid

It really was invalid in the end. Even with reinitialization, you still need to refresh the page to reconnect to Echo using the new key.

So, my final suggestion for the comment here:

"Swap the currently bound Broadcaster singleton (resolved earlier with the central credentials)
for one resolved through the tenant BroadcastManager, so that anything resolving the Broadcaster
contract gets a broadcaster that uses the tenant's credentials instead of the stale central one.
Unlike broadcasters resolved through the manager (re-resolved on each call), this instance is
resolved once, so credential changes made later in tenant context don't affect it until tenancy
is reinitialized."

(corrected the comment a bit and omitted the incorrect example)

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: I think I added the arg for additional clarity (though now I see it doesn't help in that regard at all). Confirmed that it's absolutely not needed:
```php
public static function clearResolvedInstance($name = null)
{
    unset(static::$resolvedInstance[$name ?? static::getFacadeAccessor()]);
}
```

Since `Broadcast:: getFacadeAccessor()` returns `Illuminate\Contracts\Broadcasting\Factory::class`, which is exactly what we're passing to `clearResolvedInstance` (just aliased to `BroadcastingFactory`), we can just use `Broadcast::clearResolvedInstance()`. The behavior will be identical.

So I'll make this just `Broadcast::clearResolvedInstance()`

lukinovec: (same in `bootstrap()`)

lukinovec: Updated in https://github.com/archtechx/tenancy/pull/1448/commits/937c8c87d981854ed3e7ca4a3a382a4f056d4f32 (also improved the comment above `clearResolvedInstance` in `bootstrap()` -- it didn't fully explain why we need to clear the resolved instance).

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: https://github.com/archtechx/tenancy/pull/1448/commits/9af173553e6551d6de012fd7cec25c200fc2113f

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: Definitely a leftover. My IDE shows typehints like these as gray text, and when I click that text accidentally, it just adds the typehint -- so my bad for including the change in a commit here..

**The typehint adds nothing, so I'd just remove it** (also for consistency with the other singleton reset)

Now that we're dealing with this, maybe `fn (Application $app) => ...` could be changed to `fn () => ...`, since that `Application` typehint adds nothing (and `$app` isn't used at all)? Seems like in code similar to this, we only add have `$app` in the closure's params when it's used, so removing this would make things more consistent.

lukinovec: Took care of both the typehint and the unused `$app` here: https://github.com/archtechx/tenancy/pull/1448/commits/24aeb23431f9fac1104ea98e8f034be724aba8dc

src/Overrides/TenancyBroadcastManager.php
lukinovec: Right. I agree that there probably aren't many use cases for updating the config after tenancy is initialized, so I guess we can forget about it entirely.

Even with this re-resolving logic, after direct broadcasting config updates, a page reload is necessary for the frontend to reconnect to Echo using the new credentials. The re-resolving logic here loses its purpose, the channel-passing logic is the thing we care about here.

Now I found out that when the drivers are re-resolved, only _the central channels_ get passed to them. So when someone registers a channel in tenant context (pretty unlikely but possible), and the driver gets re-resolved later e.g. during the `/broadcasting/auth` request (during `Broadcast::auth()`), the driver will only have the channels registered in the central context again (since the driver got re-resolved, its channel list is fresh, until we pass the central channels to it -- and then, it'll only have the central channels in the list). So the `/broadcasting/auth` request would throw a `403` as if the channel never got registered because the driver just doesn't have it in its list at this point due to the re-resolution. This is probably not a huge problem, but it's not right.

If we don't care about direct config updates and just establish that "for broadcasters to use the up-to-date credentials/config, tenancy has to be reinitialized", `TenancyBroadcastManager` could actually get simplified a lot. Passing the channels to freshly resolved broadcasters could be its only purpose (after initializing tenancy, all drivers would get freshly resolved because we create a new `TenancyBroadcastManager` instance with no drivers cached). Instead of overriding `get()` (which we do now just so that the broadcasters are always re-resolved), we could override `resolve()` just so that the freshly resolved broadcasters receive the central channels (and if someone registers a channel in the tenant context, that'll work without a problem because we won't re-resolve the broadcasters).

We could get rid of `TenancyBroadcastManager::$tenantBroadcasters` entirely since its purpose was to list names of broadcasters that should get re-resolved _and_ receive the central channels (the coupling here was weird anyway). Since in this theoretical implementation, we'd just be passing central channels to broadcasters newly resolved (and then cached) by `TenancyBroadcastManager`, I think it'd be pretty safe to just do this for all broadcasters, not just some specific ones.

Anyways, these are pretty big changes. I'll try changing the code around and see if the changes would make sense in the end.

(Also, yeah, the tests would have to get updated)

lukinovec: > I'll try changing the code around and see if the changes would make sense in the end.

Tried that, and it turned out good. Will push the changes after polishing everything.

Worth noting that now that there's no re-resolving going on, `app(BroadcasterContract::class)` and `Broadcast::driver()` will be the same instances in tenant context (which is consistent with Laravel's behavior).

> Now I found out that when the drivers are re-resolved, only the central channels get passed to them. So when someone registers a channel in tenant context (pretty unlikely but possible), and the driver gets re-resolved later e.g. during the /broadcasting/auth request (during Broadcast::auth()), the driver will only have the channels registered in the central context again

Added a regression test for this and made it pass -- a channel registered via `Broadcast::channel()` in tenant context now survives subsequent `driver()` calls (the test fails with the original `get()` override, passes with the new `resolve()` one). Also tested that the channels registered in tenant context only live for the duration of that context, so the channels don't leak from tenant to central context, and they don't leak from one tenant's context to another's.

lukinovec: Pushed the current changes: https://github.com/archtechx/tenancy/pull/1448/commits/2a0464d694dc12b819824394fa8b3f1918121314

src/Overrides/TenancyBroadcastManager.php
```diff
@@ -7,52 +7,46 @@
...
+        // Broadcasters only have to implement the Illuminate\Contracts\Broadcasting\Broadcaster contract
+        // which doesn't require the channels property, so we only pass the channels to
+        // Illuminate\Broadcasting\Broadcasters\Broadcaster instances (= all the default broadcasters, e.g. PusherBroadcaster).
+        if ($originalBroadcaster instanceof Broadcaster && $newBroadcaster instanceof Broadcaster) {
+            $this->passChannelsFromOriginalBroadcaster($originalBroadcaster, $newBroadcaster);
         }
 
-        return parent::get($name);
+        return $newBroadcaster;
     }
```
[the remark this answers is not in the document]
lukinovec: This is a non-issue.

`BroadcastingConfigBootstrapper::bootstrap()`  does `$this->originalBroadcaster = $this->app->make(Broadcaster::class)` before the manager is bound. By the time `resolve()` runs, `BroadcasterContract` is always an already-resolved instance here.

So "app->make(BroadcasterContract::class) depends on the central broadcaster having already been resolved before TenancyBroadcastManager is swapped in" is completely fine.

lukinovec:
@coderabbitai full review

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -64,17 +73,47 @@ public function bootstrap(Tenant $tenant): void
...
+        // for the tenant BroadcastManager's default broadcaster, so that anything resolving the Broadcaster
+        // contract gets the same tenant broadcaster that the manager uses, instead of the stale central one.
+        $this->app->extend(Broadcaster::class, function () {
+            return $this->app->make(BroadcastManager::class)->connection();
+        });
+
+        // Extending the binding doesn't update the Broadcast facade's cached instance,
+        // so clear it to make the facade re-resolve to TenancyBroadcastManager instead of the central
+        // BroadcastManager — e.g. in the Broadcast::auth() call in BroadcastController (/broadcasting/auth).
+        Broadcast::clearResolvedInstance();
```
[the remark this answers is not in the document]
lukinovec: Technically valid, but not a real scenario. `connection()` can only throw here if a tenant explicitly stores `null`/invalid credentials. Tenants without credential attributes fall back to the central config and resolve fine, and central credentials are already proven working (Laravel would fail the same way otherwise). 

Storing null credentials is just bad usage that fails loudly either way, so we're not adding the try/catch here.

tests/Bootstrappers/BroadcastChannelPrefixBootstrapperTest.php
```diff
@@ -137,3 +140,96 @@ protected function formatChannels(array $channels)
...
+
+    expect($tenantChannelClosure($centralUser, $tenant->getTenantKey(), $centralUser->name))->toBeTrue();
+    expect($tenantChannelClosure($centralUser, $tenant->getTenantKey(), $tenantUser->name))->toBeFalse();
+
+    tenancy()->initialize($tenant);
+
+    // The channel closure runs in the tenant context
+    // Only the tenant user is available
+    expect($tenantChannelClosure($centralUser, $tenant->getTenantKey(), $tenantUser->name))->toBeFalse();
+    expect($tenantChannelClosure($tenantUser, $tenant->getTenantKey(), $tenantUser->name))->toBeTrue();
```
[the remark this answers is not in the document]
lukinovec: Makes sense, since `$tenantChannelClosure` and `$centralChannelClosure` are set before `tenant_channel()`, comparing them after calling `tenant_channel()` won't tell us anything about what the `tenant_channel()` call did.

So retrieving the registered closure after `tenant_channel()` and comparing that one is correct here -- pushed the fix.

(Maybe this test could be improved a bit further -- the `$tenantChannelClosure` variable may be a bit more confusing now, but the test works and is clear enough I think. So I'm keeping it as is)

src/Overrides/TenancyBroadcastManager.php
lukinovec: In the end, I deleted the entire `TenancyBroadcastManager` class in https://github.com/archtechx/tenancy/pull/1448/commits/9e6e0a988bbd1988254a78a5257ddc6280ff9737 as its only purpose was to pass the central channel auth closures to each newly resolved broadcaster in the tenant context. Only the default broadcaster's channel auth closures matter.

(see the PR's description)

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -54,27 +61,72 @@ public function __construct(
...
     }
 
     public function revert(): void
     {
-        // Change the BroadcastManager and Broadcaster singletons back to what they were before initializing tenancy
-        $this->app->singleton(BroadcastManager::class, fn (Application $app) => $this->originalBroadcastManager);
-        $this->app->singleton(Broadcaster::class, fn (Application $app) => $this->originalBroadcaster);
+        // Revert the bound BroadcastManager and Broadcaster singletons back to their original state
+        $this->app->singleton(BroadcastManager::class, fn () => $this->originalBroadcastManager);
+        $this->app->singleton(BroadcasterContract::class, fn () => $this->originalBroadcaster);
```
stancl:
```suggestion
        // Revert the bound BroadcastManager and Broadcaster singletons back to their original state
        $this->app->instance(BroadcastManager::class, $this->originalBroadcastManager);
        $this->app->instance(BroadcasterContract::class, $this->originalBroadcaster);
```
Would this be equivalent?

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -54,27 +61,72 @@ public function __construct(
...
+            // manually, otherwise, Broadcast::auth() would throw a 403 for those channels.
+            // Since Laravel only ever uses the default broadcaster's channel auth closures for broadcasting auth,
+            // we only have to pass the channel closures to the default broadcaster.
+            //
+            // The $channels property and the channel() method aren't part of the Broadcaster contract -- they come
+            // from the abstract Broadcaster class, so the closures can only be copied between broadcasters extending it
+            // (which all of Laravel's default broadcasters, e.g. PusherBroadcaster, do).
+            if ($centralBroadcaster instanceof Broadcaster && $tenantBroadcaster instanceof Broadcaster) {
+                // invade() because channels can't be retrieved through any of the broadcaster's public methods
+                $centralBroadcaster = invade($centralBroadcaster);
```
stancl: There's a `getChannels()` method on `Broadcaster` I think

src/Bootstrappers/BroadcastingConfigBootstrapper.php
```diff
@@ -54,27 +61,72 @@ public function __construct(
...
+            $tenantManager = new BroadcastManager($this->app);
+
+            // Pass the custom driver creators registered in the central context to the new manager
+            // so that custom drivers work in tenant context without having to re-register the creators manually.
+            foreach (invade($centralManager)->customCreators as $driver => $creator) {
+                $tenantManager->extend($driver, $creator);
+            }
+
+            return $tenantManager;
+        });
```
stancl, line 82: Is the reason for creating the instance manually like this that if we simply forgot the bound instance from the container, the custom creators that were already registered would be lost?

To confirm - when would `BroadcastManager::extend()` be used? Is this change simply because of how we do things in our tests or would real applications also extend the manager *before tenancy bootstrap*?

And what are some common use cases.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: Weird, I thought there was no method like that, but looking at the `Broadcaster`'s blame, the method was added in Laravel 10.2.0. But nevermind, good catch, I'll use `getChannels()` here.

For the options we still need `invade()` either way. Both `retrieveChannelOptions()` and the `$channelOptions` prop are protected, so `getChannels()` doesn't let us remove `invade()` entirely. We'd still need either `invade($centralBroadcaster)->retrieveChannelOptions($channel)` per channel, or `$channelOptions = invade($centralBroadcaster)->channelOptions` once.

I guess the latter could be a bit more correct because `retrieveChannelOptions()` does some pattern matching and returns the first matching pattern's options, so e.g. if we had `orders.{x}` and `orders.{id}` channels registered, `retrieveChannelOptions('orders.{id}')` would return `orders.{x}`'s options and we'd copy those onto the wrong channel. This couldn't really break anything in practice because when patterns overlap like that, Laravel always uses the first matching one during auth (for both the callback and the options), so the second channel's options never get read anyway. But reading `$channelOptions[$channel]` directly means the copy is just a 1:1 copy, no pattern matching involved.

So the code could look like this:
```php
if ($centralBroadcaster instanceof Broadcaster && $tenantBroadcaster instanceof Broadcaster) {
    // invade() because the channel options can't be retrieved through any of the broadcaster's public methods
    $channelOptions = invade($centralBroadcaster)->channelOptions;

    foreach ($centralBroadcaster->getChannels() as $channel => $callback) {
        $tenantBroadcaster->channel($channel, $callback, $channelOptions[$channel] ?? []);
    }
}
```

Or we can just correct the comment about `invade()` (so that it says `invade()` is needed to access the channel options, not the channels) and keep using `retrieveChannelOptions()`.

I guess I'd just change the code like I suggested in the example above. There's no issue with using `$channelOptions` instead of the `retrieveChannelOptions()` method, and it'll let us use `invade()` for the thing it's actually needed for.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: Tested this and yeah, `instance()` works, no tests fail.

It might even be a bit better since in `bootstrap()`, `extend()` on an already-resolved singleton only swaps the instance (the binding is never touched), so `instance()` puts the container back exactly as it was before `bootstrap()`, while `singleton()` also overwrites the original binding with our closure.

Applying this.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > Is the reason for creating the instance manually like this that if we simply forgot the bound instance from the container, the custom creators that were already registered would be lost?

Exactly, yeah

> To confirm - when would BroadcastManager::extend() be used? Is this change simply because of how we do things in our tests or would real applications also extend the manager before tenancy bootstrap?

I made the change because when I reviewed the tests, I realized that this could be a real problem with any custom driver.

Custom drivers are registered via `Broadcast::extend('foobar', ...)` in a service provider's `boot()`, which runs before tenancy is initialized, so in a real app the creators are always on the central manager before bootstrap. Without copying them, a tenant using a custom driver as the default broadcaster would get "Driver [foobar] is not supported" in the tenant context.

> And what are some common use cases

Writing your own driver, or installing a package that ships one. I see that for example, the Centrifugo package registers its driver exactly like this, in its provider's `boot()` (it calls `extend()` on an injected `BroadcastManager` instead of the facade, but that's the same thing since `Broadcast::extend()` proxies to the manager): https://github.com/denis660/laravel-centrifugo/blob/bec36ea8323fb56dff42bfc3999600abc567afaf/src/CentrifugoServiceProvider.php#L27-L29

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: Updated here https://github.com/archtechx/tenancy/pull/1448/commits/81d41b4690fe731d2b606191563399358f003b0a

lukinovec: I see that a PHPStan error came up:

`Error: Ignored error pattern #Spatie\\Invade\\Invader# (method.notFound) was not matched in reported errors.`

> method.notFound

That's because now, we don't do `invade($thing)->method()` -- we do `invade($thing)->property`, right (= instead of a method call on `invade()`, we access a property through `invade()`)?

stancl: It just says an ignore was not used for anything. The ignore is in `phpstan.neon`, there was one for method.notFound and one for property.notFound. Former is now unused. I'll just comment it out.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
stancl: Is there a reason we are recreating the manager from scratch, and then copying existing stuff into the new manager, instead of trying to reset parts of the manager - if possible. The ones that contain state that becomes utdated in the tenant context.

As a brief test, I looked at what `forget*` methods there are in the manager and what properties exist there. I saw `forgetDrivers()` and that there aren't really many other properties that could hold outdated state. If I replace this whole thing with just `forgetDrivers()` on the bound `BroadcastManager`, all tests except two assertions that are coupled to this specific implementation seem to pass.

Have we tried that before and was it insufficient to make everything work? Wondering why we'd be re-instantiating managers and copying private properties to new instances and things like that in this bootstrapper when we normally just try to reset state.

Thought of this when I looked at the `invade()`-related review since I checked where we use invade and it's only this bootstrapper and some tests.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
stancl: > I guess the latter could be a bit more correct because retrieveChannelOptions() does some pattern matching and returns the first matching pattern's options, so e.g. if we had orders.{x} and orders.{id} channels registered

Can you confirm in confident terms if there would be any real issue here.

> Or we can just correct the comment about invade() (so that it says invade() is needed to access the channel options, not the channels) and keep using retrieveChannelOptions().

And then it'd be just `invade()` just because of the method being protected?

I'm thinking if we could address this instance similarly to my other suggestion with `forgetDrivers()` assuming that suggestion is usable.

Main things I'm looking at:
- facade points to the manager
- manager redirects `__call` to the default driver
- that's how e.g. `broadcast()` is handled by the manager but `channel()` is handled by the default connection/driver
- what we're doing here is creating a fresh instance on the new driver and then copying over old properties

There are some issues with this approach and the alternative approach (resetting the *existing* object).

Current approach:
- only handles channels and channel options? There's a bunch of other state on `Broadcaster` (also included in the facade docblocks) such as `resolveAuthenticatedUserUsing` which we do not copy
- requires `invade()`

Resetting the existing object:
- we could just keep all this various state on the object and only reset the *parts that are central/tenant-specific*, the parts that matter, but that would be driver specific. We'd need driver-specific logic but perhaps it could work better

For instance on Pusher: `getPusher()` and `setPusher()` are public and there's a public `BroadcastManager::pusher()` method.

But there may be a bigger problem with this approach which is that `BroadcastManager->drivers['default driver']` and `app(BroadcasterContract::class)` go out of sync. The singleton instance remains bound but isn't part of the manager anymore. So we'd still need to copy that into the new `BroadcastManager` either with some reflection or some ugly hack that leverages how custom creators or something like that is exactly implemented.

Another separate thing: if we're using reflection anyway in this closure, is there a reason to do things as verbose as:
```php
$channelOptions = invade($centralBroadcaster)->channelOptions;

foreach ($centralBroadcaster->getChannels() as $channel => $callback) {
    $tenantBroadcaster->channel($channel, $callback, $channelOptions[$channel] ?? []);
}
```
Instead of literally just copying the `channels` property, `channelOptions` property and all the other state that should be looked at as I mentioned here?

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > As a brief test ...  If I replace this whole thing with just forgetDrivers() on the bound BroadcastManager, all tests except two assertions that are coupled to this specific implementation seem to pass.

Tested this too -- these tests are coupled to the specific implementation, so yeah, these fail. Other tests pass because we don't deal with the custom creator inheritance, the `BroadcastManager` extend and `forgetDrivers` have similar primary purpose (= make the manager resolve fresh broadcasters).

(actually, I think there's also a gap in the tests -- we don't have a test for whether revert restores the central credentials after a tenant with overrides, and that one would fail with `forgetDrivers()` -- more on that below)

But `forgetDrivers` is problematic.

> Have we tried that before and was it insufficient to make everything work? Wondering why we'd be re-instantiating managers and copying private properties to new instances and things like that in this bootstrapper when we normally just try to reset state.

I don't think we have tried that at all because before, we were swapping the bound broadcast manager for the custom TenancyBroadcastManager (so there was no reason to try that). And after we removed that class, it didn't occur to me to test using just `forgetDrivers()` instead of the `extend()`.

`forgetDrivers()` keeps the custom creators (since it only clears `$drivers`), so custom drivers would still work without the need to copy anything. The only custom creator difference is that a creator registered inside a tenant context would leak into later contexts instead of being isolated per tenant (that's why one of the tests fails), but that's an edge case.

But the worse thing is that we have to properly `revert()` what `bootstrap()` does. If we just did `forgetDrivers()` on the bound manager in `bootstrap()`, things will work just fine in tenant context (if we ignore custom creators). Then on `revert()`, if we did `forgetDrivers()`, we'd discard the tenant broadcaster, but nothing would copy the channel closures back to the central broadcaster. To make the full `forgetDrivers()` implementation actually correct, we'd have to copy the channel closures onto the rebuilt central broadcaster on revert too.

And if instead we kept `instance()` on revert (not `forgetDrivers()`), the manager's cached default driver would still hold the previous (tenant) broadcaster, so /broadcasting/auth in central context would use the old tenant's key (the channel closures were copied over during bootstrap, so /broadcasting/auth wouldn't 403, but the response would be signed using the tenant's key, and the websocket server would just reject it).

So both revert solutions need copying extra state to be correct, which is why extend() (restoring the untouched central instances via instance()) probably ends up cleaner. I'll try how that implementation could look, but like I wrote, I think the `extend()` may be cleaner in the end -- I'll let you know.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > Can you confirm in confident terms if there would be any real issue [with retrieveChannelOptions()] here.

If registering channels like `orders.{x}` and `orders.{y}` together is a common thing, `retrieveChannelOptions` would be a problem as I wrote in the previous reply. If that's not common, there should be no issue with `retrieveChannelOptions` (either way, `channelOptions` won't be a problem)

> And then it'd be just invade() just because of the method being protected?

Yes

> the `get/setPusher()`/reset the existing object idea

Yeah, that could work for Pusher/Ably, but custom drivers would be a problem since we'd need to fall back to recreating these via the manager anyway (I think), and that'd complicate the code. 

Maybe it'd make sense to implement the resetting solution since it has real benefits (preserving broadcaster state + keeping already-injected instances valid), but we still want to support custom drivers, and I'm not sure if that's worth the extra complexity.

So it's either just keep the current recreate + copy logic, or reset the existing object **and**  keep the recreate + copy logic as a fallback for custom drivers.

> But there may be a bigger problem with this approach which is that BroadcastManager->drivers['default driver'] and app(BroadcasterContract::class) go out of sync ... So we'd still need to copy that into the new BroadcastManager either with some reflection or some ugly hack...

Right, the syncing problem would be real. What you sent on Discord should fix it though:
```php
$defaultBroadcaster = ...
Manager::forgetDrivers()
$defaultBroadcaster->setPusher
invade($manager)->drivers = [ $defaultBroadcaster ]
```

`invade($manager)->drivers = [....]` would keep the singleton and the manager cache pointing at one object.

> Another separate thing: if we're using reflection anyway in this closure, is there a reason to do things as verbose as ... Instead of literally just copying the channels property, channelOptions property and all the other state that should be looked at as I mentioned here?

I think the reason to do things the verbose way is to call the `$broadcaster->channel(...)` method so that the channels are copied "properly" since the `channel()` method does some extra things besides setting the channels/channelOptions properties. But that might not be relevant. You're right that we're missing `authenticatedUserCallback` here. There's another property, `bindingRegistrar`, but that's resolved lazily, so I don't think we need to copy that.

I'm testing your suggestions now (related: https://github.com/archtechx/tenancy/pull/1448#discussion_r3585949144), and I'll let you know how the implementation could look.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: (Keeping all the changes local for now since this review is tied to https://github.com/archtechx/tenancy/pull/1448#discussion_r3586720056, and we'll be doing more testing/code changes)

lukinovec: > actually, I think there's also a gap in the tests -- we don't have a test for whether revert restores the central credentials after a tenant with overrides, and that one would fail with forgetDrivers()

Added assertions for this to the credential mapping test. They fail with the forgetDrivers() thing we tested in bootstrap(), pass with extend().

https://github.com/archtechx/tenancy/pull/1448/commits/9efebf964c3687e53fd49abad4122bbc2d1fbdc6

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > Current approach: only handles channels and channel options? There's a bunch of other state on Broadcaster (also included in the facade docblocks) such as resolveAuthenticatedUserUsing which we do not copy

Updated a test so that the auth user callback copying is tested (the test fails):
https://github.com/archtechx/tenancy/pull/1448/commits/50afc49d2b8705256d29352f19bdc39cd4f74de2

> ... literally just copying the channels property, channelOptions property and all the other state that should be looked at

For now, I changed the implementation like this: https://github.com/archtechx/tenancy/pull/1448/commits/fe7468ab7222f3676a0f716e285f30188f1c66f6

It makes the test added in previous linked commit pass (and all the other tests pass too, so this approach works).

Looking into the "resetting the existing object" implementation more now.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > So both revert solutions need copying extra state to be correct, which is why extend() (restoring the untouched central instances via instance()) probably ends up cleaner. I'll try how that implementation could look

Tried it, and it's like I wrote -- `bootstrap()` gets simpler (`forgetDrivers()` + the same auth state copying), but `revert()` has to re-resolve the central broadcaster and copy the auth state back onto it:
```php
public function revert(): void
{
    $this->unsetConfig();

    // Clear the tenant broadcasters so they get re-resolved using the central config again
    $this->app->make(BroadcastManager::class)->forgetDrivers();

    // A freshly resolved broadcaster has no auth state, so copy it back from the original broadcaster
    $centralBroadcaster = $this->app->make(BroadcastManager::class)->connection();
    $this->copyAuthState($this->originalBroadcaster, $centralBroadcaster);
    $this->app->instance(BroadcasterContract::class, $centralBroadcaster);

    Broadcast::clearResolvedInstance();
}
```

So the copying happens in both directions instead of once (meaning `invade()` stays too), the original broadcaster isn't restored on revert -- a fresh instance replaces it, so instances resolved before initializing tenancy end up stale. This all would be a bit more messy than the current implementation, so imo, switching to this won't be worth it.

(About the new assertions I mentioned in the previous comment -- with the full implementation, these assertions pass too, since `revert()` re-resolves the broadcaster with the central config -- only the instance check assertions fail)

Also tested the "resetting the existing object" idea from the other review, that's more interesting. I'll comment on that in a minute.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > Looking into the "resetting the existing object" implementation more now

So I tried implementing this.

<details>
<summary>The full diff</summary>

```diff

diff --git a/src/Bootstrappers/BroadcastingConfigBootstrapper.php b/src/Bootstrappers/BroadcastingConfigBootstrapper.php
index 2afd61a..594605f 100644
--- a/src/Bootstrappers/BroadcastingConfigBootstrapper.php
+++ b/src/Bootstrappers/BroadcastingConfigBootstrapper.php
@@ -4,18 +4,20 @@ declare(strict_types=1);
 
 namespace Stancl\Tenancy\Bootstrappers;
 
+use Illuminate\Broadcasting\Broadcasters\AblyBroadcaster;
 use Illuminate\Broadcasting\Broadcasters\Broadcaster;
+use Illuminate\Broadcasting\Broadcasters\PusherBroadcaster;
 use Illuminate\Broadcasting\BroadcastManager;
 use Illuminate\Config\Repository;
 use Illuminate\Contracts\Broadcasting\Broadcaster as BroadcasterContract;
 use Illuminate\Foundation\Application;
-use Illuminate\Support\Facades\Broadcast;
 use Stancl\Tenancy\Contracts\TenancyBootstrapper;
 use Stancl\Tenancy\Contracts\Tenant;
 
 /**
- * Maps tenant credentials to the broadcasting config and rebinds BroadcastManager
- * and Broadcaster so that broadcasters get resolved using the tenant credentials.
+ * Maps tenant credentials to the broadcasting config and swaps the default broadcaster's
+ * underlying client (e.g. the Pusher instance) for one built with the tenant credentials,
+ * keeping the broadcaster and manager instances intact.
  */
 class BroadcastingConfigBootstrapper implements TenancyBootstrapper
 {
@@ -33,8 +35,19 @@ class BroadcastingConfigBootstrapper implements TenancyBootstrapper
 
     public static string|null $broadcaster = null;
 
+    /**
+     * Callbacks for making a broadcaster use the passed config, keyed by connection name.
+     * Usually that means swapping the broadcaster's client for one created using the passed
+     * config (that's how Pusher and Ably, supported out of the box, work), but it's up to the
+     * callback -- e.g. a driver storing its config in a property can just overwrite it:
+     *
+     * ['foobar' => fn (Broadcaster $broadcaster, BroadcastManager $manager, array $config) => invade($broadcaster)->config = $config]
+     *
+     * Custom drivers without a resetter fall back to the recreate + copy logic (see resetBroadcasting()).
+     */
+    public static array $broadcasterResetters = [];
+
     protected array $originalConfig = [];
-    protected BroadcastManager|null $originalBroadcastManager = null;
     protected BroadcasterContract|null $originalBroadcaster = null;
 
     public static array $mapPresets = [
@@ -66,44 +79,84 @@ class BroadcastingConfigBootstrapper implements TenancyBootstrapper
 
     public function bootstrap(Tenant $tenant): void
     {
-        $this->originalBroadcastManager = $this->app->make(BroadcastManager::class);
         $this->originalBroadcaster = $this->app->make(BroadcasterContract::class);
 
         $this->setConfig($tenant);
 
-        // Make BroadcastManager resolve to a fresh manager with no cached broadcasters,
-        // so that its broadcasters get resolved using the updated (tenant) broadcasting
-        // config and stay cached for the duration of the tenant's context.
-        $this->app->extend(BroadcastManager::class, function (BroadcastManager $centralManager) {
-            $tenantManager = new BroadcastManager($this->app);
+        $this->resetBroadcasting();
+    }
 
-            // Pass the custom driver creators registered in the central context to the new manager
-            // so that custom drivers work in tenant context without having to re-register the creators manually.
-            foreach (invade($centralManager)->customCreators as $driver => $creator) {
-                $tenantManager->extend($driver, $creator);
+    /**
+     * Make the broadcasters use the current broadcasting config.
+     *
+     * The default broadcaster's client gets swapped for a client created using the current config,
+     * while the broadcaster instance itself stays the same (its auth state stays untouched, and
+     * all the already-resolved broadcaster/manager instances will be up-to-date with the changes).
+     */
+    protected function resetBroadcasting(): void
+    {
+        $manager = $this->app->make(BroadcastManager::class);
+        $default = $this->config->get('broadcasting.default');
+
+        // Clear the manager's cached broadcasters so that the non-default broadcasters get
+        // re-resolved using the current config (custom driver creators stay registered).
+        $manager->forgetDrivers();
+
+        $broadcaster = $this->originalBroadcaster;
+
+        if ($broadcaster instanceof Broadcaster && $this->resetBroadcaster($broadcaster, $manager, $default)) {
+            // The broadcaster now uses the current config. Cache the broadcaster instance in the
+            // manager (since forgetDrivers() cleared it), so that the manager and the bound
+            // Broadcaster singleton keep using the same instance.
+            invade($manager)->drivers = [$default => $broadcaster];
+
+            $this->app->instance(BroadcasterContract::class, $broadcaster);
+        } else {
+            // If the broadcaster can't be reset (a custom driver without a registered resetter),
+            // fall back to resolving a fresh broadcaster using the current config
+            // and copying the auth state to it.
+            $freshBroadcaster = $manager->connection();
+
+            if ($broadcaster) {
+                $this->copyAuthState($broadcaster, $freshBroadcaster);
             }
 
-            return $tenantManager;
-        });
+            $this->app->instance(BroadcasterContract::class, $freshBroadcaster);
+        }
+    }
+
+    /**
+     * Make the broadcaster use the given connection's current config.
+     *
+     * Different drivers store the credentials in different places, so this is driver-specific:
+     * Pusher and Ably store them in the broadcaster's client, so their client gets swapped for one
+     * created using the current config. Custom drivers get reset using the resetter registered for the connection
+     * (see $broadcasterResetters). Returns false when there's no known way to reset the broadcaster
+     * (no registered resetter and not a Pusher/Ably broadcaster).
+     */
+    protected function resetBroadcaster(Broadcaster $broadcaster, BroadcastManager $manager, string $connection): bool
+    {
+        $config = $this->config->get("broadcasting.connections.{$connection}") ?? [];
 
-        // Swap the currently bound Broadcaster singleton (resolved earlier with the central credentials)
-        // for the tenant BroadcastManager's default broadcaster, so that anything resolving the Broadcaster
-        // contract gets the same tenant broadcaster that the manager uses, instead of the stale central one.
-        // The closure runs immediately (the extended singleton is already resolved), and it's also what makes
-        // channel auth work in tenant context -- the broadcaster resolved here gets cached as the tenant
-        // manager's default driver and receives the central broadcaster's auth state (see copyAuthState()).
-        $this->app->extend(BroadcasterContract::class, function (BroadcasterContract $centralBroadcaster) {
-            $tenantBroadcaster = $this->app->make(BroadcastManager::class)->connection();
+        if ($resetter = static::$broadcasterResetters[$connection] ?? null) {
+            $resetter($broadcaster, $manager, $config);
 
-            $this->copyAuthState($centralBroadcaster, $tenantBroadcaster);
+            return true;
+        }
+
+        if ($broadcaster instanceof PusherBroadcaster) {
+            $broadcaster->setPusher($manager->pusher($config));
 
-            return $tenantBroadcaster;
-        });
+            return true;
+        }
 
-        // Extending the binding doesn't update the Broadcast facade's cached instance,
-        // so clear it to make the facade re-resolve to the tenant BroadcastManager instead of the central
-        // one — e.g. in the Broadcast::auth() call in BroadcastController (/broadcasting/auth).
-        Broadcast::clearResolvedInstance();
+        if ($broadcaster instanceof AblyBroadcaster) {
+            $broadcaster->setAbly($manager->ably($config));
+
+            return true;
+        }
+
+        return false;
     }
 
     /**
@@ -134,14 +187,9 @@ class BroadcastingConfigBootstrapper implements TenancyBootstrapper
 
     public function revert(): void
     {
-        // Revert the bound BroadcastManager and Broadcaster singletons back to their original state
-        $this->app->instance(BroadcastManager::class, $this->originalBroadcastManager);
-        $this->app->instance(BroadcasterContract::class, $this->originalBroadcaster);
-
-        // Clear the resolved Broadcast facade instance so that it gets re-resolved as the central BroadcastManager
-        Broadcast::clearResolvedInstance();
-
         $this->unsetConfig();
+
+        $this->resetBroadcasting();
     }
 
     protected function setConfig(Tenant $tenant): void
diff --git a/tests/Bootstrappers/BroadcastingConfigBootstrapperTest.php b/tests/Bootstrappers/BroadcastingConfigBootstrapperTest.php
index 110d028..6796986 100644
--- a/tests/Bootstrappers/BroadcastingConfigBootstrapperTest.php
+++ b/tests/Bootstrappers/BroadcastingConfigBootstrapperTest.php
@@ -15,6 +15,7 @@ use Illuminate\Contracts\Broadcasting\Broadcaster as BroadcasterContract;
 $cleanup = function () {
     BroadcastingConfigBootstrapper::$broadcaster = null;
     BroadcastingConfigBootstrapper::$credentialsMap = [];
+    BroadcastingConfigBootstrapper::$broadcasterResetters = [];
     BroadcastingConfigBootstrapper::$mapPresets = [
         'pusher' => [
             'broadcasting.connections.pusher.key' => 'pusher_key',
@@ -40,6 +41,10 @@ beforeEach(function () use ($cleanup) {
     Event::listen(TenancyEnded::class, RevertToCentralContext::class);
 
     $cleanup();
+
+    BroadcastingConfigBootstrapper::$broadcasterResetters['testing'] = function ($broadcaster, $manager, array $config) {
+        $broadcaster->config = $config;
+    };
 });
 
 afterEach($cleanup);
@@ -85,6 +90,26 @@ test('ending tenancy reverts the bound broadcaster to the original instance', fu
     expect($originalBroadcaster)->toBe(app(BroadcasterContract::class));
 });
 
+test('the bound broadcaster and the manager default driver are the same instance in all contexts', function () {
+    config([
+        'tenancy.bootstrappers' => [BroadcastingConfigBootstrapper::class],
+        'broadcasting.default' => 'testing',
+        'broadcasting.connections.testing.driver' => 'testing',
+    ]);
+
+    app(BroadcastManager::class)->extend('testing', fn ($app, $config) => new TestingBroadcaster('testing', $config));
+
+    expect(app(BroadcasterContract::class))->toBe(app(BroadcastManager::class)->driver());
+
+    tenancy()->initialize(Tenant::create());
+
+    expect(app(BroadcasterContract::class))->toBe(app(BroadcastManager::class)->driver());
+
+    tenancy()->end();
+
+    expect(app(BroadcasterContract::class))->toBe(app(BroadcastManager::class)->driver());
+});
+
 test('BroadcastingConfigBootstrapper maps tenant properties to broadcaster credentials correctly', function (string $driver) {
     config([
         'broadcasting.default' => $driver,

 
```
</details>

`bootstrap()` and `revert()` both end up doing the same thing: update the config -> `forgetDrivers()` -> swap the default broadcaster's client on the existing broadcaster (`$broadcaster->setPusher($manager->pusher($config))`) -> cache that broadcaster in the manager again.

>  invade($manager)->drivers = [....] would keep the singleton and the manager cache pointing at one object.

Confirmed -- added a minimal test for this (assert `app(BroadcasterContract::class) === app(BroadcastManager::class)->driver()` in all contexts) and it passes. It's a separate test only because the existing `ending tenancy reverts the bound broadcaster to the original instance` test (which already checks this, _but only in tenant context_) fails on its first assertion with this implementation, so assertions added there wouldn't even run. If we keep the current implementation, I'd just add the missing assertions to that existing test instead of adding the new one.

>  ... either with some reflection or some ugly hack that leverages how custom creators or something like that is exactly implemented.

Tried if this could work without `invade()`, using `$manager->extend($default, fn () => $broadcaster)` (`forgetDrivers()` doesn't delete the creators, so the manager keeps returning the same instance) -- the tests pass the same as with `invade()`, but it's wrong: `$drivers` are keyed by _connection_ name, while `$customCreators` are keyed by _driver_ name (`$config['driver']`). So for a connection named differently than its driver (e.g. connection `foo` with the `pusher` driver), the creator wouldn't get used at all, and if it got registered by the driver name instead, it would hijack all connections that use that driver (+ override a creator the user registered themselves). So `invade()` is correct here.

> So it's either just keep the current recreate + copy logic, or reset the existing object and keep the recreate + copy logic as a fallback for custom drivers.

This turned out to be right -- custom drivers have no `setPusher()` equivalent, so I added a static `$broadcasterResetters` property where the user can register a closure that resets their driver's broadcaster (makes it use the current config), and connections without a resetter fall back to the recreate + copy logic.

(About the diff: the tests use `TestingBroadcaster` everywhere, and the reset path only handles `PusherBroadcaster`/`AblyBroadcaster` **+ connections with a registered resetter**. So in the tests, I registered a resetter for the 'testing' connection to cover the reset path, and the tests using other connections -- e.g. the credential mapping test -- have no resetter, so they cover the fallback.)

Note that we use the custom 'testing' driver in the config bootstrapper tests, so I set the resetter for it using `BroadcastingConfigBootstrapper::$broadcasterResetters['testing'] = ...` in `beforeEach`. So tests that use the 'testing' driver exercise the reset path.

There's also a problem we didn't bring up before (wrote about this on Discord): preserving all the state on the single broadcaster also preserves the state we don't want to keep. Auth state (channels, channel options and user auth callbacks) registered in tenant context stays on the broadcaster after tenancy ends, and custom creators registered in tenant context persist between contexts too -- the two isolation tests fail because of this. It's fixable by saving the original state in `bootstrap()` and restoring it in `revert()`, but that's the auth state copying logic again. That said, registering these things in tenant context is pretty unlikely in real apps -- channels, creators and the user auth callback are normally registered in service providers (= in central context) -- so it _is_ wrong, but probably wouldn't be a problem in the vast majority of apps.

Overall, 11/15 tests pass -- the two isolation tests fail (as mentioned above), and then the two instance check tests which assert that the manager/broadcaster are fresh instances. That's the opposite of what this approach does, so those two don't really tell us anything.

So the advantage of resetting the existing object is that the already-resolved/injected instances stay valid across context switches. The disadvantages:
- driver-specific reset logic
- the resetter static property solution for custom drivers
- the "recreate + copy" fallback
- saving/restoring the broadcaster's channels and the manager's custom creators, so that the ones registered in tenant context don't stay registered after tenancy ends (not in the diff -- this would have to be added to make the failing isolation tests pass)

IMO the current implementation is the better trade-off. One approach instead of two, and the context isolation (custom creator and auth state) isn't a problem since each context has its own instances.

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > Also tested the "resetting the existing object" idea from the other review, that's more interesting. I'll comment on that

See https://github.com/archtechx/tenancy/pull/1448#discussion_r3597119317

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: (see the `reset-existing-broadcaster` branch: the resetting solution is implemented there)

In the end, we'll go with the current implementation instead of the "resetting the existing object" one discussed and tested here.

For the record, a few corrections/clarifications after the Discord discussion:
- The custom creator isolation disadvantage I mentioned here is irrelevant since custom creators aren't really context-specific state (no real reason to register a creator in tenant context in a way that would cause problems in the central context)
- Why `revert()` runs the reset again instead of restoring a backup: in this variant, the original broadcaster gets mutated while tenancy is initialized (its client is swapped), so reverting means swapping the client back -- there's no untouched instance to restore (which is different from the _current_ implementation, where nothing touches the original broadcaster while tenancy is initialized, so `revert()` can just restore it).
- The `if`/`else` in `resetBroadcasting()`: the else covers two cases -- an abstract Broadcaster without a registered resetter (fresh broadcaster + property copy), and a non-abstract broadcaster (fresh broadcaster, the the `instanceof` check there makes the `copyAuthProperties` method return early, there's nothing to copy anyway since the auth properties are only defined on the abstract `Broadcaster`).

src/Bootstrappers/BroadcastingConfigBootstrapper.php
lukinovec: > https://github.com/archtechx/tenancy/pull/1448#discussion_r3596292638

One correction from the Discord discussion: `revert()` doesn't have to rebuild the central broadcaster and copy the properties back. Since nothing touches the central broadcaster while tenancy is initialized, it can be backed up in `bootstrap()` and `revert()` can just be:

```php
$manager->forgetDrivers();
$this->app->instance(BroadcasterContract::class, $this->originalBroadcaster);
invade($manager)->drivers = [$default => $this->originalBroadcaster];
```

With that, only the custom creator isolation and manager instance check assertions fail (both coupled to the current implementation, and the creator isolation is a non-issue anyway since creators aren't context-specific state). More on the `invade($manager)->drivers` part in https://github.com/archtechx/tenancy/pull/1448#discussion_r3597119317. We're keeping the current implementation either way.

lukinovec:
@coderabbitai full review

lukinovec:
@coderabbitai full review

PR #1453, opened 2026-04-13
[4.x] Prevent mkdir() race conditions in FilesystemTenancyBootstrapper

### LINKED ISSUE #1452: Race condition in FilesystemTenancyBootstrapper::storagePath() causes mkdir(): File exists warning
### Bug description

When multiple queue workers (e.g. Laravel Horizon) initialize the same tenant concurrently, the `storagePath()` method in `FilesystemTenancyBootstrapper` triggers an `ErrorException: mkdir(): File exists` warning due to a TOCTOU (time-of-check-time-of-use) race condition.

## Affected Code

`src/Bootstrappers/FilesystemTenancyBootstrapper.php` lines 77–80:

```php
if (! is_dir($path)) {
    // Create tenant framework/cache directory if it does not exist
    mkdir($path, 0750, true);
}
```

Two or more workers can evaluate `is_dir($path)` as `false` simultaneously, then the first `mkdir()` succeeds and subsequent calls fail with "File exists" because the directory was created between the check and the call.

## Environment

- **PHP:** 8.4
- **Laravel:** 12
- **Tenancy:** dev-master (fb654e7)
- **Queue:** Laravel Horizon with multiple workers
- **OS:** Linux (production)

## Full Stack Trace

```
ErrorException: mkdir(): File exists

FilesystemTenancyBootstrapper::storagePath()
  → FilesystemTenancyBootstrapper.php:79

FilesystemTenancyBootstrapper::bootstrap()
  → FilesystemTenancyBootstrapper.php:30

BootstrapTenancy::handle()
  → BootstrapTenancy.php:22

Dispatcher::invokeListeners()
  → Dispatcher.php:319

Tenancy::initialize()
  → Tenancy.php:101

Tenancy::runForMultiple()
  → Tenancy.php:280
```

### Option A — Suppress the warning (preferred)

The `@` operator is the standard PHP pattern for this exact scenario. The race is benign — if the directory already exists, the goal is achieved. This is the same approach used by Symfony's `Filesystem::mkdir()` and Laravel's own `Filesystem::ensureDirectoryExists()`.

```php
if (! is_dir($path)) {
    @mkdir($path, 0750, true);
}
```

### Option B — Attempt unconditionally and verify the outcome

```php
if (! is_dir($path)) {
    @mkdir($path, 0750, true);

    if (! is_dir($path)) {
        throw new \RuntimeException("Failed to create directory: {$path}");
    }
}
```

This adds a post-condition check so that real failures (e.g. permission denied) are still surfaced.

## Note

The same race condition pattern may apply to the `mkdir()` call in `scopeSessions()` if file-based session storage is used.

### Steps to reproduce

1. Configure Horizon with multiple worker processes.
2. Dispatch jobs that initialize the same tenant concurrently (e.g. via `tenancy()->runForMultiple()`).
3. Target a tenant whose `storage/{suffix}/framework/cache` directory does not yet exist.
4. Both workers race through `storagePath()` and the second one throws.

### Expected behavior

No exception 

### Laravel version

12

### stancl/tenancy version

v4

stancl:
Prevent mkdir() race conditions in FilesystemTenancyBootstrapper

This prevents race conditions that may occur if there are two concurrent
processes trying to create the storage path for the tenant. The
storagePath() method runs during bootstrap() which can easily happen
in two places at once. The race condition specifically occurs in between
the is_dir() check and the mkdir() call, the latter producing an
exception if the dir already exist. We simply ignore any error coming
out of mkdir() and then check for success separately.

We could omit that success check since failure is unlikely and would
only occur due to a server misconfiguration that would manifest itself
in other ways as well, but this way the simple TOC/TOU race condition
is prevented while other errors are still reported.

We apply the same change to the mkdir() in scopeSessions() as the logic
is similar.

Resolves #1452

## Summary by CodeRabbit

* **Bug Fixes**
  * Suppressed spurious filesystem warnings during concurrent startup and added clearer failure reporting when required cache and session storage directories cannot be created, reducing noisy logs and improving reliability during application boot.

### DIFF
--- a/src/Bootstrappers/FilesystemTenancyBootstrapper.php
+++ b/src/Bootstrappers/FilesystemTenancyBootstrapper.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Stancl\Tenancy\Bootstrappers;
 
+use Exception;
 use Illuminate\Foundation\Application;
 use Illuminate\Session\FileSessionHandler;
 use Illuminate\Support\Facades\Storage;
@@ -75,8 +76,13 @@ class FilesystemTenancyBootstrapper implements TenancyBootstrapper
             : $this->originalStoragePath . '/framework/cache';
 
         if (! is_dir($path)) {
-            // Create tenant framework/cache directory if it does not exist
-            mkdir($path, 0750, true);
+            // Create tenant framework/cache directory if it does not exist.
+            // We ignore errors due to TOCTOU race conditions, instead we check for success below.
+            @mkdir($path, 0750, true);
+
+            if (! is_dir($path)) {
+                throw new Exception("Unable to create tenant storage directory [{$path}].");
+            }
         }
 
         if ($suffix === false) {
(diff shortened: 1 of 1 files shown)


stancl:
@coderabbitai full review

PR #1456, opened 2026-04-22
Fix tenant sync not triggering when pivot is created in queued jobs

leandrogehlen:
When a tenant is attached to a synced resource (e.g., User) inside a queued job (`Events\TenantCreated::class`), the current filtering logic in updateResourceInCentralDatabaseAndGetTenants() may exclude that tenant from the update cycle. As a result, updateResourceInTenantDatabases() is not triggered for newly attached tenants.

This occurs because the existing filter unconditionally removes the current tenant mapping, even when the pivot record has just been created during the same execution.


stancl:
The way the issue is described I'm not sure if there's a real issue in Tenancy or a usage issue. Important to note though that we will not be merging any major functionality changes to v3 resource syncing - there are some known issues that cannot be fixed without breaking changes (and they have been fixed in v4). Don't think this was one of them/ever reported though.

PR #1458, opened 2026-04-28
[4.x] Create pending tenants with pending_since, improve --with-pending

lukinovec:
> Minor breaking change: Pending tenants would previously go through the creation pipeline as *not* pending and would only be marked as pending after full creation. Now, pending tenants go through the creation process with pending_since set from the start.

Pending tenants aren't getting their `pending_since` set until they're created completely (e.g. their DB was created, migrated and seeded -- first, the tenant is created fully, and only after that, the tenant is updated to have `pending_since`). This is a problem if someone wants to e.g. add a job to the `DatabaseCreated` job pipeline that would check `$this->tenant->pending()`. Since at the point of `DatabaseCreated`, the tenant's `pending_since` isn't set yet, `$this->tenant->pending()` returns `false`, even for tenants created using `createPending()`. So instead of letting the pending tenant get fully created, and only after that, setting its `pending_since` (using `update()`), we now set `pending_since` in `create()`. `CreatingPendingTenant` is now dispatched from the `static::creating` hook, and `PendingTenantCreated` is dispatched from `static::created` for consistency.

Setting `pending_since` right in `create()` made the `MigrateDatabase` and `SeedDatabase` jobs exclude the pending tenants during their creation if the `tenancy.pending.include_in_queries` config was set to `false` -- in that case, these jobs would never migrate or seed the databases of pending tenants. So these jobs now pass `--with-pending` to their underlying commands, with the value set in their `$includePending` static property (`true` by default). This overrides the `tenancy.pending.include_in_queries` config -- unless the `$includePending` properties are set to `false`, these jobs will always include pending tenants.

The `--with-pending` tenant command option originally worked just to opt-in for including pending tenants in the command. Now, `--with-pending` can accept values (`true`/`1` or `false`/`0`), so e.g.
- `tenants:run foo` with `--with-pending`/`--with-pending=true`/`--with-pending=1` includes pending tenants
- `tenants:run foo` with `--with-pending=false`/`--with-pending=0` **excludes** pending tenants (also `--with-pending=foobar` -- invalid input, considered `false`)

Passing `--with-pending` makes the command bypass the `tenancy.pending.include_in_queries` config (so e.g. if `tenancy.pending.include_in_queries` is set to `true`, and `--with-pending=false` is passed to a command, the command will exclude pending tenants). When `--with-pending` is not passed, the command will include or exclude pending tenants based on the `tenancy.pending.include_in_queries` config.

In tenancy-docs, I opened a PR with a note about the `--with-pending` and job changes (https://github.com/archtechx/tenancy-docs/pull/5).

## Summary by CodeRabbit

* **New Features**
  * The pending option now accepts explicit boolean-like values; when omitted, configuration governs behavior.
  * Migration and seeding jobs gain configurable flags to include/exclude pending tenants.
* **Bug Fixes**
  * Pending tenants record their "pending" timestamp at creation and emit lifecycle events consistently.
* **Tests**
  * Expanded coverage for pending-option parsing, job-driven migration/seeding behavior, and lifecycle/job integration tests.

### DIFF
--- a/src/Database/Concerns/HasPending.php
+++ b/src/Database/Concerns/HasPending.php
@@ -49,18 +49,15 @@ trait HasPending
      */
     public static function createPending(array $attributes = []): Model&Tenant
     {
-        $tenant = null;
+        $tenant = static::make(array_merge(
+            ['pending_since' => now()->timestamp],
+            static::getPendingAttributes($attributes),
+            $attributes
+        ));
 
-        try {
-            $tenant = static::create(array_merge(static::getPendingAttributes($attributes), $attributes));
-            event(new CreatingPendingTenant($tenant));
-        } finally {
-            // Update the pending_since value only after the tenant is created so it's
-            // not marked as pending until after migrations, seeders, etc are run.
-            $tenant?->update([
-                'pending_since' => now()->timestamp,
-            ]);
-        }
+        event(new CreatingPendingTenant($tenant));
+
+        $tenant->save();
 
         event(new PendingTenantCreated($tenant));
 
--- a/src/Database/Concerns/HasPending.php
+++ b/src/Database/Concerns/HasPending.php
@@ -28,6 +28,18 @@ trait HasPending
     public static function bootHasPending(): void
     {
         static::addGlobalScope(new PendingScope());
+
+        static::creating(function (self $tenant): void {
+            if ($tenant->pending()) {
+                event(new CreatingPendingTenant($tenant));
+            }
+        });
+
+        static::created(function (self $tenant): void {
+            if ($tenant->pending()) {
+                event(new PendingTenantCreated($tenant));
+            }
+        });
     }
 
     /** Initialize the trait. */
@@ -49,19 +61,11 @@ trait HasPending
      */
     public static function createPending(array $attributes = []): Model&Tenant
     {
-        $tenant = static::make(array_merge(
+        return static::create(array_merge(
             ['pending_since' => now()->timestamp],
             static::getPendingAttributes($attributes),
             $attributes
         ));
-
-        event(new CreatingPendingTenant($tenant));
-
-        $tenant->save();
-
-        event(new PendingTenantCreated($tenant));
-
-        return $tenant;
     }
 
     /**
--- a/src/Database/Concerns/HasPending.php
+++ b/src/Database/Concerns/HasPending.php
@@ -62,9 +62,9 @@ trait HasPending
     public static function createPending(array $attributes = []): Model&Tenant
     {
         return static::create(array_merge(
-            ['pending_since' => now()->timestamp],
             static::getPendingAttributes($attributes),
-            $attributes
+            $attributes,
+            ['pending_since' => now()->timestamp],
         ));
     }
 
--- a/src/Jobs/MigrateDatabase.php
+++ b/src/Jobs/MigrateDatabase.php
@@ -17,6 +17,12 @@ class MigrateDatabase implements ShouldQueue
 {
     use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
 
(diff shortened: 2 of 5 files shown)


src/Database/Concerns/HasPending.php
```diff
@@ -49,18 +49,15 @@ public function pending(): bool
...
-        }
+        $tenant = static::make(array_merge(
+            ['pending_since' => now()->timestamp],
+            static::getPendingAttributes($attributes),
+            $attributes
+        ));
+
+        event(new CreatingPendingTenant($tenant));
+
+        $tenant->save();
```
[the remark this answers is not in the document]
lukinovec: Does this commit https://github.com/archtechx/tenancy/pull/1458/changes/e3673f5557bfb2efd4a7b60cea2eb8673e817f14 resolve this?

lukinovec:
@coderabbitai full review

src/Jobs/MigrateDatabase.php
```diff
@@ -25,6 +31,7 @@ public function handle(): void
     {
         Artisan::call('tenants:migrate', [
             '--tenants' => [$this->tenant->getTenantKey()],
+            '--with-pending' => static::$includePending,
```
stancl: The docblock for `includePending` says:
```php
/**
 * Should pending tenants be included while migrating,
 * regardless of the tenancy.pending.include_in_queries config value.
 */
```

But there are three possible states - force include in pending, force exclude from pending, and default to config. This docblock sounds like it's just about whether a forced inclusion should be done, but if the value of the static property is set to false, `'--with-pending' => false` will be passed resulting in forced exclusion when the user would likely expect just defaulting to the config.

stancl: To change this to default to the config we could just add `?: config('tenancy.pending.include_in_queries')` but I guess there might be use cases for specifically excluding.

So maybe making this nullable, adding a `??` default, and expanding the docblock to mention that behavior would be ideal. Would also more closely match the behavior expected in this PR's tests.

stancl: https://github.com/archtechx/tenancy/pull/1458/commits/b7bd439a9d4f72b6c587cdace59c8f2922688bd2

tests/PendingTenantsTest.php
```diff
@@ -236,3 +267,105 @@
...
+    config([
+        'tenancy.bootstrappers' => [DatabaseTenancyBootstrapper::class],
+    ]);
+
+    Event::listen(TenancyInitialized::class, BootstrapTenancy::class);
+    Event::listen(TenancyEnded::class, RevertToCentralContext::class);
+    Event::listen(TenantCreated::class, JobPipeline::make([
+        CreateDatabase::class,
+        PendingTenantJob::class,
+    ])->send(function (TenantCreated $event) {
```
stancl, line 352: The PR description talks about:
> This is a problem if someone wants to e.g. add a job to the DatabaseCreated job pipeline

But the added test uses a `TenantCreated` pipeline. Those two might behave differently so it might be worth seeing if this regression test acts as intended.

stancl: Trying this test on `master` - I get `false` with both `TenantCreated` and `DatabaseCreated`.

On this branch it passes with both this code as well as
```php
Event::listen(TenantCreated::class, JobPipeline::make([
    CreateDatabase::class,
])->send(function (TenantCreated $event) {
    return $event->tenant;
})->toListener());

Event::listen(DatabaseCreated::class, JobPipeline::make([
    PendingTenantJob::class,
])->send(function (DatabaseCreated $event) {
    return $event->tenant;
})->toListener());
```

So we can probably merge it like this since it's simpler. Just keeping this review here to note the difference in tests and PR desc.

PR #1459, opened 2026-04-29
[4.x] Parameter validation and other DB manager improvements

lukinovec:
### Parameter validation

In `statement()` calls of `TenantDatabaseManager`s, use parameter binding when possible. When that's not possible, validate the parameters using `validateParameter()` or `validatePassword()`.

Passwords use a less strict allowlist than other parameters (e.g. DB names), since passwords tend to use more special characters but we can afford to be more restrictive in the generic `validateParameter()`.

In `SQLiteDatabaseManager`, names of file-based databases are validated (in `createDatabase`, `deleteDatabase` and `databaseExists`) using a similar allowlist to the (non-password) parameters in other DB managers, with an additional character: `.` (this addition is necessary since file-based SQLite databases end with `.sqlite`). 

### DatabaseTenancyBootstrapper - harden() and the lost test file

While checking for more places that could use validation, I realized that it's possible to update tenant's db_name to the central DB or the DB of another tenant. Added the `DatabaseTenancyBootstrapper::$harden` property -- setting it to true prevents tenants from connecting to the wrong databases (`RuntimeException` is thrown after connecting to the wrong database).

Also, the DatabaseTenancyBootstrapper test file was ignored while running tests because it lacked the `Test` suffix. Added the suffix and fixed the broken `DATABASE_URL` test in the file.

### SQLiteDatabaseManager - respect static $path property in makeConnectionConfig()

SQLiteDatabaseManager had a bug in `makeConnectionConfig`: the method didn't respect the static `$path` property, it used `database_path()` instead. Added a regression test for that. Also recognizing in-memory SQLite databases (using `isInMemory()`) is more strict now so that simply having a db_name with `_tenancy_inmemory_`  somewhere in the name doesn't make a file-based database considered in-memory.

### MySQLDatabaseManager - charset and collation defaulting

Creating databases with `null` charsets and collations resulted in a `QueryException`, since null isn't a valid charset/collation. To solve that, in the `CREATE DATABASE` statement in MySQLDatabaseManager, only add charset/collation to the statement if they are not null.

MySQL defaults to the server's charset and collation, so it's safe to not pass any charset/collation in the `CREATE DATABASE` statement and let MySQL choose. Also, if e.g. collation is non-null and charset is null, MySQL will use a charset compatible with the used collation, and this works both ways.

## Release Notes

* **Security**
  * Strengthened SQL injection protection by parameterizing queries across database managers instead of interpolating identifiers.
  * Added stricter allowlist-based validation for database-related inputs (database names, usernames, passwords), failing fast on forbidden characters.

* **New Features**
  * Introduced optional hardening mode to verify tenant database isolation during initialization and block unsafe reuse.

* **Improvements**
  * Improved identifier handling for PostgreSQL schema/database operations and tightened SQLite in-memory/path handling validation.

* **Tests**
  * Expanded validation and hardening coverage across supported database managers.


lukinovec:
@coderabbitai full review

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -84,6 +92,8 @@ public function createDatabase(TenantWithDatabase $tenant): bool
             return true;
         }
 
+        $this->validateParameter($name);
+
```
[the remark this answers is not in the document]
lukinovec: I don't think we need validation in `SQLiteDatabaseManager::databaseExists` since it just checks if a file exists. That's not a vulnerable to attacks, unlike SQL statements or `file_put_contents()`. Correct me if I'm wrong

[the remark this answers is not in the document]
lukinovec: Resolved in https://github.com/archtechx/tenancy/pull/1459/commits/322257f4569ad5741772496028f85630a39ac7ff

src/Database/TenantDatabaseManagers/MicrosoftSQLDatabaseManager.php
```diff
@@ -10,18 +10,20 @@ class MicrosoftSQLDatabaseManager extends TenantDatabaseManager
...
         return $this->connection()->statement("CREATE DATABASE [{$database}]");
     }
 
     public function deleteDatabase(TenantWithDatabase $tenant): bool
     {
-        return $this->connection()->statement("DROP DATABASE [{$tenant->database()->getName()}]");
+        $database = $this->validateParameter($tenant->database()->getName());
+
+        return $this->connection()->statement("DROP DATABASE [{$database}]");
     }
```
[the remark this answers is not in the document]
lukinovec: Does https://github.com/archtechx/tenancy/pull/1459/commits/75b74f2e6c7d942c83220bf20f2b5f2c989af38d resolve this?

src/Database/TenantDatabaseManagers/PermissionControlledPostgreSQLSchemaManager.php
```diff
@@ -23,23 +23,25 @@ protected function grantPermissions(DatabaseConfig $databaseConfig): bool
...
             /** @var string $primaryKey */
-            $primaryKey = $this->connection()->selectOne(<<<SQL
+            $primaryKey = $this->connection()->selectOne(<<<'SQL'
                 SELECT column_name
                 FROM information_schema.key_column_usage
-                WHERE table_name = '{$tableName}'
+                WHERE table_name = ?
                 AND constraint_name LIKE '%_pkey'
-            SQL)->column_name;
+            SQL, [$tableName])->column_name;
```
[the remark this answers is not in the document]
lukinovec: This code branch only runs when during `createUser`, there are any existing tables, which is only the case when creating users via the `tenants:rls` command. So this code is only reached when using a setup with Postgres RLS -- **a single DB/schema setup**. So I don't think this issue is relevant

lukinovec:
@coderabbitai full review

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -57,6 +60,11 @@ class SQLiteDatabaseManager implements TenantDatabaseManager
      */
     public static Closure|null $closeInMemoryConnectionUsing = null;
 
+    protected static function parameterAllowlist(): string
+    {
+        return 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-.';
+    }
```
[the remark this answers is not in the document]
lukinovec: What's the problematic DB name input here, since e.g. `../some-file` is forbidden because of `/`?

src/Database/TenantDatabaseManagers/PermissionControlledPostgreSQLSchemaManager.php
```diff
@@ -23,23 +23,25 @@ protected function grantPermissions(DatabaseConfig $databaseConfig): bool
...
```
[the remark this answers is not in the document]
lukinovec: Noted, but won't address it in this PR (we're dealing with parameter validation here)

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @throws InvalidArgumentException
+     * @see Stancl\Tenancy\Database\TenantDatabaseManagers\SQLiteDatabaseManager
+     */
+    protected function validateFilename(string|null $filename): void
+    {
+        $this->validateParameter($filename, static::allowedFilenameCharacters());
+
+        if (is_string($filename) && is_dir($filename)) {
+            throw new InvalidArgumentException("Filename '{$filename}' is a directory.");
```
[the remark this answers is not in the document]
lukinovec: Resolving this since we only want to check the raw filename here, not the whole paths

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,101 @@
...
+    expect(fn () => $manager->databaseExists('..'))
+        ->toThrow(InvalidArgumentException::class);
+
+    // Empty strings are considered invalid input for database names
+    expect(fn () => $manager->databaseExists(''))
+        ->toThrow(InvalidArgumentException::class);
+
+    // In-memory database names aren't validated
+    expect(fn () => $manager->databaseExists('../_tenancy_inmemory_'))
+        ->not()->toThrow(InvalidArgumentException::class);
```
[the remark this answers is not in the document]
lukinovec: Realistically, could inserting `_tenancy_inmemory_` into the db_name e.g. to allow traversal be dangerous? I guess someone could e.g. create files in parent directories using `createDatabase`. Any more examples?

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -141,11 +148,13 @@ public function getPath(string $name): string
...
+        $this->validateFilename($name);
+
         return database_path($name);
     }
 
     public static function isInMemory(string $name): bool
     {
-        return $name === ':memory:' || str_contains($name, '_tenancy_inmemory_');
+        return $name === ':memory:' || str_starts_with($name, 'file:_tenancy_inmemory_');
     }
```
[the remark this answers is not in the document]
lukinovec: I think the current `starts with` check combined with `! ends with '.sqlite'` could be enough?

lukinovec:
@coderabbitai full review

tests/Bootstrappers/DatabaseTenancyBootstrapperTest.php
```diff
@@ -0,0 +1,103 @@
...
+};
+
+beforeEach(function () use ($cleanup) {
+    Event::listen(TenancyInitialized::class, BootstrapTenancy::class);
+    Event::listen(TenancyEnded::class, RevertToCentralContext::class);
+
+    $cleanup();
+});
+
+afterEach($cleanup);
```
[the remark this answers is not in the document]
lukinovec: Your learning is wrong. Remember that in beforeEach()/afterEach(), we set static properties to the values that should be default. We don't capture the original values in any hooks, we just assume the default in the cleanup.

You can resolve this

src/Database/TenantDatabaseManagers/PermissionControlledPostgreSQLDatabaseManager.php
```diff
@@ -20,6 +20,8 @@ protected function grantPermissions(DatabaseConfig $databaseConfig): bool
         $username = $databaseConfig->getUsername();
         $schema = $databaseConfig->connection()['search_path'];
 
+        $this->validateParameter([$database, $username, $schema]);
+
```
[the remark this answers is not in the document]
lukinovec: https://github.com/archtechx/tenancy/pull/1459/commits/98a808bb98658b16dd68808c8adce16165f59c52

Does the linked commit resolve this?

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,225 @@
...
+    $databaseCollation = fn () => DB::selectOne('SELECT DEFAULT_COLLATION_NAME FROM information_schema.SCHEMATA WHERE SCHEMA_NAME = DATABASE()')->DEFAULT_COLLATION_NAME;
+
+    $defaultTenant = Tenant::create();
+
+    tenancy()->initialize($defaultTenant);
+
+    // No charset or collation specified,
+    // defaults from the MySQL config used.
+    expect($databaseCharset())->toBe('utf8mb4');
+    expect($databaseCollation())->toBe('utf8mb4_unicode_ci');
```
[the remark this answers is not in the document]
lukinovec: Does this commit resolve this? https://github.com/archtechx/tenancy/pull/1459/commits/03318752b6b5eecad68948697db5d7a70c903998

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,227 @@
...
+
+    $manager = app($databaseManager);
+
+    if ($manager instanceof StatefulTenantDatabaseManager) {
+        $manager->setConnection($driver);
+    }
+
+    $invalidDatabaseName = "\"database_with_quotes\"";
+
+    if (! $manager instanceof ManagesDatabaseUsers) {
```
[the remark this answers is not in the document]
lukinovec: Fixed in https://github.com/archtechx/tenancy/pull/1459/commits/bbd8f6fd986893e51052fd17941f988f902f4730

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,227 @@
...
+        ->not()->toThrow(InvalidArgumentException::class);
+
+    // Directory names are considered invalid input for database names
+    expect(fn () => $manager->databaseExists('..'))
+        ->toThrow(InvalidArgumentException::class);
+
+    // Empty strings are considered invalid input for database names
+    expect(fn () => $manager->databaseExists(''))
+        ->toThrow(InvalidArgumentException::class);
+});
```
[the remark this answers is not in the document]
lukinovec: Aren't these two already tested in `database managers validate parameters that cannot be bound`? This test should basically just cover that the SQLite manager uses different rules/allowlist than the rest of the managers, the databaseExists calls are a simple way of checking if validation works.

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,227 @@
...
+            config(['database.connections.mysql.charset' => []]);
+            DB::purge('mysql');
+
+            $tenantWithNonStringCharset = Tenant::make([
+                'tenancy_db_name' => 'valid_db_name',
+            ]);
+
+            expect(fn () => $manager->createDatabase($tenantWithNonStringCharset))
+                ->toThrow(InvalidArgumentException::class, 'Parameter has to be a string.');
+        }
```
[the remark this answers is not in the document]
lukinovec: Resolved in https://github.com/archtechx/tenancy/pull/1459/commits/587f347b6495c1251f93ebd6f0a522358c614575

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,231 @@
...
+            ->toThrow(InvalidArgumentException::class, 'Forbidden character');
+
+        $tenantWithInvalidPassword = Tenant::make([
+            'tenancy_db_name' => 'valid_database_name890',
+            'tenancy_db_username' => 'valid_USERNAME',
+            'tenancy_db_password' => "p'ssword",
+        ]);
+
+        expect(fn () => $manager->createUser($tenantWithInvalidPassword->database()))
+            ->toThrow(InvalidArgumentException::class, 'Forbidden character');
```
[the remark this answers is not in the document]
lukinovec: Addressed in https://github.com/archtechx/tenancy/pull/1459/commits/099a666dbcb6aaab582833d9254ccd0b2c454091

lukinovec:
@coderabbitai review

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,241 @@
...
+
+        // Special characters are allowed in password
+        $tenantWithValidPassword = Tenant::make([
+            'tenancy_db_name' => 'valid_database_name890',
+            'tenancy_db_username' => 'valid_USERNAME',
+            'tenancy_db_password' => "]pa$$ ;word",
+        ]);
+
+        expect(fn () => $manager->createUser($tenantWithValidPassword->database()))
+            ->not()->toThrow(InvalidArgumentException::class, 'Forbidden character');
```
lukinovec, line 617: Resolved in https://github.com/archtechx/tenancy/pull/1459/commits/649c8027f44964c5294300613d89e2389f90ac9a

lukinovec:
@coderabbitai review

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,245 @@
...
+        $tenantWithNullDbParameters = Tenant::make([
+            'tenancy_db_name' => null,
+            'tenancy_db_username' => null,
+            'tenancy_db_password' => null,
+        ]);
+
+        // validateParameter() doesn't throw InvalidArgumentException if a parameter is null
+        // (an exception will be thrown, but not by validateParameter()).
+        expect(fn () => $manager->createUser($tenantWithNullDbParameters->database()))
+            ->not()->toThrow(InvalidArgumentException::class);
```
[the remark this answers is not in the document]
lukinovec: We're testing validation here. The assertion passes because InvalidArgumentException doesn't get thrown, that's all the assertion cares about (even though e.g. QueryException would get thrown, the assertion passes, and it's clear that we only care about validateParameter() results here). Resolving this

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,93 @@
...
+use InvalidArgumentException;
+
+/**
+ * Provides methods to validate database parameters (e.g. database names, usernames, passwords)
+ * before using them in SQL statements (or in file paths in the case of SQLiteDatabaseManager).
+ *
+ * Used where parameters can be provided by users, and where parameter binding cannot be used.
+ *
+ * @mixin \Stancl\Tenancy\Database\TenantDatabaseManagers\TenantDatabaseManager
+ * @mixin \Stancl\Tenancy\Database\TenantDatabaseManagers\SQLiteDatabaseManager
```
stancl: Why the mixin annotations? We don't appear to be using methods from either of these classes in this trait?

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -5,14 +5,24 @@
...
 use Stancl\Tenancy\Database\Exceptions\TenantDatabaseDoesNotExistException;
+use Throwable;
 
 class DatabaseTenancyBootstrapper implements TenancyBootstrapper
 {
+    /**
+     * When true, throw an exception if a tenant gets connected to
+     * another tenant's database or to the central database.
+     */
+    public static bool $harden = false;
```
stancl, line 42: There should certainly be a second paragraph explaining what tradeoff exists here to explain why this is turned off by default or why one might want to turn this on.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +51,38 @@ public function bootstrap(Tenant $tenant): void
...
         $this->database->reconnectToCentral();
     }
+
+    protected function harden(Tenant $tenant): void
+    {
+        $dbName = DB::getDatabaseName();
+
+        // Check if any other tenant uses this tenant's database
+        if ($tenant::where($tenant->getTenantKeyName(), '!=', $tenant->getTenantKey())
+            ->where('data->tenancy_db_name', $dbName)
```
stancl: We could probably make this a bit more solid by using `$tenant::getDataColumn()` and `$tenant::internalPrefix()`.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +51,38 @@ public function bootstrap(Tenant $tenant): void
         }
 
         $this->database->connectToTenant($tenant);
+
+        if (static::$harden) {
+            try {
+                $this->harden($tenant);
```
stancl: `$harden` is an ok name for the property but I'd name the method a bit different to make it clear what it more clear what it does just from its name.

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,93 @@
...
+     */
+    protected function allowedParameterCharacters(): string
+    {
+        return 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-';
+    }
+
+    /**
+     * Characters allowed in database user passwords.
+     *
+     * Passwords are always quoted in the SQL statements, so it's safe
```
stancl: Are other parameters not always quoted? Does this refer to database names or which other parameters can we not quote?

Besides the difference in quoting we might also want to mention just that we *can* be stricter with other params as they don't use that many characters so a stricter allowlist there is better.

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $allowedCharacters ??= $this->allowedParameterCharacters();
+
+        foreach ((array) $parameters as $parameter) {
+            if (is_null($parameter)) {
+                // Skip if there's nothing to validate
+                // (e.g. when $tenant->database()->getUsername() of an
+                // improperly created tenant is null and it gets passed).
+                continue;
+            }
```
stancl: Can you describe that case in a bit more detail?

I think `null` may be something worth thinking about more because you can end up with a bit weird queries. They'd be fine in most cases, I assume just `""` values but worth considering when exactly this comes up and what are the exact implications.

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,93 @@
...
+
+            if (! is_string($parameter)) {
+                // E.g. if a parameter is retrieved from the config, it isn't necessarily a string
```
stancl: Any examples where this comes up? I wonder if there are cases where a parameter could be numeric and then we might want to cast it to a string. But I'm not sure about where this would come up.

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,93 @@
...
+                if (! str_contains($allowedCharacters, $character)) {
+                    throw new InvalidArgumentException("Forbidden character '{$character}' in parameter.");
+                }
+            }
+        }
+    }
+
+    /**
+     * Ensure password only contains allowed characters (allowedPasswordCharacters())
+     * before used in SQL statements.
```
stancl:
```suggestion
     * before being used in SQL statements.
```
or

```suggestion
     * before they're used in SQL statements.
```

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * Null parameters are skipped.
+     *
+     * @throws InvalidArgumentException
+     */
+    protected function validateParameter(string|array|null $parameters, string|null $allowedCharacters = null): void
+    {
+        $allowedCharacters ??= $this->allowedParameterCharacters();
+
+        foreach ((array) $parameters as $parameter) {
```
stancl: Not sure if `(array)` always works well for strings or perhaps the null case I commented on in another review. `(array) null` returns `[]` I think so you'd never get the other case I reviewed in the first place.

It's standard to use `Arr::wrap()` in Laravel code.

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -122,21 +136,26 @@ public function databaseExists(string $name): bool
...
+
             $baseConfig['database'] = $databaseName;
 
             if (static::$persistInMemoryConnectionUsing !== null) {
                 $dsn = "sqlite:$databaseName";
                 (static::$persistInMemoryConnectionUsing)(new PDO($dsn), $dsn);
             }
         } else {
-            $baseConfig['database'] = database_path($databaseName);
+            $baseConfig['database'] = $this->getPath($databaseName);
```
stancl, line 146: Seems like the old code using `database_path()` instead of `getPath()` was either overlooked or intentional. Could this change have any impact that'd change behavior?

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -122,21 +136,26 @@ public function databaseExists(string $name): bool
     public function makeConnectionConfig(array $baseConfig, string $databaseName): array
     {
         if ($this->isInMemory($databaseName)) {
+            // Named in-memory DBs are formatted like 'file:_tenancy_inmemory_tenant123?mode=memory&cache=shared'
+            $this->validateDatabaseName($databaseName, ':?=&');
```
stancl: I'd probably use a named param for the second arg here since it's the only place where the second param allowlist is *extra* characters.

tests/Bootstrappers/DatabaseTenancyBootstrapperTest.php
```diff
@@ -0,0 +1,103 @@
...
+
+test('harden prevents tenants from using the central database', function () {
+    config([
+        'tenancy.bootstrappers' => [DatabaseTenancyBootstrapper::class],
+    ]);
+
+    DatabaseTenancyBootstrapper::$harden = true;
```
stancl: I think I'd use a dataset for this setting so the test more clearly asserts it's actually `$harden` doing its job here. Same thing in the following test.

tests/Bootstrappers/DatabaseTenancyBootstrapperTest.php
```diff
@@ -0,0 +1,103 @@
...
+    // Connection should be reverted back to central
+    expect(DB::connection()->getName())->toBe('central');
+});
+
+test('database tenancy bootstrapper throws an exception if DATABASE_URL is set', function (string|null $databaseUrl) {
+    config(['database.connections.central.url' => $databaseUrl]);
+
+    if ($databaseUrl) {
+        pest()->expectException(Exception::class);
+    }
```
stancl: Can we make this expectation more targeted to a specific line here? As in expect x to throw.

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,241 @@
     expect($manager->connection()->getConfig('url'))->toBeNull();
 });
 
+test('database managers validate parameters that cannot be bound', function ($driver, $databaseManager) {
```
stancl: Test name is a bit weird

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,241 @@
...
+    ]);
+
+    $tenant = Tenant::make([
+        'tenancy_db_name' => 'tenant.sqlite',
+    ]);
+
+    // SQLiteDatabaseManager::$path is null, the database path is built using database_path()
+    expect($tenant->database()->connection()['database'])->toBe(database_path('tenant.sqlite'));
+
+    SQLiteDatabaseManager::$path = '/custom/path/';
```
stancl, line 697: This static prop is reset in `beforeEach` but not `afterEach` I think

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,241 @@
...
+    $invalidDatabaseName = "\"database_with_quotes\"";
+
+    if (! ($manager instanceof ManagesDatabaseUsers)) {
+        // Only test createDatabase() and deleteDatabase() with non-permission controlled managers here
+        // since permission controlled managers override these methods to e.g. delete users before
+        // calling parent::deleteDatabase(), and with invalid DB name, the user deletion will already
+        // fail before we even get to actual deleteDatabase() logic.
+        $tenant = Tenant::make([
+            'tenancy_db_name' => $invalidDatabaseName,
+            'tenancy_db_username' => 'valid-username',
```
stancl: Why set the username if this is guarded by `! ($manager instanceof ManagesDatabaseUsers)`?

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,241 @@
...
+                'tenancy_db_name' => 'valid_db_name',
+            ]);
+
+            expect(fn () => $manager->createDatabase($tenantWithNonStringCharset))
+                ->toThrow(InvalidArgumentException::class, 'Parameter has to be a string.');
+
+            // Restore the default charset
+            config(['database.connections.mysql.charset' => 'utf8mb4']);
+            DB::purge('mysql');
+        }
```
stancl: Wouldn't it make more sense to move mysql charset assertions to a separate test that just uses MySQL?

tests/TenantDatabaseManagerTest.php
```diff
@@ -539,6 +540,241 @@
...
+        $tenantWithNullDbParameters = Tenant::make([
+            'tenancy_db_name' => null,
+            'tenancy_db_username' => null,
+            'tenancy_db_password' => null,
+        ]);
+
+        // validateParameter() doesn't throw InvalidArgumentException if a parameter is null
+        // (an exception will be thrown, but not by validateParameter()).
+        expect(fn () => $manager->createUser($tenantWithNullDbParameters->database()))
+            ->not()->toThrow(InvalidArgumentException::class);
```
stancl: Just adding a note to revisit this along with the other review about null values

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +51,38 @@ public function bootstrap(Tenant $tenant): void
...
+
+        // Check if any other tenant uses this tenant's database
+        if ($tenant::where($tenant->getTenantKeyName(), '!=', $tenant->getTenantKey())
+            ->where('data->tenancy_db_name', $dbName)
+            ->exists()) {
+            throw new RuntimeException('Tenant cannot use a database of another tenant.');
+        }
+
+        // Check if the current database doesn't have the tenants table (i.e. it's not the central database)
+        if (Schema::hasTable($tenant->getTable())) {
```
stancl, line 102: Wondering if this is the right way to assert this or if we could check that we're on the central connection in another way. Maybe if the config changes by connecting to a tenant that'd be a bit harder? Worth revisiting though

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +51,38 @@ public function bootstrap(Tenant $tenant): void
...
+    protected function harden(Tenant $tenant): void
+    {
+        $dbName = DB::getDatabaseName();
+
+        // Check if any other tenant uses this tenant's database
+        if ($tenant::where($tenant->getTenantKeyName(), '!=', $tenant->getTenantKey())
+            ->where('data->tenancy_db_name', $dbName)
+            ->exists()) {
+            throw new RuntimeException('Tenant cannot use a database of another tenant.');
+        }
```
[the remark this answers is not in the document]
stancl: This is similar to what I suggested in my review about using `internalPrefix()` and `getCustomColumns()` but I'm not sure if this addresses the SQLite concern (which sounds reasonable?)

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
```diff
@@ -146,6 +165,28 @@ public function getPath(string $name): string
...
+    /**
+     * Ensure database name only contains allowed characters
+     * (allowedDatabaseNameCharacters() + $extraAllowedCharacters) and is not a directory name.
+     *
+     * @throws InvalidArgumentException
+     */
+    protected function validateDatabaseName(string $name, string $extraAllowedCharacters = ''): void
+    {
+        $this->validateParameter($name, $this->allowedDatabaseNameCharacters() . $extraAllowedCharacters);
+
```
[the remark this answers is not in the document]
stancl: Don't think this is true, we use level 8 and this isn't an issue in our setup.

src/Database/Concerns/ValidatesDatabaseParameters.php
lukinovec: True, meant to use `@see` there (as in  "see the classes that use this trait")

Fixed here https://github.com/archtechx/tenancy/pull/1459/changes/6e82a9ee55d755b1c9ef36bc1d58da8b61f8d3fc

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: Maybe add something like:

"For example, if `tenancy_db_name` can be set during tenant creation by the users, the user creates
 a tenant that could connect to the database of another tenant. Setting `$harden` to `true` prevents this, but we keep it `false` by default since it's not common to let users specify the tenant database names."
 
 Updated the annotation here: https://github.com/archtechx/tenancy/pull/1459/commits/42a2c8efd95e78092586d5f0fd926153550bf3f6

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: Yeah `$tenant::getDataColumn() . '->' . $tenant->internalPrefix() . 'db_name'` would be more solid.

I'd just need to add proper `@var` annotation to `$tenant`. I guess the right way to do this would be to annotate `$tenant` as `Stancl\Tenancy\Database\Models\Tenant`

lukinovec: Updated here: https://github.com/archtechx/tenancy/pull/1459/commits/b4244be6589b035b314f8ac8be1d5cfcb9c3e5ad

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: Agree. Maybe something like `verifyTenantCanUseDatabase`?

lukinovec: Renamed the method here https://github.com/archtechx/tenancy/pull/1459/changes/b7045c52d8f93cdf627ae660bf87381e03cfb080

src/Database/Concerns/ValidatesDatabaseParameters.php
lukinovec: Now that I'm looking at this, I think this docblock is a bit weird since everything is either quoted or passed using bindings.

> we might also want to mention just that we can be stricter with other params as they don't use that many characters so a stricter allowlist there is better

Yeah, I think that this should be pointed out as the primary difference here.

lukinovec: Updated the annotations here https://github.com/archtechx/tenancy/pull/1459/commits/4386a3b1a384f1444e834a1e6fe3cfe1a9d00001

src/Database/Concerns/ValidatesDatabaseParameters.php
lukinovec: > I assume just "" values

You're right about that -- with MySQL, it'd create an "anonymous" user (empty string name), which sounds invalid, but the user creation would pass. Postgres would instead reject the null/empty string. So maybe actually worth considering throwing an exception here instead of just skipping?

src/Database/Concerns/ValidatesDatabaseParameters.php
lukinovec: E.g. charset set to an array like `['utf8']` -- basically invalid user config

> I wonder if there are cases where a parameter could be numeric and then we might want to cast it to a string

Sounds valid. Though every parameter we pass to `validateParameter()` is either `string|null` (e.g `$databaseConfig->getUsername()`, has return type `string|null`), or mixed (config value like charset/collation, which could be anything)

Adding to this: https://github.com/archtechx/tenancy/pull/1459#discussion_r3371877397 -- if e.g. username is null, and there's any attempt to use it, it sounds like a good idea to throw an exception becuase it's just invalid. But there might be some cases where `null` could mean "default", which could be completely valid? I guess this is worth looking into

src/Database/Concerns/ValidatesDatabaseParameters.php
lukinovec: I think `(array)` and `Arr:wrap` would be equivalent in this case, but yeah not sure why I did not just use `Arr::wrap()` here as usual. I'll just use that

lukinovec: Updated https://github.com/archtechx/tenancy/pull/1459/commits/f9636b15cff3620ee01991ef111051f2fa093cd7

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
lukinovec: Yeah, this was an overlooked bug. See the test added here: https://github.com/archtechx/tenancy/pull/1459/changes/26c161a9407e1b71e3bea5ca4d63ed008591d5eb

Changing `$this->getPath` back to `database_path` makes it fail.

src/Database/TenantDatabaseManagers/SQLiteDatabaseManager.php
lukinovec: Makes sense, that'd be clearer

lukinovec: Updated https://github.com/archtechx/tenancy/pull/1459/commits/b3111f1dde34bc4b0ab255be775c48f693622527

tests/Bootstrappers/DatabaseTenancyBootstrapperTest.php
lukinovec: I think this should do it: https://github.com/archtechx/tenancy/pull/1459/commits/407197b1907981760f75b8fb32ac3e24b4bf154c

tests/Bootstrappers/DatabaseTenancyBootstrapperTest.php
lukinovec: Made the assertions more targeted here: https://github.com/archtechx/tenancy/pull/1459/commits/49356a5513a9f66b8c43e29b76ca33736a5a73ac

tests/TenantDatabaseManagerTest.php
lukinovec: Agree, thought about it for quite a while but couldn't come up with anything better. Maybe something more generic, like "database managers validate parameters before using them in queries"? But that may be too broad/inaccurate..

tests/TenantDatabaseManagerTest.php
lukinovec: Right, added afterEach with the reset.

tests/TenantDatabaseManagerTest.php
lukinovec: Probably a leftover. Removing the username changes nothing, so it's redundant to set it. Fixed (https://github.com/archtechx/tenancy/pull/1459/commits/d28abdec892c8a1d6c2cd53d2212167b009e5730).

tests/TenantDatabaseManagerTest.php
lukinovec: Yeah that would make the test cleaner (https://github.com/archtechx/tenancy/pull/1459/commits/36782ebee76c88dd55c0a016c74e873802e9c778)

src/Bootstrappers/DatabaseTenancyBootstrapper.php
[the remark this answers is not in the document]
lukinovec: I think the SQLite concern is worth looking into, will confirm if it's valid to use the suggested approach instead of the current one.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +51,38 @@ public function bootstrap(Tenant $tenant): void
...
+
+        // Check if the current database doesn't have the tenants table (i.e. it's not the central database)
+        if (Schema::hasTable($tenant->getTable())) {
+            throw new RuntimeException('Tenant cannot use the central database.');
```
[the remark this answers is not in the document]
lukinovec: Definitely, but won't commit or resolve this until https://github.com/archtechx/tenancy/pull/1459#discussion_r3370160389 is resolved

src/Database/Concerns/ValidatesDatabaseParameters.php
stancl: Hmm yeah I was thinking if that could break anything but we *selectively choose* when to validate some parameter so all those places that call `validateParameter(...)` should already have established that the value is not null (or at least *expect* the value to not be null, which would make an exception appropriate here).

So for optional things like charset you'd only validate the param after checking that there is some charset set — it's not null — while for things like username/password those places simply expect the values to not be null, so an exception seems appropriate to validate that expectation.

tests/TenantDatabaseManagerTest.php
stancl: Further improved in 3fb01768

tests/TenantDatabaseManagerTest.php
stancl: Went with `database managers validate parameters used in raw sql statements` in 88156d17.

src/Database/Concerns/ValidatesDatabaseParameters.php
stancl: Let's discuss null in those other threads to keep conversations separate. I've replied to the other review.

> Though every parameter we pass to validateParameter() is either string|null (e.g $databaseConfig->getUsername(), has return type string|null), or mixed (config value like charset/collation, which could be anything)

This doesn't say much to me. "Either it's string or it's mixed". So it's mixed.

I went ahead and added a cast in case this ever comes up in someone's app, not sure if regular Eloquent casts could affect this since most of these params would come from the virtual column but still worth "healing" numeric values like this since they're perfectly usable b3d11587aeee.

You can mark this review as resolved.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
stancl: Hmm that's not really what I meant.

> it's not common to ...

It should never happen, it's explicitly bad and dangerous to do that rather than just a matter of being uncommon.

By tradeoffs I meant more so things like the fact that this would probably make cached lookup pointless? Since we're running queries against the central DB in the check.

Updated here 13e32dd6ab39, if anything is wrong there let me know.

src/Database/Concerns/ValidatesDatabaseParameters.php
stancl: The new docblock for `allowedPasswordCharacters` doesn't make much sense to me:
```diff
      *
      * Used as the default allowlist in validateParameter(), which validates non-password
      * parameters such as database names or usernames.
+     *
+     * Since special characters are not used in non-password parameters commonly,
+     * we can be more strict about them to prevent SQL injection and other related issues.
      */
     protected function allowedParameterCharacters(): string
     {
@@ -31,10 +34,13 @@
     /**
      * Characters allowed in database user passwords.
      *
-     * Passwords are always quoted in the SQL statements, so it's safe
+     * Parameters are always quoted in the SQL statements, so it's safe
      * to allow a wider range of characters, as long as it doesn't include
      * characters that can break out of the quoted SQL strings (so e.g.
      * ', ", \, and ` aren't allowed).
+     *
+     * The allowlist is less strict for passwords than for other parameters
+     * because it's more common to use special characters in passwords.
      */
     protected function allowedPasswordCharacters(): string
     {
```

"Parameters are always quoted ... so it's safe to allow a **wider** range of characters". Wider meant "wider than parameters" but now we're talking about parameters here as well.

Corrected and simplified the docblocks here fbffeb84b31b. The section about escaping quoting wasn't necessary.

stancl:
One more thing that comes to mind is that the base allowlists should probably be in static properties so people can change them if needed. The current approach with methods would be super annoying to try to customize.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: OK that's much better. Resolving this

src/Database/Concerns/ValidatesDatabaseParameters.php
lukinovec: Updated the logic accordingly https://github.com/archtechx/tenancy/pull/1459/changes/48b8aac42d6893ee025f103f5d6b9e5a48f571e5

tests/TenantDatabaseManagerTest.php
lukinovec: Updated here https://github.com/archtechx/tenancy/commit/48b8aac42d6893ee025f103f5d6b9e5a48f571e5#diff-739362afc407b7efc70b323c578436cbb0336af5747051b1645ee08670154260

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: > if we could check that we're on the central connection in another way.

Maybe check if the current DB name is the central database name, and if it is, throw the exception?

Something like this could be good:
```php
$centralDbName = DB::connection(
    config('tenancy.database.central_connection', 'central')
)->getDatabaseName();

if ($dbName === $centralDbName) {
    throw new RuntimeException('Tenant cannot use the central database.');
}
```

lukinovec: https://github.com/archtechx/tenancy/pull/1459/commits/565bc41bf33766507e50a81a5657a8f48b30fb47 This seems better to me, let me know if you think the same.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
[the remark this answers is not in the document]
lukinovec: Yeah so the SQLite concern is valid.

`DatabaseConfig::makeCredentials()` stores a name like `tenant_123.sqlite` in `data->tenancy_db_name`

`SQLiteDatabaseManager::makeConnectionConfig()` stores a full db path in the db config (e.g. `/app/database/tenant_123.sqlite`)

` DB::getDatabaseName()` here returns the full db path, so `$dbName` and tenant's `tenancy_db_name` never match (full db path vs the shortened one)

So I think using `$dbName = $tenant->database()->getName()` instead of `$dbName = DB::getDatabaseName()` is the right call. We also don't have any test that would cover this, so I'm adding it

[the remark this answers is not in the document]
lukinovec: Addressed here https://github.com/archtechx/tenancy/pull/1459/commits/540e3635e27ade5cd36f2f7dde737d873fd2eda7

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +69,43 @@ public function bootstrap(Tenant $tenant): void
...
+
+    protected function verifyTenantCanUseDatabase(Tenant $tenant): void
+    {
+        /** @var \Stancl\Tenancy\Database\Models\Tenant&TenantWithDatabase $tenant */
+        $tenantDbName = $tenant->database()->getName();
+
+        // Check if any other tenant uses this tenant's database
+        if ($tenant::where($tenant->getTenantKeyName(), '!=', $tenant->getTenantKey())
+            ->where($tenant::getDataColumn() . '->' . $tenant->internalPrefix() . 'db_name', $tenantDbName)
+            ->exists()) {
```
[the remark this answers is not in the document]
stancl: Hmm I think even with distributed DBs we'd still want globally unique DB names. If someone had a setup where that wasn't the case, they could simply choose to not enable this feature (it's opt in) and implement similar logic themselves.

stancl:
I think the `validateParameter()` API accepting both single values and arrays is a bit problematic especially because we use empty arrays as example invalid values in our tests:

```php
test('mysql database manager validates charset and collation correctly', function (string $param) {
    $manager = app(MySQLDatabaseManager::class);
    $manager->setConnection('mysql');

    // using a non-string value (empty array) which is invalid
    config(["database.connections.mysql.$param" => []]);
    DB::purge('mysql');

    $tenant = Tenant::make([
        'tenancy_db_name' => 'valid_db_name',
    ]);

    expect(fn () => $manager->createDatabase($tenant))
        ->toThrow(InvalidArgumentException::class, 'Parameter has to be a string.');
})->with(['charset', 'collation']);
```

I wanted to push a minor refactor:
1. Checking `$parameters === null` outside the loop should be unnecessary because it's the first thing we check inside the loop as well, so when we iterate over `[null]` we'd notice it there
2. I really don't like seeing `validateParameter(array_filter(...))` because it makes it seem like validation can just be skipped for some things. Validating in place (as in my diff) communicates what's happening better and is cleaner.

But 1) doesn't work since `Arr::wrap(null)` actually produces `[]` so the loop never runs. And 2) also appears to break some things, I'm guessing the fact that we use `[]` for invalid values in the test above, so if we want to use `[]` as an invalid value it needs to be *part of a wider array* such that it's `[something, [], ...]` rather than just `[]`.

I think we should just make the method single value-only and add a second method for validating multiple things at once if needed. Or just make separate calls.

```diff
diff --git a/src/Database/Concerns/ValidatesDatabaseParameters.php b/src/Database/Concerns/ValidatesDatabaseParameters.php
index b66a31c784..13574f3d66 100644
--- a/src/Database/Concerns/ValidatesDatabaseParameters.php
+++ b/src/Database/Concerns/ValidatesDatabaseParameters.php
@@ -48,10 +48,6 @@
      */
     protected function validateParameter(string|array|null $parameters, string|null $allowedCharacters = null): void
     {
-        if ($parameters === null) {
-            throw new InvalidArgumentException('Parameter cannot be null.');
-        }
-
         $allowedCharacters ??= static::$allowedParameterCharacters;

         foreach (Arr::wrap($parameters) as $parameter) {
diff --git a/src/Database/TenantDatabaseManagers/MySQLDatabaseManager.php b/src/Database/TenantDatabaseManagers/MySQLDatabaseManager.php
index 912533b895..10385208c1 100644
--- a/src/Database/TenantDatabaseManagers/MySQLDatabaseManager.php
+++ b/src/Database/TenantDatabaseManagers/MySQLDatabaseManager.php
@@ -14,7 +14,7 @@
         $charset = $this->connection()->getConfig('charset');
         $collation = $this->connection()->getConfig('collation');

-        $this->validateParameter(array_filter([$database, $charset, $collation], fn ($param) => $param !== null));
+        $this->validateParameter($database);

         // MySQL defaults to the server's charset and collation
         // if charset and collation are not specified.
@@ -23,10 +23,12 @@
         $statement = "CREATE DATABASE `{$database}`";

         if ($charset !== null) {
+            $this->validateParameter($charset);
             $statement .= " CHARACTER SET `{$charset}`";
         }

         if ($collation !== null) {
+            $this->validateParameter($collation);
             $statement .= " COLLATE `{$collation}`";
         }
```

stancl:
Also I've noticed `validatePassword()` still allows `null`. Not sure if we want that, given the change we made to `validateParameter()`?

stancl:
Fixed in a7aa03158de6, let me know what you think.

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public static string $allowedPasswordCharacters = ' !#$%&()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_abcdefghijklmnopqrstuvwxyz{|}~';
+
+    /**
+     * Ensure that parameter (database name, username, etc.)
+     * only contains allowed characters before being used in SQL statements
+     * (or paths in the case of SQLiteDatabaseManager).
+     *
+     * By default, only the characters in allowedParameterCharacters() are allowed.
+     *
```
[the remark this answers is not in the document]
stancl: c9fa41111d5e

src/Database/Concerns/ValidatesDatabaseParameters.php
```diff
@@ -0,0 +1,89 @@
...
+            throw new InvalidArgumentException('Parameter has to be a string.');
+        }
+
+        $allowedCharacters ??= static::$allowedParameterCharacters;
+
+        foreach (str_split($parameter) as $character) {
+            if (! str_contains($allowedCharacters, $character)) {
+                throw new InvalidArgumentException("Forbidden character '{$character}' in parameter.");
+            }
+        }
```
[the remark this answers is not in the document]
stancl: c9fa41111d5e

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +69,43 @@ public function bootstrap(Tenant $tenant): void
...
     }
 
     public function revert(): void
     {
         $this->database->reconnectToCentral();
     }
+
+    protected function verifyTenantCanUseDatabase(Tenant $tenant): void
+    {
+        /** @var \Stancl\Tenancy\Database\Models\Tenant&TenantWithDatabase $tenant */
```
stancl, line 92: Not sure about this docblock as mentioned on Discord

src/Bootstrappers/DatabaseTenancyBootstrapper.php
```diff
@@ -41,10 +69,44 @@ public function bootstrap(Tenant $tenant): void
...
+        }
+
+        $centralDbName = DB::connection(
+            config('tenancy.database.central_connection', 'central')
+        )->getDatabaseName();
+
+        if (DB::getDatabaseName() === $centralDbName) {
+            // Throw if the current database is central.
+            // DB::getDatabaseName() is the current DB name, which should not be central at this point.
+            throw new RuntimeException('Tenant cannot use the central database.');
```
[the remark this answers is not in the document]
lukinovec: Is PostgreSQLSchemaManager intended to be used with DatabaseTenancyManager? Isn't RLS (instead of this db bootstrapper) intended with PostgreSQLSchemaManager?

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: Clarified here: https://github.com/archtechx/tenancy/pull/1459/changes/028b985e5450724bbdac5f4818e25d784044909f

The @var docblock is for the $tenant variable, it was confusingly placed right above the $tenantDbName assigment. Added a newline below the annotation to make it a bit clearer.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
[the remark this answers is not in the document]
lukinovec: Note: resolved here https://github.com/archtechx/tenancy/pull/1459/changes/34d19e94e208029834dad490905df091f44f82c1

lukinovec:
> Fixed in [a7aa031](https://github.com/archtechx/tenancy/commit/a7aa03158de643bb7a6ba7dc9c36a3ae9d101bfa), let me know what you think.

@stancl I think it's good, seems like a significant improvement. Also, the reason for the validate method to accept arrays was purely cosmetic/so that the method usage is a bit nicer, but the null-handling trade-off wasn't worth it.

I completely agree with your changes.

src/Bootstrappers/DatabaseTenancyBootstrapper.php
lukinovec: > Added a newline below the annotation to make it a bit clearer

php-cs-fixer keeps removing the newline...

PR #1460, opened 2026-05-15
Add MigrateStatus command and corresponding tests

tizianozonta:
Adds a new Artisan command tenants:migrate-status that mirrors Laravel's built-in migrate:status, but runs against tenant databases. Useful for inspecting which migrations have already been applied (or are
  still pending) for one or more tenants, without having to switch context manually.

  Usage

  # All tenants
  php artisan tenants:migrate-status

  # Specific tenant(s)
  php artisan tenants:migrate-status --tenants=<id>
  php artisan tenants:migrate-status --tenants=<id1> --tenants=<id2>

  For each tenant, the command prints a header line (Tenant: <key>) followed by the standard Laravel migration status table (Ran / Pending / Batch).

### DIFF
--- a/src/Commands/MigrateStatus.php
+++ b/src/Commands/MigrateStatus.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Stancl\Tenancy\Commands;
+
+use Illuminate\Database\Console\Migrations\StatusCommand;
+use Illuminate\Database\Migrations\Migrator;
+use Stancl\Tenancy\Concerns\DealsWithMigrations;
+use Stancl\Tenancy\Concerns\ExtendsLaravelCommand;
+use Stancl\Tenancy\Concerns\HasATenantsOption;
+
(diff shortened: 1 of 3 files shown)


stancl:
Is this not already achievable with the `tenants:run` command wrapper? And from what I can tell reading the code here this would list the status of all migrations for all tenants. These multi-tenant commands are meant to work with thousands of tenants and I'm not sure that'd be a useful output structure. If we added a dedicated command for this purpose we'd probably want some more compact or otherwise lower noise output.

tizianozonta:
You are right, running this command you get the same result:

php artisan tenants:run migrate:status \
    --option=path=/code/project//database/migrations/tenant \
    --option=realpath=1

it's only a shortcut. 
I can implement your suggestion if you want and provide a command that highlights only "problematic" migrations  on each tenant.

stancl:
Will hold off on this for now 👍 Can always be easily added later. Thanks for the PR.

PR #1468, opened 2026-07-04
perf(rls): wrap current_setting() in the generated tenant policy for per-statement evaluation

dmitrymaranik:
`archtechx/tenancy`'s RLS policy generator emits `current_setting('<tenant setting>')` **unwrapped** in the policy predicate. Postgres re-evaluates a bare `current_setting(...)` **once per row** the policy scans; wrapping it in a scalar subquery — `(select current_setting(...))` — makes it a per-statement InitPlan the planner evaluates once and caches. It's predicate-equivalent (tenant isolation unchanged) and the documented Postgres RLS performance pattern; the benefit grows with table size.

This wraps the generated call and keeps the test assertions in sync (`TraitRLSManager.php`, `TableRLSManager.php`, `TraitManagerTest.php`, `TableManagerTest.php`). I couldn't run the PHP / Laravel test suite locally — if CI surfaces another assertion on the generated policy SQL, point me at it and I'll update it.

Spotted with [pgrls](https://github.com/pgrls/pgrls) (an open-source Postgres RLS analyzer). Happy to adjust.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant-aware access checks so generated database policies evaluate the current tenant value more reliably.
  * Updated policy behavior across direct and nested relationships to keep row-level security consistent in more query paths.

### DIFF
--- a/src/RLS/PolicyManagers/TableRLSManager.php
+++ b/src/RLS/PolicyManagers/TableRLSManager.php
@@ -528,7 +528,7 @@ class TableRLSManager implements RLSPolicyManager
 
             if ($table === tenancy()->model()->getTable()) {
                 // Convert tenant key to text to match the session variable type
-                $query .= "{$column}::text = current_setting('{$sessionTenantKey}')\n";
+                $query .= "{$column}::text = (select current_setting('{$sessionTenantKey}'))\n";
                 continue;
             }
 
--- a/src/RLS/PolicyManagers/TraitRLSManager.php
+++ b/src/RLS/PolicyManagers/TraitRLSManager.php
@@ -67,7 +67,7 @@ class TraitRLSManager implements RLSPolicyManager
 
         return <<<SQL
         CREATE POLICY {$table}_rls_policy ON {$table} USING (
-            {$tenantKeyColumn}::text = current_setting('{$sessionTenantKey}')
+            {$tenantKeyColumn}::text = (select current_setting('{$sessionTenantKey}'))
         );
         SQL;
     }
@@ -87,7 +87,7 @@ class TraitRLSManager implements RLSPolicyManager
             {$parentRelationship->getForeignKeyName()} IN (
                 SELECT {$parent->getKeyName()}
                 FROM {$parent->getTable()}
-                WHERE {$tenantKeyColumn}::text = current_setting('{$sessionTenantKey}')
+                WHERE {$tenantKeyColumn}::text = (select current_setting('{$sessionTenantKey}'))
             )
         );
         SQL;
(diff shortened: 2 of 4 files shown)


stancl:
Hi, thanks for the PR.

Can you provide a little more context about the mechanics of the cache involved?

dmitrymaranik:
Hi @stancl, thanks for taking a look!

The "cache" here is Postgres's **InitPlan**, not an application-level cache.

`current_setting('my.current_tenant')` is a `STABLE` function, so in a predicate like `tenant_id::text = current_setting('my.current_tenant')` Postgres evaluates that call **once per row** the query scans — it can't fold it at plan time (it isn't `IMMUTABLE`). On a 100k-row scan that's 100k calls.

Wrapping it as `(select current_setting('my.current_tenant'))` makes it an *uncorrelated* scalar subquery. Because it doesn't reference the outer row, the planner hoists it into an **InitPlan that runs once per statement** and reuses the cached result (`$0`) for every row — the same scan becomes a single call.

It stays semantically equivalent: the GUC is set once (via `set_config`/`SET`) before the query and is constant for the statement's lifetime, so caching the value can't change which rows match — tenant isolation is identical. It's the optimization Postgres/Supabase document for RLS ([call functions with `select`](https://supabase.com/docs/guides/database/postgres/row-level-security#call-functions-with-select)), and you can confirm it in `EXPLAIN`: the wrapped form shows `InitPlan 1 (returns $0)` evaluated once, versus the bare form re-evaluating per row. The win scales with the number of rows the policy filters.

Happy to add a short benchmark or an inline comment in the PolicyManager if that'd be useful. It came up via [pgrls](https://github.com/pgrls/pgrls) flagging the generated policies (PERF001).

stancl:
So just to confirm my understanding — this is specifically at query time, the RLS policy itself would be evaluated for each row during a query (that may be trying to select a row from a large table). If we make it a subquery, the result of that subquery will be cached (for the duration of the query?) and therefore only run once.

dmitrymaranik:
Exactly right.

At query time the predicate is applied to each candidate row the scan touches, so a bare `current_setting()` fires once per row. Wrapping it in `(select …)` makes Postgres evaluate it a single time as an InitPlan at the start of that statement and reuse the cached value for every row in the same query.

And yes on the scope — it's cached **per statement execution**: each new query re-runs the InitPlan once, then caches it for that query's rows. Nothing persists across queries, so there's no staleness — a fresh query always reads the current GUC. Net effect: the tenant check goes from O(rows) `current_setting` calls to O(1) per query, with identical results.

Happy to drop a one-line comment above the generated policy explaining the `(select …)` rationale if you'd like it documented for future readers.

PR #1469, opened 2026-07-07
[4.x] Fix duplicate entry error when updating TenantPivot columns

### LINKED ISSUE #1467: Updating pivot values on TenantPivot causes duplicate entry errors
### Bug description

Having additional pivot values on a class extending `TenantPivot` (resource syncing) is currently bugged on v4.

When updating an entry in the pivot table, the `static::saved()` event listener defined in `TriggerSyncingEvents` is called and triggers `triggerAttachEvent()`. This should only happen when creating the entry and not when updating the pivot entry.

Maybe this can simply be fixed by changing `static::saved()` to `static::created`, but maybe there are annoying side effects I haven't thought of?

### Steps to reproduce

1. Create a pivot table with an additional column
2. Create a relationship that extends `TenantPivot`
3. Create a pivot entry
4. Call `$model->relationship()->updateExistingPivot($tenant->id, ['column' => 'value])`
5. After updating the entry, resource syncing is triggered and it tries to create the $model entry on the relevant tenant. As it had already been created in step 3, this triggers a duplicate key error.

### Expected behavior

If only pivot values of a `TenantPivot` are updated, resource syncing should not try to attach the model to a tenant.

### Laravel version

12.62.0

### stancl/tenancy version

v4 / main

lukinovec:
`TriggerSyncingEvents` registered the pivot attach listener on `saved`, which fires on both inserts and updates. So updating pivot columns on a `TenantPivot` (e.g. via `updateExistingPivot()`) re-ran the attach flow and tried to create the tenant resource again, causing a duplicate entry error.

Switched the listener to `created` so we only attach when the pivot record is first created (detaching already uses `deleting`, so using `created` makes things a bit more consistent).

Added a regression test before the fix (https://github.com/archtechx/tenancy/pull/1469/commits/5e3eb4322cc487d540796317e813ac2ffe076646). The fix (https://github.com/archtechx/tenancy/pull/1469/commits/d2fb4bc0d524cd98ec680e973b791f475afc3548) then made the test pass.

Also improved the `saving` listener's comment a bit so that it's clear where the "central resource not available" exception actually comes from, since that wasn't obvious (also added the `@throws` annotation to `getCentralResourceAndTenant()`).

Closes #1467

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate synced tenant resources when updating values on an existing pivot record.
  * Improved clarity around behavior when a central resource can’t be resolved, reducing the risk of creating invalid pivot records.
* **Tests**
  * Added a regression test ensuring updating pivot-only fields (without changing attachments) does not recreate or duplicate the synced tenant resource.

### DIFF
--- a/src/ResourceSyncing/TriggerSyncingEvents.php
+++ b/src/ResourceSyncing/TriggerSyncingEvents.php
@@ -28,7 +28,8 @@ trait TriggerSyncingEvents
             $pivot->getCentralResourceAndTenant();
         });
 
-        static::saved(static function (self $pivot) {
+        // Only attach when the pivot is created
+        static::created(static function (self $pivot) {
             /**
              * @var static&Pivot $pivot
              * @var SyncMaster|null $centralResource
--- a/src/ResourceSyncing/TriggerSyncingEvents.php
+++ b/src/ResourceSyncing/TriggerSyncingEvents.php
@@ -22,9 +22,9 @@ trait TriggerSyncingEvents
     public static function bootTriggerSyncingEvents(): void
     {
         static::saving(static function (self $pivot) {
-            // Try getting the central resource to see if it is available
-            // If it is not available, throw an exception to interrupt the saving process
-            // And prevent creating a pivot record without a central resource
+            // Try getting the central resource to see if it is available.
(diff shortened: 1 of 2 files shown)


lukinovec:
@coderabbitai full review

PR #1471, opened 2026-07-21
Fix: preserve custom onFail callback in UniversalRoutes::bootstrap()

### LINKED ISSUE #1470: UniversalRoutes::bootstrap() silently overrides custom onFail callbacks on identification middlewares (v3)
### Bug description

Hi, and thanks so much for your continued work on this package, it's been a great fit for our multi-tenant app, and we appreciate the effort that's gone into it.

I've run into a small issue in v3 and wanted to flag it in case it's worth a patch release.

When UniversalRoutes is enabled, bootstrap() unconditionally overwrites the static $onFail property on both identification middlewares:

https://github.com/archtechx/tenancy/blob/v3.10.0/src/Features/UniversalRoutes.php#L26

```php
foreach (static::$identificationMiddlewares as $middleware) {
    $middleware::$onFail = function ($exception, $request, $next) {
        if (static::routeHasMiddleware($request->route(), static::$middlewareGroup)) {
            return $next($request);
        }

        throw $exception;
    };
}
```

If an application defines its own onFail callback on InitializeTenancyByDomain::class or InitializeTenancyBySubdomain::class (for example, to customize error handling when tenant identification fails), that callback is silently replaced as soon as UniversalRoutes bootstraps. There's no check for whether $onFail was already set, so any custom logic is lost without warning.

### Steps to reproduce

1. Enable `UniversalRoutes` in the tenancy config.
2. Set a custom static `onFail` closure on `InitializeTenancyByDomain::class` (or the subdomain equivalent), e.g. in a service provider's boot() method.
3. Trigger a failed tenant identification.
4. The custom `onFail` logic never runs.

### Expected behavior

The custom `onFail` callback should still run (or at least be composed with / wrapped by) the `UniversalRoutes` fallback logic, rather than being discarded outright.

### Laravel version

12.x

### stancl/tenancy version

v3.10.0

JeremyRed:
Fixes #1470

This is a small, low-risk fix for the issue linked above, happy to adjust the approach if you'd prefer to solve it differently, but since the change is minimal and the impact felt high-value, I thought it was worth putting together as a starting point.

What this changes

UniversalRoutes::bootstrap() currently overwrites $onFail on both identification middlewares unconditionally, discarding any custom callback the application had set. This PR captures the existing callback before overwriting it, and falls back to it (instead of throwing) when the UniversalRoutes bypass check doesn't apply.

We've tested this against our own setup and confirmed it resolves the behaviour described in #1470. Open to feedback if you'd rather handle this differently (e.g. scoping it to v3 only given the v4/main split).

### DIFF
--- a/src/Features/UniversalRoutes.php
+++ b/src/Features/UniversalRoutes.php
@@ -23,9 +23,16 @@ class UniversalRoutes implements Feature
     public function bootstrap(Tenancy $tenancy): void
     {
         foreach (static::$identificationMiddlewares as $middleware) {
-            $middleware::$onFail = function ($exception, $request, $next) {
+            // preserve any onFail callback set before universal routes bootstraps
+            $customOnFail = $middleware::$onFail;
+
+            $middleware::$onFail = function ($exception, $request, $next) use ($customOnFail) {
                 if (static::routeHasMiddleware($request->route(), static::$middlewareGroup)) {
                     return $next($request);
+                }
+
+                if ($customOnFail instanceof Closure) {
+                    return $customOnFail($exception, $request, $next);
                 }
 
                 throw $exception;
--- a/tests/UniversalRouteTest.php
+++ b/tests/UniversalRouteTest.php
@@ -88,6 +88,24 @@ class UniversalRouteTest extends TestCase
             ->assertSuccessful()
             ->assertSee('Tenancy is initialized.');
     }
+
+    #[Test]
+    public function universal_route_runs_custom_failed_initialization_override()
+    {
+        InitializeTenancyByDomain::$onFail = function () {
+            return 'custom failed logic.';
+        };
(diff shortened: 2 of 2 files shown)


stancl:
Thanks for the PR. Unfortunately we don't intend on making these sorts of changes to v3 when v4 (with the issue addressed and the whole feature later implemented differently) will be released shortly, see https://github.com/archtechx/tenancy/issues/1470#issuecomment-5029574399

JeremyRed:
Thanks for taking a look, and for the quick response. I know v4 has been in development for a while now, figured a small, low-risk patch might still be worth having but completely understand not wanting to carry changes forward that'll be superseded soon anyway.

Thanks again for maintaining the package, we'll keep an eye out for v4!

PR #1478, opened 2026-08-05
Fix tenants:seed command name

### LINKED ISSUE #1474: tenants:seed command registers as db:seed on Laravel 13.24 (SeedCommand switched from $name to $signature)
### Summary
  
Since `laravel/framework` **v13.24.0**, the `tenants:seed` command silently registers itself as `db:seed`, overwriting Laravel's built-in `db:seed` command. As a result `tenants:seed` no longer exists, and anything that calls it (including `SeedDatabase` job and test setups using `Artisan::call('tenants:seed', ...)`) fails with a "command not found" error.

### Versions

- `stancl/tenancy`: v3.10.0
- `laravel/framework`: v13.24.0 (v13.23.0 and earlier are **not** affected)
- PHP: 8.4

### Root cause
  
In Laravel **v13.24.0**, `Illuminate\Database\Console\Seeds\SeedCommand` was refactored to define its name via `protected $signature` instead of `protected $name`:

  ```php
  // v13.23.0 and earlier
  protected $name = 'db:seed';

  // v13.24.0
  protected $signature = 'db:seed
                  {class? : ...}
                  {--class=... }
                  {--database= : ...}
                  {--force : ...}';

``` 

### Steps to reproduce

1. Install stancl/tenancy v3.10.0 on laravel/framework v13.24.0.
2. Run php artisan list.

Expected:
db:seed          Seed the database with records
tenants:seed     Seed tenant database(s).

Actual:
db:seed          Seed tenant database(s).    ← stancl's command, wrong name
(tenants:seed is missing)

Any Artisan::call('tenants:seed', ...) then throws a "The command 'tenants:seed' does not exist" error.

### Expected behavior

`tenants:seed` stays registered under its own name and Laravel's built-in `db:seed` command remains intact, regardless of how the framework's `SeedCommand` defines its name.

### Laravel version

13.24.0

### stancl/tenancy version

3.10.0

lucasmichot:
Fixes #1474.

Replaces the `version_compare` check in `Seed::__construct()` with a `configure()` override that calls `setName('tenants:seed')`. `configure()` always runs after the command name is set by Symfony's `Command` constructor, regardless of whether the installed Laravel version's `SeedCommand` sets its name via `$name` (pre-13.24) or `$signature` (13.24+), so this works across versions without checking `app()->version()`.

`configure()` is also the general place to add, remove, or tweak a parent command's arguments/options (`$this->addArgument()`, `$this->addOption()`, or `$this->getDefinition()->remove*()`), since it's guaranteed to run after the parent's own definition is built and before the command executes. Worth keeping in mind for other commands here that extend Laravel's built-in ones and need to adjust their inherited signature.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tenant database seeding command registration so it no longer replaces Laravel’s built-in database seeding command.
  * Preserved separate access to both tenant seeding and standard database seeding commands.

* **Tests**
  * Added coverage confirming each seeding command resolves and operates independently.

### DIFF
--- a/src/Commands/Seed.php
+++ b/src/Commands/Seed.php
@@ -20,10 +20,6 @@ class Seed extends SeedCommand
     {
         parent::__construct($resolver);
 
-        // Our --tenants/--skip-tenants/--with-pending options only get added automatically
-        // when the parent command isn't signature-based. Since Laravel 13.24, SeedCommand is,
-        // so we add them ourselves here -- checking first so we don't add them twice on older
-        // Laravel versions, where they're already there by this point.
         if (! $this->getDefinition()->hasOption('tenants')) {
             $this->specifyParameters();
         }
@@ -33,10 +29,6 @@ class Seed extends SeedCommand
     {
         parent::configure();
 
-        // We inherit SeedCommand's name ('db:seed') since we don't redeclare $name/$signature,
-        // so without this we'd overwrite Laravel's own db:seed command (see #1474). configure()
-        // always runs after the name is set, regardless of Laravel version, so setting it here
-        // is safe no matter which of $name/$signature the installed Laravel version uses.
         $this->setName('tenants:seed');
     }
 
--- a/src/Commands/Seed.php
+++ b/src/Commands/Seed.php
@@ -18,19 +18,26 @@ class Seed extends SeedCommand
 
     public function __construct(ConnectionResolverInterface $resolver)
     {
-        // See https://github.com/archtechx/tenancy/issues/1474
-        if (version_compare(app()->version(), '13.24.0', '>=')) {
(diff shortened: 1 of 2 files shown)


stancl:
Thanks for the PR.

The linked issue is already fixed by the code that's being replaced here.

I can see `configure()` being more appropriate, however the `if (! $this->getDefinition()->hasOption('tenants')) {` check still appears pretty magical, making the code perhaps more confusing than it is now.

lucasmichot:
Re-added the comments explaining the `hasOption('tenants')` check and the `setName()` placement - they got stripped in a later commit before I opened the PR. Should make the reasoning clear without needing an `app()->version()` check.

